PR #7, opened 2017-10-16
ipc support

step21:
Hi, everything apart from tracing should work with ipc. (I still have to find the right function for that so for now it is a stub) I am sure there are things that can be improved but it would be great if you could check if generally you are fine with how I implemented it. All tests pass as well, I think.


b-mueller:
Very cool, thanks!

PR #10, opened 2017-10-19
add ipc function for tracing

step21:
This adds tracing functionality to ipc. Probably I will add the remaining functions (that are not currently used) with a future pull request, but this one was more important as it is actually used.


step21:
Tested with `myth --ipc -t -a 0xFa52274DD61E1643d2205169732f29114BC240b3`

b-mueller:
Thanks, merged in release 0.5.6.

PR #12, opened 2017-10-20
Tls option

maurelian:
This allows you to use 

```
./myth --rpchost "mainnet.infura.io/{INFURA_KEY}" --rpcport "443" --rpctls="True" -g ./graphlol.html -a "0x2a0c0dbecc7e4d658f48e01e3fa353f44050c208"
```

or any other node which requires https to connect to.

I'm not sure why I'm being partially credited for the README.md update commit, but I spent a bit of time trying to fix it without success.


b-mueller:
Great idea, thanks!

PR #22, opened 2017-11-04
corrected sync-all flag logic

AgeManning:
This corrects the bug where mythril skips all contracts if the `--sync-all` flag is used.


b-mueller:
Thanks @AgeManning

PR #35, opened 2017-12-22
[weak random] add blockhash dependent calls detection

Raz0r:
Added second check: `block.blockhash()` dependent Ether transfers. The following patterns are detected:

- `block.blockhash(block.number)`
- `block.blockhash(block.number - N)`;
- `block.blockhash(block.number - storage_N)`
- `block.blockhash(storage_N)`

The last check has informational severity since it is quite common to compute blockhash of a value that is stored in storage. That value is usually `block.number` which is perfectly fine unless there is no check that the resulting `bytes32` value is zero (see the notice http://solidity.readthedocs.io/en/develop/units-and-global-variables.html). We could add a new constraint like `blockhash_block_storage_0 == 0` but I could not find any reference to this symbolic variable in the stack or memory.

### DIFF
--- a/mythril/analysis/modules/weak_random.py
+++ b/mythril/analysis/modules/weak_random.py
@@ -1,3 +1,4 @@
+import re
 from z3 import *
 from mythril.analysis.ops import *
 from mythril.analysis import solver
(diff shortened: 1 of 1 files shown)


b-mueller:
Included in release 0.8.25.

PR #37, opened 2017-12-25
Add retcode error checking, solc_binary argument

10a7:
Mythril `compile_solidity`:

* doesn't use its own `solc_binary` argument
* doesn't offer NIX signal tracking if `solc` has internal fatal issues,
such as SIGSEGV

This commit adds a more verbose message for a fatal error in `solc` and uses the `solc_binary` argument. It may be worthwhile to set the default argument for `solc_binary` in the function declaration.

*Note:* My system has a strange issue in which my `solc` will segfault when run *only* from the Mythril subprocess. Running the same code as below in the Python repl. It wasn't clear what the problem Mythril was running into was, as it just displayed a blank string on the segfault. This helps others debug this same issue, and fixes a bug in which `solc_binary` was not passed in.

Tested on Python 3.6.3, Arch Linux x64 4.14.5-1.

### DIFF
--- a/mythril/ether/util.py
+++ b/mythril/ether/util.py
@@ -2,7 +2,6 @@ from ethereum.abi import encode_abi, encode_int
 from ethereum.utils import zpad
 from ethereum.abi import method_id
 from mythril.exceptions import CompilerError
-import subprocess
 from subprocess import Popen, PIPE
 import binascii
 import os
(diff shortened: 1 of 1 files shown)


b-mueller:
Thanks @10a7

b-mueller:
Included in release 0.8.25.

PR #50, opened 2018-02-02
Solc versioning

step21:
It works. But might be WIP depending on what you think :)
Like, I moved the solc part in `myth` it seemed more appropriate there if it does more now.
Otherwise it imports solc and adds a function to check for a path based on version in `ether/util.py` hope that is the right place.
(Also might have to be rebased but I wanted to see if you had other comments first)


b-mueller:
Looks good! I think the `solc` package needs to be added to requirements.txt and setup.py?

step21:
Should be better. There is still one concatenation as otherwise I would need an intermediary variable but I could get rid of that too if you prefer. I tried to make the checking for an installed solc more robust too, but probably does not change much as for better checking I would need either a helper function or another import like shutil to go check in path.

b-mueller:
Doesn't work on Mac unfortunately. Looks like the solc installer doesn't support current versions of Mac OS:

```
(mythril) Cerebral-Cortex:mythril bernhardmueller$ ./myth --solv 0.4.18 -x solidity_examples/ether_send.sol 
Checking out solidity repository @ v0.4.18
Executing: git clone --recurse-submodules --branch v0.4.18 --depth 10 https://github.com/ethereum/solidity.git /Users/bernhardmueller/.py-solc/solc-v0.4.18/source
Cloning into '/Users/bernhardmueller/.py-solc/solc-v0.4.18/source'...
remote: Counting objects: 1825, done.
remote: Compressing objects: 100% (1214/1214), done.
remote: Total 1825 (delta 1169), reused 936 (delta 588), pack-reused 0
Receiving objects: 100% (1825/1825), 1.20 MiB | 1.05 MiB/s, done.
Resolving deltas: 100% (1169/1169), done.
Note: checking out '9cf6e910bd2b90d0c9415d9c257f85fe0c518de8'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

Running dependency installation script `install_deps.sh` @ /Users/bernhardmueller/.py-solc/solc-v0.4.18/source/scripts/install_deps.sh
Executing: sh /Users/bernhardmueller/.py-solc/solc-v0.4.18/source/scripts/install_deps.sh
Unsupported macOS version.
We only support Mavericks, Yosemite and El Capitan, with work-in-progress on Sierra.
```

step21:
Mmmh, sorry about that. Tbh I only tested in Linux. I have a hackintosh install of High Sierra I think that I could use to test. But it would take a bit.

step21:
It seems that the whole install process is different, as there are prebuilt binaries for linux, so py-solc just uses those, and then never uses this install script from solidity (or maybe only in some edge cases). Should I add a note about this in the docs or the help text? I guess it is unlikely that solidity will add os-x builds soon and it would take me a lot of time to try to make the solidity install script work better on os-x.

b-mueller:
> it would take me a lot of time to try to make the solidity install script work better on os-x.
Not sure if that would be worth the effort! Can you just add a note to the README?

step21:
https://github.com/ConsenSys/mythril/pull/53

PR #59, opened 2018-02-16
update security readme

blackwatertepes:
Thought perhaps you might find an update to the security readme useful.

* Weak Randoms module link moved to the correct column
* Additional direct links to Consensys security issues
* Denote that the "Call Depth Attack" has been deprecated with the EIP150 Hard Fork

Love this project, btw!


b-mueller:
Thank you @blackwatertepes

PR #62, opened 2018-02-19
added Dockerfile

thec00n:


b-mueller:
Should we use this instead of the [existing Dockerfile](https://github.com/ConsenSys/mythril/blob/master/Dockerfile)?

thec00n:
I did not see that there was already one in the master. Disregard, it's pretty much the same.

b-mueller:
ok cool, thx!

b-mueller:
Reopening this as there seem to be [issues with the current Dockerfile](https://github.com/ConsenSys/mythril/issues/65).

PR #72, opened 2018-03-06
Allow for different output formats in truffle projects

vdrg:
Now you can do stuff like `myth -to json` (it will output a json object per contract).


b-mueller:
Thank you @vdrg!

Issue #76, opened 2018-03-12
Another `IndexError: list index out of range`


thoppe:
**Version** `mythril==0.13.9`

Perhaps related to #25, contract [`0xE0F7e56e61b4267062172494D7506087205A4229`](https://etherscan.io/address/0xe0f7e56e61b4267062172494d7506087205a4229) fails to parse. Etherscan shows there are 89,000 [other contracts](https://etherscan.io/find-similiar-contracts?a=0xe0f7e56e61b4267062172494d7506087205a4229&lvl=5) with this matching bytecode, a large fraction of which were created at the same time?

``` bash
>> myth --rpc localhost:8545 -x -a 0xE0F7e56e61b4267062172494D7506087205A4229

Traceback (most recent call last):
  File "/usr/local/bin/myth", line 406, in <module>
    issues = fire_lasers(sym)
  File "/usr/local/lib/python3.5/dist-packages/mythril/analysis/security.py", line 19, in fire_lasers
    issues += module.execute(statespace)
  File "/usr/local/lib/python3.5/dist-packages/mythril/analysis/modules/unchecked_retval.py", line 53, in execute
    _state = call.node.states[call.state_index + i]
IndexError: list index out of range
```

b-mueller:
Working for me now so I'm closing this :)

```
$ myth --infura-mainnet -x -a 0xE0F7e56e61b4267062172494D7506087205A4229
==== Message call to external contract ====
Type: Warning
Contract: 0xE0F7e56e61b4267062172494D7506087205A4229
Function name: fallback
PC address: 377
The function fallback contains a function call to an address provided as a function argument.The available gas is forwarded to the called contract. Make sure that the logic of the calling contract is not adversely affected if the called contract misbehaves (e.g. reentrancy).
--------------------

==== Exception state ====
Type: Informational
Contract: 0xE0F7e56e61b4267062172494D7506087205A4229
Function name: fallback
PC address: 384
A reachable exception (opcode 0xfe) has been detected. This can be caused by type errors, division by zero, out-of-bounds array access, or assert violations. This is acceptable in most situations. Note however that assert() should only be used to check invariants. Use require() for regular input checking. The exception is triggered under the following conditions:

extcodesize: 0xffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffffff
caller: 0x0
storage_0: 0x0
[...]

Issue #75, opened 2018-03-11
False positive on SUICIDE instruction check


thoppe:
**Version**: `0.10.13`

I've tried to parse down a contract to the most minimal working example of a (potential) false positive. In the code below, I'm not sure how the function `setFlag` can modify the address `owner`, but myth seems to think I can:

``` java
pragma solidity ^0.4.2;

contract TOY  {

    address public owner;
    bool public FLAG;

    modifier checkOwner {
        require(msg.sender == owner);  _;
    }

    function TOY() { owner = msg.sender; }
    function setFlag(bool newFLAG) public { FLAG = newFLAG;  }
    function boom() public checkOwner {
        selfdestruct(owner);
    }
}
```

When running `myth -x TOY.sol` I get the output

```
==== Unchecked SUICIDE ====
Type: Warning
Contract: TOY
Function name: boom()
PC address: 505
The function boom() executes the SUICIDE instruction.
The remaining Ether is sent to a stored address

There is a check on storage index 6. This storage index can be written to by calling the function 'setFlag(bool)'.
```

b-mueller:
Thanks @thoppe. There's likely something wrong with the SSTORE analysis, I'll check.

b-mueller:
OK, I found the cause of the false positive. The compiler packs both the `address public owner` and `bool public FLAG` into storage slot 0. It looks similar to this:

```
0x[000..][FLAG (1 byte)][owner (20 bytes)]
```

Mythril sees that `setFlag` writes to the same slot and erroneously assumes that this affects the `owner` variable.

Unfortunately the analysis of storage operations is still pretty dumb. I have some ideas for improving it though, and hopefully will come up with something in the next couple of weeks!

b-mueller:
> Version: 0.10.13

Better update to 0.13.9 -> rewritten analysis engine and source mapping, makes it much easier to interpret the results.

thoppe:
**Version:** `mythril==0.13.15`

Another example, this time of contract code in the wild. [0x0dc568fc30eb731b6551848382e53bdf2c47a230](https://etherscan.io/address/0x0dc568fc30eb731b6551848382e53bdf2c47a230#code)

```==== Unchecked SUICIDE ====
Type: Warning
Contract: CryptoBetOn
Function name: killContract()
PC address: 1416
The function killContract() executes the SUICIDE instruction. The remaining Ether is sent to a stored address.
[...]

PR #80, opened 2018-03-13
Add badge to README to indicate latest pypi version

thoppe:
An issue with the recent tickets I've submitted (#75, #76) comes from not having the latest version. Since this project actively being developed, it may be useful to have the current pypi version programmatically displayed on the README. This uses 

https://badge.fury.io/

to render the badge, with the version taken from pypi itself. It should render like this:

[PyPI](https://pypi.python.org/pypi/mythril)


b-mueller:
That's useful, thanks!

PR #88, opened 2018-03-23
Myth refactoring

freewind:
I will put some comments on changes

### DIFF
--- a/myth
+++ b/myth
@@ -4,6 +4,18 @@
    http://www.github.com/b-mueller/mythril
 """
 
+from json.decoder import JSONDecodeError
+import logging
+import json
+import sys
+import argparse
+import os
+import re
+
+from ethereum import utils
+from solc.exceptions import SolcError
+import solc
+
 from mythril.ether import util
 from mythril.ether.contractstorage import get_persistent_storage
 from mythril.ether.ethcontract import ETHContract
@@ -19,18 +31,6 @@ from mythril.analysis.symbolic import SymExecWrapper
 from mythril.analysis.callgraph import generate_graph
 from mythril.analysis.security import fire_lasers
 from mythril.analysis.report import Report
-from laser.ethereum import helper
-from ethereum import utils
-from pathlib import Path
-from json.decoder import JSONDecodeError
-from solc.exceptions import SolcError
-import solc
-import logging
-import json
-import sys
-import argparse
-import os
-import re
 
 
 def searchCallback(code_hash, code, addresses, balances):
@@ -41,12 +41,11 @@ def searchCallback(code_hash, code, addresses, balances):
 
 
 def exitWithError(format, message):
-    if (format == 'text' or format == 'markdown'):
+    if format == 'text' or format == 'markdown':
         print(message)
     else:
         result = {'success': False, 'error': str(message), 'issues': []}
         print(json.dumps(result))
-
     sys.exit()
 
 
@@ -107,36 +106,31 @@ except KeyError:
 
 # Detect unsupported combinations of command line args
 
-if (args.dynld) and not (args.address):
+if args.dynld and not args.address:
     exitWithError(args.outform, "Dynamic loader can be used in on-chain analysis mode only (-a).")
 
-# Initialize data directry and signature database
+# Initialize data directory and signature database
 
 if not os.path.exists(mythril_dir):
     logging.info("Creating mythril data directory")
-
     os.mkdir(mythril_dir)
 
 # If no function signature file exists, create it. Function signatures from Solidity source code are added automatically.
 
 signatures_file = os.path.join(mythril_dir, 'signatures.json')
 
+sigs = {}
 if not os.path.exists(signatures_file):
-    logging.info("No signature database found. Creating empty database: " + signatures_file + "\n"
-        "Consider replacing it with the pre-initialized database at "
-        "https://raw.githubusercontent.com/ConsenSys/mythril/master/signatures.json")
-
-    sigs = {}
-
+    logging.info("No signature database found. Creating empty database: " + signatures_file + "\n" +
+                 "Consider replacing it with the pre-initialized database at https://raw.githubusercontent.com/ConsenSys/mythril/master/signatures.json")
     with open(signatures_file, 'a') as f:
         json.dump({}, f)
 
-else:
-    with open(signatures_file) as f:
-        try:
-            sigs = json.load(f)
-        except JSONDecodeError as e:
-            exitWithError(args.outform, "Invalid JSON in signatures file " + signatures_file + "\n" + str(e))
+with open(signatures_file) as f:
+    try:
+        sigs = json.load(f)
+    except JSONDecodeError as e:
+        exitWithError(args.outform, "Invalid JSON in signatures file " + signatures_file + "\n" + str(e))
 
 # Parse cmdline args
 
@@ -144,22 +138,21 @@ if not (args.search or args.init_db or args.hash or args.disassemble or args.gra
     parser.print_help()
     sys.exit()
 
-if (args.v):
-    if (0 <= args.v < 3):
+if args.v:
+    if 0 <= args.v < 3:
         logging.basicConfig(level=[logging.NOTSET, logging.INFO, logging.DEBUG][args.v])
+    else:
+        exitWithError(args.outform, "Invalid -v value, you can find valid values in usage")
 
-elif (args.hash):
+if args.hash:
     print("0x" + utils.sha3(args.hash)[:4].hex())
     sys.exit()
 
-
 if args.truffle:
-
     try:
         analyze_truffle_project(args)
     except FileNotFoundError:
         print("Build directory not found. Make sure that you start the analysis from the project root, and that 'truffle compile' has executed successfully.")
-
     sys.exit()
 
 # Figure out solc binary and version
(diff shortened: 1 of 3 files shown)


myth
```diff
@@ -4,6 +4,18 @@
...
+import json
+import sys
+import argparse
+import os
+import re
+
+from ethereum import utils
+from solc.exceptions import SolcError
+import solc
+
```
freewind: I moved the imports from python or libs before the ones come from  this project

myth
```diff
@@ -41,12 +41,11 @@ def searchCallback(code_hash, code, addresses, balances):
 
 
 def exitWithError(format, message):
-    if (format == 'text' or format == 'markdown'):
```
freewind: Removing `()` in `if` make it a little bit clearer

myth
```diff
@@ -107,59 +106,53 @@ except KeyError:
...
+sigs = {}
 if not os.path.exists(signatures_file):
-    logging.info("No signature database found. Creating empty database: " + signatures_file + "\n"
-        "Consider replacing it with the pre-initialized database at "
-        "https://raw.githubusercontent.com/ConsenSys/mythril/master/signatures.json")
-
-    sigs = {}
-
+    logging.info("No signature database found. Creating empty database: " + signatures_file + "\n" +
+                 "Consider replacing it with the pre-initialized database at https://raw.githubusercontent.com/ConsenSys/mythril/master/signatures.json")
```
freewind: Add a missing `+`

myth
```diff
@@ -107,59 +106,53 @@ except KeyError:
...
 if not os.path.exists(mythril_dir):
     logging.info("Creating mythril data directory")
-
     os.mkdir(mythril_dir)
 
 # If no function signature file exists, create it. Function signatures from Solidity source code are added automatically.
 
 signatures_file = os.path.join(mythril_dir, 'signatures.json')
 
+sigs = {}
```
freewind: Put `sigs` definition here because it's used in several `if` branches

myth
```diff
@@ -107,59 +106,53 @@ except KeyError:
...
-        "https://raw.githubusercontent.com/ConsenSys/mythril/master/signatures.json")
-
-    sigs = {}
-
+    logging.info("No signature database found. Creating empty database: " + signatures_file + "\n" +
+                 "Consider replacing it with the pre-initialized database at https://raw.githubusercontent.com/ConsenSys/mythril/master/signatures.json")
     with open(signatures_file, 'a') as f:
         json.dump({}, f)
 
-else:
```
freewind: This `else` is not needed because if that signature file is not exist, the `sigs` will still be empty after reading it from file again.

myth
```diff
@@ -107,59 +106,53 @@ except KeyError:
...
     parser.print_help()
     sys.exit()
 
-if (args.v):
-    if (0 <= args.v < 3):
+if args.v:
+    if 0 <= args.v < 3:
         logging.basicConfig(level=[logging.NOTSET, logging.INFO, logging.DEBUG][args.v])
+    else:
+        exitWithError(args.outform, "Invalid -v value, you can find valid values in usage")
```
freewind: Add error and tips, you may want to improve the message :)

myth
```diff
@@ -107,59 +106,53 @@ except KeyError:
...
 
-elif (args.hash):
+if args.hash:
```
freewind: removed the `else`

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
+    try:
+        solc_binary = os.environ['SOLC']
+    except KeyError:
+        solc_binary = 'solc'
 
 # Establish RPC/IPC connection if necessary
 
-if (args.address or args.init_db):
-
-        if args.infura_mainnet:
```
freewind: Fixed the indentation

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
-
-                try:
-                    host, port = args.rpc.split(":")
-                except ValueError:
-                    exitWithError(args.outform, "Invalid RPC argument, use HOST:PORT")
-
-                tls = args.rpctls
-
-                eth = EthJsonRpc(host, int(port), tls)
+eth = None
```
freewind: `eth` is used in several `if/elif` branches, define it out of them

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
+    elif args.infura_ropsten:
+        eth = EthJsonRpc('ropsten.infura.io', 443, True)
+    elif args.ganache:
+        eth = EthJsonRpc('localhost', 7545, False)
+    elif args.rpc:
+        try:
+            host, port = args.rpc.split(":")
+        except ValueError:
+            exitWithError(args.outform, "Invalid RPC argument, use HOST:PORT")
         else:
```
freewind: Add a `else` to `try/except`, make the use of `host` and `port` in correct scope

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
+        if code == "0x":
+            exitWithError(args.outform, "Received an empty response from eth_getCode. Check the contract address and verify that you are on the correct chain.")
+        else:
+            contracts.append(ETHContract(code, name=args.address))
 
 # Compile Solidity source file(s)
 
-elif (len(args.solidity_file)):
-
+elif args.solidity_file:
```
freewind: To check if a string or list `True` or `False`, we probably can use it directly without `len()`:  <https://stackoverflow.com/questions/53513/how-do-i-check-if-a-list-is-empty>

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
 
     for file in args.solidity_file:
-
         if ":" in file:
             file, contract_name = file.split(":")
         else:
             contract_name = None
 
-        file = file.replace("~", str(Path.home()))  # Expand user path
+        file = os.path.expanduser(file)
```
freewind: Please help to confirm if this change is correct, thank you

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
 
         try:
-            signatures.add_signatures_from_file(file, sigs)  # Parse file for new function signatures
```
freewind: I also removed some unnecessary comments if the variable or function name is clear enough

myth
```diff
@@ -177,147 +170,126 @@ if args.solv:
...
-
+            signatures.add_signatures_from_file(file, sigs)
             contract = SolidityContract(file, contract_name, solc_args=args.solc_args)
-
             logging.info("Analyzing contract %s:%s" % (file, contract.name))
-
-            contracts.append(contract)
+        except FileNotFoundError:
```
freewind: I put `FileNotFoundError` first because I think it may be easier to happen than the later ones, e.g. `NoContractFoundError`

myth
```diff
@@ -330,66 +302,51 @@ if args.storage:
     if not args.address:
         exitWithError(args.outform, "To read storage, provide the address of a deployed contract with the -a option.")
     else:
-        position = 0
-        length = 1
-        array = 0
```
freewind: `array` is not necessary, and it can be a `bool` rather than `int`

myth
```diff
@@ -330,66 +302,51 @@ if args.storage:
...
-            params = (args.storage).split(",")
-            if len(params) >= 1 and len(params) <= 3:
-                position = int(params[0])
-            if len(params) >= 2 and len(params) <= 3:
-                length = int(params[1])
-            if len(params) == 3:
-                if re.match("array", params[2]):
-                    array = 1
+            params = args.storage.split(",")
             if len(params) >= 4:
```
freewind: I put this checking first, so I can remove `len(params) <= 3` in later ones

myth
```diff
@@ -330,66 +302,51 @@ if args.storage:
...
-            position = int(utils.sha3(position_formated), 16)
-
         try:
             if length == 1:
-                print("{}: ".format(position) + eth.eth_getStorageAt(args.address, position))
+                print("{}: {}".format(position, eth.eth_getStorageAt(args.address, position)))
             else:
                 for i in range(position, position + length):
-                    print("{}: ".format(hex(i)) + eth.eth_getStorageAt(args.address, i))
+                    print("{}: {}".format(hex(i), eth.eth_getStorageAt(args.address, i)))
```
freewind: A question: why we need `hex()` to the `i`(which may be a `position`), but we don't need it in previous `print("{}: ".format(position)`?

myth
```diff
@@ -400,67 +357,38 @@ elif (args.graph) or (args.fire_lasers):
             exitWithError(args.outform, "Error saving graph: " + str(e))
 
     else:
-
-        _issues = []
-        report = Report(args.verbose_report)
```
freewind: I delayed the `report` creation in later place where we need to generate the report

myth
```diff
@@ -400,67 +357,39 @@ elif (args.graph) or (args.fire_lasers):
...
-            else:
-                result = {'success': True, 'error': None, 'issues': []}
-                print(json.dumps(result))
+        # Finally, output the results
+        report = Report(args.verbose_report)
+        for issue in all_issues:
+            report.append_issue(issue)
+
+        outputs = {
+            'json': report.as_json(),
```
freewind: I put the json report into `report` followed the other ones `markdown` and `text`, is it OK?

myth
```diff
@@ -400,67 +357,39 @@ elif (args.graph) or (args.fire_lasers):
...
+            'text': report.as_text() or "The analysis was completed successfully. No issues were detected.",
+            'markdown': report.as_markdown() or "The analysis was completed successfully. No issues were detected."
+        }
```
freewind: Use map to present the report choices makes it a bit more clearer, although we compute more than before. I mean we need to compute `as_text`, `as_json`, `as_markdown` no matter what the `outform` is.

mythril/analysis/report.py
```diff
@@ -76,12 +76,10 @@ def as_text(self):
 
     def as_json(self):
         issues = []
-
         for key, issue in self.issues.items():
-
             issues.append(issue.as_dict())
-
-        return json.dumps(issues)
+        result = {'success': True, 'error': None, 'issues': issues}
```
freewind: Is it good to put the whole json here?

b-mueller:
All merged, thanks!

PR #97, opened 2018-03-28
fix #90: try speeding up ipc reading

### LINKED ISSUE #90: IPC Sync is very slow
IPC sync is very slow. Currently every single IPC socket read blocks until the timeout of 2 seconds is reached. Even with the timeout reduced to 0.1 seconds, it appears that RPC is still faster.
This should be improved (but either way, reading state from the LevelDB #15 would be much better anyway).

freewind:
I see in this issue: <https://github.com/ConsenSys/mythril/issues/90>, current myth is quite slow to read data from ipc socket after removing web.py.

Current implementation of reading data from socket, has 2 issues with performance I think:

1. it connects the ipc socket every time, and the operation is expensive
2. it doesn't know if the response is complete, so it will still call `socket.recv` until the time is out, and the timeout is at least 0.2 second

The implementation is inspired from web3.py.

When I run the code (by command `myth --init-db`), it show me about 1 second for each IPC call, which is too slow (if you add `print(blockNumber)` in `contractstorage.py#initialize`)

In this PR, I tried 2 improvements:

1. reuse the socket all the time, across ipc calls, to reduce the connection cost. And the socket will be re-created only if we can't read data from it (raised `socket.timeout`)
2. when we got data from `socket.recv`, we check it immediately if we have received a valid JSON-format string. We return it immediately if it's valid JSON, otherwise make another `socket.recv`

After these changes, the `--init-db` command is much faster, here is my comparison:

```
IPC: 10000 blocks cost 108.00932908058167 seconds
RPC: 10000 blocks cost 124.78162384033203 seconds
```
IPC is a little bit faster than RPC.

Also note the timeout setting of socket is actually not affect the speed, since we almost always got correct response in time.

You can see the code changes from this PR, and which is base on an old commit(the one before the pool version), and to see if the new speed is good enough. Since I'm not a python expert, I'm not sure if there is any issues in the new code.

Another question: if the speed is good enough, do we still need to support level-db: <https://github.com/ConsenSys/mythril/issues/15>

### DIFF
--- a/mythril/ether/__init__.py
+++ b/mythril/ether/__init__.py
@@ -1,0 +1,3 @@
+import time
+
+start_time = time.time()
--- a/mythril/ether/contractstorage.py
+++ b/mythril/ether/contractstorage.py
@@ -1,6 +1,5 @@
 import os
 import hashlib
-import persistent
 import persistent.list
 import transaction
 from BTrees.OOBTree import BTree
@@ -9,7 +8,8 @@ from ZODB import FileStorage
 from multiprocessing import Pool
 import logging
 from mythril.ether.ethcontract import ETHContract, InstanceList
-
+from mythril import ether
+import time
 
 BLOCKS_PER_THREAD = 256
 NUM_THREADS = 8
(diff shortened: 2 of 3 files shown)


b-mueller:
Hi @freewind, thanks for the pull request!

It appears you were working from a branch that didn't yet have the latest multithreading changes in `/ether/contractstorage.py` (see current master). Can you try merging those changes into your branch and test if it still works?

freewind:
@b-mueller Yes, sure. And do we still need the multithreading? I just tried my changes to the latest code with multithreading, there is something wrong about the resourcing, since in my new code, I reused the same socket across different calls to EthIpc without closing it each time, which causes race condition problem in multithreading. And if we add lock to it, which will make the multi-threading no sense.

Shall we remove the multi-threading code?

freewind:
@b-mueller PR updated, please review and have a try.

In order to improve speed, I try to reuse connected socket as much as possible, and since it's in multi-threading environment,  I use `threading.local()` to make sure each thread has its own socket to avoid race condition

b-mueller:
Cool, thx.

PR #103, opened 2018-04-01
Delete unused ipc rpc methods

freewind:
You can apply this PR if you want to remove those unused ipc/rpc methods


freewind:
This one should be applied after <https://github.com/ConsenSys/mythril/pull/102>, since it contains two commits:

- make sure use same signatures.json in every test
- ignore signatures.json 

which belongs to <https://github.com/ConsenSys/mythril/pull/102>

PR #112, opened 2018-04-13
Features/integeroverflow

JoranHonig:
Implementation of integer overflow detection.

### Overflow detection
When a state is found with an add instruction the stack is inspected and we symbolically check if the result of the ADD operation can be bigger than 2^32 - 1 ( the maximum value ), in which case there is an integer overflow.
Now after some discussion with @b-mueller, we needed some way to deal with benign cases. I have implemented a simple function which explores the states after the add instruction and inspects specifically those with JUMPI and SSTORE instructions. It checks whether they are influenced by the overflowed expression and returns all these interesting states. If there is an interesting case for an integer overflow, then we report it, otherwise it's probably benign and we can leave it unnoticed.

### Other changes
I also changed the layout of the integer.py module a bit to separate the over and underflow logic


b-mueller:
Nicely done @JoranHonig! Thanks for adding tests also!

PR #114, opened 2018-04-14
Fix file comparison tests

JoranHonig:
Tests contained a bug where the expected output would be compared to the expected output instead of the actual output.

It turned out that this showed the presence of several changed files, I regenerated the changed files. @b-mueller could you check if there are no errors in the edited files?

I also added a little change to the integer module, where there was a small problem with the issue description wording.


b-mueller:
For me some of the tests are failing because some of the addresses in the generated reports and call graphs differ. For example:

```
$ diff outputs_expected/calls.sol.json outputs_current/calls.sol.json
10c10
<             "address": 661,
---
>             "address": 665,
21c21
<             "address": 779,
---
>             "address": 787,
32c32
<             "address": 858,
---
>             "address": 870,
43c43
<             "address": 869,
---
>             "address": 881
```

I think the problem is that we're not locking in the solc version, so the bytecode  generated differs depending on which version is installed. Mine is:

```
Version: 0.4.19+commit.c4cbbb05.Darwin.appleclang
```

What version do you have installed?

JoranHonig:
I'm currently on version:
```
Version: 0.4.21+commit.dfe3193c.Linux.g++
```

That would explain the difference in byte code

Issue #116, opened 2018-04-16
CircleCI build fails


b-mueller:
CircleCI build fails after merging #114 from @JoranHonig, which fixed a bug in `report_test.py` (`output_expected` had been compared to `output_expected` instead of `output_current`).

The failure is likely due a difference in to the `solc` version installed in the CircleCI image, which causes bytecode generated from Solidity files to differ slightly. It is likely that updating `solc` to 0.4.21 fixes the issue.

@birdofpreyru, as I'm not sure how to update the image, can you please look into that?

birdofpreyru:
@b-mueller I have re-generated the image, it now includes `solc:amd64 v0.4.21`, but it does not fix the failing tests.

b-mueller:
That was quick, thanks. I'll look into it.

birdofpreyru:
@b-mueller as I understand from the log, these tests are kind of snapshot tests, that compare operation output to the expected one. There is instruction in https://github.com/ConsenSys/mythril/blob/master/README_DEV.md on how to update the expected outputs, so, I guess, you just have to manually check, whether the output difference make sense to you, and update them; or look for related bugs, otherwise.

b-mueller:
Yes, that's how it's supposed to work, however some of the tests weren't working as expected before (the expected output was compared with itself). I'll try to reproduce this manually and check the outputs.

birdofpreyru:
Hmm... but two failing tests are fixed now, I see. Probably, you still missed some outputs to update?

b-mueller:
Hmm just checked again, but it seems all outputs are updated (and all tests pass locally now).

birdofpreyru:
Still failing in Circle CI? Do these test depend on some specific number of blocks in the database? Right now the test Docker image does full sync for 10 minutes during the build, thus it has only initial blocks on board, which can fail tests that depend on blocks with higher numbers.

b-mueller:
> Do these test depend on some specific number of blocks in the database?

No, that shouldn't be an issue.There's differences in the disassembly and call graphs generated from source. I still suspect that the problem is `solc` creating slightly different byte-code (even if it's the same compiler version but a different build) so that some PC addresses in the output differ.

b-mueller:
I'm currently preparing a new refactoring & bug fixing package for TopCoder, does it make sense to include this issue?

birdofpreyru:
[...]

PR #117, opened 2018-04-16
Add dependency check to test script

JoranHonig:


b-mueller:
Thanks @JoranHonig. Syncing up compilers does help it seems, but there's still some issues with the CircleCI build failing -  see #116.

Issue #108, opened 2018-04-05
Suggestions to improve Markdown report formatting


maurelian:
Some small changes would go a long way. Here is some typical output, marked up with feedback:

<img width="672" alt="2____projects_audits_civil-monorepo_civil-sale-audit__zsh__and_assertions__division___hangs_with_z3_ _issue__3823_ _ethereum_solidity" src="https://user-images.githubusercontent.com/23033765/38392017-0ad5408c-38f4-11e8-87fe-7c4232ae645f.png">

Here is how it would look with the suggestions: 

```

# Analysis result for Pausable 

No issues found.

# Analysis result for BurnableToken

## Exception state

- Type: Informational
- Contract: BurnableToken
- Function name: `_function_0x42966c68`
- PC address: 752

### Description

A reachable exception (opcode 0xfe) has been detected. This can be caused by type errors, division by zero, out-of-bounds array access, or assert violations. This is acceptable in most situations. Note however that assert() should only be used to check invariants. Use require() for regular input checking.

```

maurelian:
It would also be good to wrap all functions in back ticks `  `` ` even in paragraphs regardless of whether the function name could be extracted from the signature, ie. in both of these situations: 

`A possible integer underflow exists in the function transfer(address,uint256).`
`A possible integer underflow exists in the function _function_0x95d89b41.`

b-mueller:
Done.

PR #124, opened 2018-04-18
wrap all functions in back ticks #108

suppermancool:
wrap all functions in back ticks #108


b-mueller:
Thanks @suppermancool

Issue #118, opened 2018-04-17
Enable consistency between builds in different environments


b-mueller:
Many tests generate output files and compare them with pre-generated expected outputs. However, some generated files differ depending on the environment. For example, when the expected outputs are generated on a Mac OS system, some tests fail in he CircleCI environment (See also #116).

Most likely, this happens because the bytecode created by different versions of the solc compiler (or different builds of the same version) is slightly different. As a result, the PC addresses shown in some of the resulting graphs and analysis outputs are also different.

Refactor tests to be independent of the solc version used.

Note: The expected files in the current master are located at:

tests/testdata/outputs_expected/

They were generated on Mac OS High Sierra with the following configuration:

```
$ python --version
Python 3.6.3
$ solc --version
solc, the solidity compiler commandline interface
Version: 0.4.21+commit.dfe3193c.Darwin.appleclang
$ pip freeze
asn1crypto==0.24.0
attrdict==2.0.0
BTrees==4.4.1
certifi==2018.1.18
cffi==1.11.5
chardet==3.0.4
coincurve==7.1.0
coverage==4.5.1
cytoolz==0.9.0.1
dictionaries==0.0.1
eth-abi==1.0.0
eth-account==0.1.0a2
eth-hash==0.1.2
eth-keyfile==0.5.1
eth-keys==0.2.0b3
eth-rlp==0.1.0
eth-tester==0.1.0b23
eth-utils==1.0.2
ethereum==2.3.1
future==0.16.0
hexbytes==0.1.0
idna==2.6
laser-ethereum==0.5.20
mythril==0.15.8
pbkdf2==1.3
persistent==4.2.4.2
plyvel==1.0.4
py-ecc==1.4.2
py-flags==1.1.2
py-solc==2.1.0
pycparser==2.18
pycryptodome==3.5.1
pyethash==0.1.27
pysha3==1.0.2
PyYAML==3.12
repoze.lru==0.7
requests==2.18.4
rlp==0.6.0
scrypt==0.8.6
semantic-version==2.6.0
six==1.11.0
toolz==0.9.0
transaction==2.2.1
urllib3==1.22
z3-solver==4.5.1.0.post2
zc.lockfile==1.2.1
ZConfig==3.2.0
ZODB==5.4.0
zodbpickle==1.0
zope.interface==4.4.3
```

b-mueller:
@birdofpreyru this is the first task to be set up, please let me know if you need additional information.

PR #125, opened 2018-04-18
Migrate to Jinja2 templating

### LINKED ISSUE #123: Templating for report graph
In `mythril/analysis/callgraph.py:252` it states that you would be interested in using a templating system. I would be interested in quickly converting it to use jinja. Would that be an acceptable dependency to add to the program in your view?

joshuata:
This replicates the previous behavior with more ability to modify options.

Fixes #123

### DIFF
--- a/myth
+++ b/myth
@@ -375,7 +375,7 @@ elif args.graph or args.fire_lasers:
         else:
             sym = SymExecWrapper(contracts[0], address, max_depth=args.max_depth)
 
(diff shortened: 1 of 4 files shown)


b-mueller:
Thanks @joshuata! I'll hold off from merging this for a few days because there's an ongoing refactoring tasks that involve fixing broken integration tests #116 #118 . My concern is that merging changes into the CFG code right now might make this task harder. Keep working it though & I'll merge as soon as the refactoring is completed.

Issue #130, opened 2018-04-22
Flaky Test


JoranHonig:
# Description
Reports_test.py performs analysis on a set of contracts.
Some of the tests fail sometimes on the following contract: rubixy.sol

The following Issue is often not found:
```
        {
            "title": "Ether send",
            "description": "In the function `'_function_0xb4022950'` a non-zero amount of Ether is sent to an address taken from storage slot 5.\nThere is a check on storage index 5. This storage slot can be written to by calling the function '_function_0x67f809e9'.\n\nThere is a check on storage index 5. This storage slot can be written to by calling the function '_function_0x67f809e9'.\nThere is a check on storage index 1. This storage slot can be written to by calling the function 'fallback'.\nThere is a check on storage index 1. This storage slot can be written to by calling the function 'fallback'.",
            "function": "_function_0xb4022950",
            "type": "Warning",
            "address": 2582,
            "debug": "<DEBUG-DATA>"
        }
```

# Reproduction
If you want to reproduce the issue then you can either run the tests locally for a while. Or use circleci cli (command is ran from the project root, note that you also need to edit the config a little bit):
```
circleci build --job test -c .circleci/config.yml
```

b-mueller:
I have experienced a similar issue with Rubixi where the PC address shown in the report differed for different runs. Are you sure that the issue is not found at all, or could it be that the issue output is different?

JoranHonig:
The problem is that it sometimes does not report this issue at all.

JoranHonig:
This is probably caused by mythril.analysis.solver.get_model(constraints), due to the timeout. I'll play around with this

PR #141, opened 2018-04-25
Assertion change in integer.py and fix flaky tests

JoranHonig:
In this pull request two problems are fixed.

Fist of all. The existing constraint assumes that if expr can be bigger then the max size then there is an overflow. However, due to the types used to store elements on the stack overflow handling is already managed by z3. Integer.py therefore includes an adaption that changes the assertion to a different one, that checks if addition/multiplication can result in a lower value than one of the operands.

Second, issue #130 is fixed in this pr. The flakyness was caused by the timout used in solver. I changed this to 10000 and now I get consistent results. @b-mueller let me know if you think this delay is reasonable


b-mueller:
Well done, thanks.

10,000 timeout should be fine I guess, we'll see if it causes any issues in the long run.

Issue #143, opened 2018-04-25
Report test concurrency


JoranHonig:
The report tests currently take a lot of time, these tests should be parallelizable. This might improve the total test time a lot

joshuata:
I think a large percentage of the time is taken up in disk I/O since it reads and writes each report multiple times. Additionally, it performs each analysis three times, one for each report type.

JoranHonig:
The files aren't that big. You are right that analysis is performed three times which could be optimized, using fixtures might help here! That would improve test duration by 3X. 
The report tests could/should also be separate from the analysis test, which now kind of happens at the same time. This would also limit the need for disk IO.

However, if you profile an execution you see that most of the time is spent doing symbolic execution and analysis. The analysis runs for each separate contracts could be ran in parallel ( which should speed up the tests even more, maybe 10 times)

JoranHonig:
I'll pick this up

JoranHonig:
So, I slimmed down the entire test duration to about 60 seconds. This could be less if we decide to exclude rubixi.sol from the testcases. @b-mueller any reason that rubixi needs to be tested? Does it include a special edgecase? Otherwise I think we should exclude it from run all tests, so developers won't have to wait an entire minute to see if their code is working.

Also @joshuata, I indeed could easily build this using fixtures thanks for the tip!

b-mueller:
> @b-mueller any reason that rubixi needs to be tested?

No, it can be removed.

birdofpreyru:
@JoranHonig @b-mueller I would not focus too much on testing time, the more tests we do in CI/CD flow, the better it is. It is assumed that developers have to run all tests only when they are about to commit they changes into codebase, and at that moment we should focus on as intense testing as possible, not on whether it take 1 min more or less.

b-mueller:
I agree that more tests is good, but the Rubixi contract takes particularly long to analyze and doesn't contain any issues not found in other contracts. I think in that case the additonal 30 seconds aren't worth it.

JoranHonig:
Because analysis will be ran concurrently rubixi would only cause an additional 15 seconds (I don't know if that would be worth it). @birdofpreyru You are right that more extensive testing is better before allowing a merge or something like that. However, if the tests can be built to test some functionality in less time then that should be preferable imo. Because this allows developers to be more active in running and changing the testbase.

[...]

PR #149, opened 2018-04-27
Optimize report test

JoranHonig:
In this pullrequest several elements got implemented ( issue #143 ):
 - Analysis of different contracts is handled by a Pool
 - Analysis is only ran once using a pytest fixture
 - Report test got cleaned up ( I removed a lot of the reused code)
 - The failed tests will now print the actual difference between the expected file and the actual file, instead of just printing the changed files


b-mueller:
Thanks @JoranHonig!

PR #157, opened 2018-04-29
Remove spaces from check taint ( as these can mess with taint checkin…

### LINKED ISSUE #156: Analysis misses integer overflow
Mythril appears to miss the MUL overflow in [BECToken](https://etherscan.io/address/0xc5d105e63711398af9bbff092d4b6769c82f793d#code), a.k.a. batchOverflow. The overflows is located in the function `batchTransfer`, line 257. 

```
    uint256 amount = uint256(cnt) * _value;
```

Running the following fall to detect this issue:

```
$ myth -x BECToken.sol:BecToken
$ myth -x BECToken.sol:PausableToken
$ myth -x BECToken.sol:BecToken --max-depth 16
```

I also noticed that the runtime of the integer overflow tests increases heavily if --max-depth is increased. We might have to play with the `max_depth` setting of `search_children`.

@JoranHonig  can you look into this?

JoranHonig:
This fixes #156. 
The problem description can be found in the issue. 
This pr makes taint checking ignore spaces and changes the multiplication order.

Note that it might be possible that z3 does not swap op0 and op1 in some cases in which case taint checking will fail again. We might want to move to a more robust way of taint checking in the future.

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -72,7 +72,7 @@ def _check_integer_overflow(statespace, state, node):
     if instruction['opcode'] == "ADD":
         expr = op0 + op1
     else:
-        expr = op0 * op1
+        expr = op1 * op0
 
     # Check satisfiable
     constraint = Or(ULT(expr, op0), ULT(expr, op1))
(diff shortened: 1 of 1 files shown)


b-mueller:
Thanks @JoranHonig! I agree that the taint checking sucks, I wouldn't even say that but qualifies as taint checking. Writing a proper taint engine is certainly on the roadmap.

PR #161, opened 2018-04-30
[wip] Integer false positive detection

### LINKED ISSUE #159: Extend integer overflow false positive detection
Given the following function:
```
function add(uint256 a, uint256 b) internal constant returns (uint256) {
  uint c = a + b;
  assert(c >= a);
  return c;
}
```

Mythril currently reports an integer overflow here, however this overflow is not exploitable.
There are two things that need to be implemented. 
Firstly the analysis module should check the reachability of interesting usages. 
Secondly if an overflowed value is used by a JUMPI instruction we should check whether that JUMPI instruction can result in a REVERT action.

JoranHonig:
I added two things to limit false positives.

 _check_requires(state, node, statespace, constraint) is now used to check if a potentially interesting jumpi statement isn't part of a requires statement by checking if one of its children has a revert/assert_fail instruction.

The second part was the addition of constraint checking while searching children for usage of a tainted variable. Only reachable children with the exploitation constraint will be checked

Fixes #159

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -82,7 +82,7 @@ def _check_integer_overflow(statespace, state, node):
         logging.debug("[INTEGER_OVERFLOW] no model found")
         return issues
 
-    if instruction['opcode'] != "MUL" and not _verify_integer_overflow(statespace, node, expr, state, model, constraint, op0, op1) :
+    if not _verify_integer_overflow(statespace, node, expr, state, model, constraint, op0, op1) :
         return issues
 
     # Build issue
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -101,7 +101,7 @@ def _verify_integer_overflow(statespace, node, expr, state, model, constraint, o
     """ Verifies existence of integer overflow """
     # If we get to this point then there has been an integer overflow
     # Find out if the overflowed value is actually used
-    interesting_usages = _search_children(statespace, node, expr, index=node.states.index(state))
+    interesting_usages = _search_children(statespace, node, expr, constraint=[constraint], index=node.states.index(state))
 
     # Stop if it isn't
     if len(interesting_usages) == 0:
@@ -246,7 +246,7 @@ def _check_sstore(state, expression):
     return _check_taint(value, expression)
 
 
-def _search_children(statespace, node, expression, index=0, depth=0, max_depth=64):
+def _search_children(statespace, node, expression, constraint=[], index=0, depth=0, max_depth=64):
     """
     Checks the statespace for children states, with JUMPI or SSTORE instuctions,
     for dependency on expression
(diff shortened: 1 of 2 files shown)


mythril/analysis/modules/integer.py
```diff
@@ -270,11 +270,43 @@ def _search_children(statespace, node, expression, index=0, depth=0, max_depth=6
         current_state = node.states[j]
         current_instruction = current_state.get_current_instruction()
         if current_instruction['opcode'] in ('JUMPI', 'SSTORE'):
```
JoranHonig: For some reason, in case of BECToken.sol it never hapens that we find an SSTORE instruction. Therefore the following is not detected anymore:
```
    uint256 amount = uint256(cnt) * _value;
    require(cnt > 0 && cnt <= 20);
    require(_value > 0 && balances[msg.sender] >= amount);

    balances[msg.sender] = balances[msg.sender].sub(amount);
```

the weird thing is that it does work for the following:
```
contract Over {

  mapping(address => uint) balances;
  uint public totalSupply;

  function Token(uint _initialSupply) {
    balances[msg.sender] = totalSupply = _initialSupply;
  }

  function sendeth(address _to, uint _value) public returns (bool) {
    require(balances[msg.sender] - _value >= 0);
    balances[msg.sender] -= _value;
    if ( true) {
        balances[_to] += _value;
    }
    balances[_to] = add(2, _value);
    return true;
  }

  function add(uint256 a, uint256 b) internal constant returns (uint256) {
    uint256 c = a + b;
    assert(c >= a);
    return c;
  }

  function balanceOf(address _owner) public constant returns (uint balance) {
    return balances[_owner];
  }
}
```

b-mueller: Weird indeed. The SSTORE should happen shortly after the MUL.

JoranHonig:
The problem seems to have to do with taint checking. If I change
```
    uint256 amount = uint256(cnt) * _value;
```
to
```
    uint256 amount = uint256(cnt) + _value;
```
then it works

JoranHonig:
Also the following smaller contract shows the same problem:
```
contract Over {

  mapping(address => uint) balances;
  uint public totalSupply;

  function Token(uint _initialSupply) {
    balances[msg.sender] = totalSupply = _initialSupply;
  }

  function sendeth(address _to, uint _value) public returns (bool) {
    //if ( true) {
        balances[_to] = add(3,2 * _value);
    //}
    return true;
  }

  function add(uint256 a, uint256 b) internal constant returns (uint256) {
    uint256 c = a + b;
    assert(c >= a);
    return c;
  }

  function balanceOf(address _owner) public constant returns (uint balance) {
    return balances[_owner];
  }
}

```

JoranHonig:
I found the problem! Our taint checking is not good enough for this.
SSTORE value: 3 + 2*calldata_Over_32 + 4
expression: calldata_Over_32 + 4*2

PR #165, opened 2018-05-02
fix dockerfile

porkchop:
Apparently software-properties-common now provides for python-software-properties, which is deprecated.


b-mueller:
Thanks @porkchop

PR #168, opened 2018-05-03
Allow to call mythril as a module via python3 -m mythril <args>

tintinweb:
added this for convenience.
additional feature, not changing any functionality.

Allows one to call `python3 -m mythril <args>` instead of requiring the `myth` script to be in `PATH`. The `myth` script is still available. This works anywhere as long as mythril is installed as a package.

* moved cli script `myth` inside the package and guarded it against accidential imports
* changed `myth` to just import and call the __main__ cli script

cheers
tin


tintinweb:
I'm also planning on refactoring the cli-script (`myth` now `__main__`) to a clean interface (to make it easier to reuse that code for different scenarios). I'll submit this with another PR.

b-mueller:
Thanks @tintinweb!

PR #181, opened 2018-05-08
Bugfix _verify_integer_overflow

### LINKED ISSUE #176: Integer overflow detection misses overflow for var + 1
Example:

```
contract A {
  uint total;
  mapping (address => uint) mine;

  function add()
    public
  {
    require( total + 1 > total ); // Overflow protection
    // require( mine[msg.sender] <= total ); // This should be necessary to avoid integer overflow
    total += 1;
    mine[msg.sender] += 1; // Curiously, no integer overflow raised here...
  }

  function subtract()
    public
  {
    require( mine[msg.sender] > 0 ); // Underflow protection
    // require( mine[msg.sender] <= total ); // This avoids throwing integer underflow
    total -= 1; // mythril throws an integer underflow here
    mine[msg.sender] -= 1; // @myth-ignore UNDERFLOW - Example of syntax
  }
}
```

JoranHonig:
When checking if we can influence a value we should not consider the overflow constraint as this will lead to false positives with ++ operations. Fixes #176

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -111,11 +111,11 @@ def _verify_integer_overflow(statespace, node, expr, state, model, constraint, o
 
     if type(op0) is not int:
         op0_value = int(str(model.eval(op0, model_completion=True)))
-        model0 = _try_constraints(node.constraints, [constraint, op0 != op0_value])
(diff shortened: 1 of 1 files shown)


b-mueller:
Nice!

PR #136, opened 2018-04-24
Adds report templating

joshuata:
Fixes #134

PR #206, opened 2018-05-23
Report Templating

joshuata:
Cleans up and organizes #136.

I'm adding each type of reporting individually in separate commits so that they are easier to fix.


joshuata:
Closes #209

joshuata:
This PR is ready. The only tests that fail are due to newlines, and I believe the newer format is slightly easier to read. Due to issues with function names on my computer I can't fix the expected outputs, so somebody else will need to copy them over.

Issue #169, opened 2018-05-05
'Template not found' error when generating CFG


b-mueller:
I'm getting a 'template not found' error when attempting to generate a CFG. High Sierra, Mythril 0.16.31. This issue was initially reported by @maurelian on Gitter.

Interestingly, the tests in `graph_test.py` pass on my system.

@joshuata, can you look into this? 

```
$ myth -g graph.html solidity_examples/origin.sol 
Traceback (most recent call last):
  File "/Users/bernhardmueller/.pyenv/versions/mythril/bin/myth", line 398, in <module>
    html = generate_graph(sym, physics=args.enable_physics, phrackify=args.phrack)
  File "/Users/bernhardmueller/.pyenv/versions/3.6.3/envs/mythril/lib/python3.6/site-packages/mythril/analysis/callgraph.py", line 145, in generate_graph
    template = env.get_template('graph.html')
  File "/Users/bernhardmueller/.pyenv/versions/3.6.3/envs/mythril/lib/python3.6/site-packages/jinja2/environment.py", line 830, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/Users/bernhardmueller/.pyenv/versions/3.6.3/envs/mythril/lib/python3.6/site-packages/jinja2/environment.py", line 804, in _load_template
    template = self.loader.load(self, name, globals)
  File "/Users/bernhardmueller/.pyenv/versions/3.6.3/envs/mythril/lib/python3.6/site-packages/jinja2/loaders.py", line 113, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/Users/bernhardmueller/.pyenv/versions/3.6.3/envs/mythril/lib/python3.6/site-packages/jinja2/loaders.py", line 235, in get_source
    raise TemplateNotFound(template)
jinja2.exceptions.TemplateNotFound: graph.html
```

inf0s3cN00b:
Fixed in 0.17.1.

QuanZhang-William:
Hey, I still have this issue, did it get fixed?

joshuata:
The problem is likely cause by using a version install via `setup.py` rather than source. The templates need to be included by adding the following to `setup.py`:
```
    package_data={
        'mythril.analysis': ['templates/*']},

    include_package_data=True,
```

PR #210, opened 2018-05-24
Fixes setup.py to include templates

joshuata:
Templates were not being included in the setuptools-bundled version of mythril. This should resolve the problem.


joshuata:
Fixes #169

b-mueller:
Weird, I thought I added that already.

PR #217, opened 2018-05-29
Slimmed down Dockerfile, with `myth` set as the entrypoint.

juztin:
Also pinning to bionic, to avoid surprises with versions bumps.


b-mueller:
Thanks @juztin!

PR #258, opened 2018-06-08
Check and report status of various required packages

rocky:
Just some small shell script changes to look and report on various test prerequisites.


b-mueller:
Thanks Rocky!

PR #262, opened 2018-06-10
Remove redundant pytest requirement and sort requirements file

fgimenez:
Hi! I'm getting an error when trying to build the docker image (or install the requirements for that matter):
```
$ pip3 install -r requirements.txt
Double requirement given: pytest (from -r requirements.txt (line 20)) (already in pytest>=3.6.0 (from -r requirements.txt (line 9)), name='pytest')
```
`pytest` is declared twice in `requirements.txt`, this PR fixes it and sorts the requirements file so that this kind of errors is less likely to happen again.

BTW, this could be caught by building the docker image on CI, I can prepare a quick PR for that let me know WDYT.

Thanks,


b-mueller:
Thanks @fgimenez! AFAIK the docker image is already being built on CI. Not sure why this wasn't caught earlier.

PR #280, opened 2018-06-16
add native contracts

norhh:
gives native contract support for mythril.


b-mueller:
Hey Nikhil! Thanks, great job on that patch.

PR #304, opened 2018-07-02
Use bytecode length as CODESIZE

### LINKED ISSUE #302: CODESIZE and EXTCODESIZE
## Description
CODESIZE uses the number of instructions versus their actual size. EXTCODESIZE doesn't use a concrete number atm which it should.
@LoCorVin reported https://github.com/b-mueller/laser-ethereum/issues/41 in the laser repository.

**implementation**
To fix this problem both modules will need some way of accessing the bytecode of a contract to calculate the size in bytes, or some way to access it directly. And have the instruction implementation use that value

norhh:
fixes #302

### DIFF
--- a/mythril/disassembler/disassembly.py
+++ b/mythril/disassembler/disassembly.py
@@ -11,6 +11,7 @@ class Disassembly:
         self.xrefs = []
         self.func_to_addr = {}
         self.addr_to_func = {}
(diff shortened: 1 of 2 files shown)


norhh:
I am currently testing the EXTCODESIZE which i wrote, will set a pr by tomorrow.

PR #307, opened 2018-07-03
Get contract code from infura for EXTCODESIZE

### LINKED ISSUE #302: CODESIZE and EXTCODESIZE

norhh:
fixes #302

### DIFF
--- a/mythril/laser/ethereum/svm.py
+++ b/mythril/laser/ethereum/svm.py
@@ -1,5 +1,4 @@
 from mythril.laser.ethereum import helper, natives
-from mythril import mythril
 from ethereum import utils
 from enum import Enum
 from flags import Flags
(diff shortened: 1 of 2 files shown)


JoranHonig:
Laser should not be using infura, I'm not 100% sure but I think you need to use the dynamic loader for this

norhh:
yeah thats better

Issue #294, opened 2018-06-25
“Invalid JSON in signatures file" when generating json format output


TonyZhouCH:
**Describe the bug**
When I use mythril to output a json format,  a bug may occur which shows:  "error": "Invalid JSON in signatures file /home/xxx/.mythril/signatures.json\nExpecting ',' delimiter: line 1 column 254411 (char 254410)"

**Expected behavior**
I assume some certain contracts can cause this problem.

**Desktop (please complete the following information):**
 - OS: Ubuntu 16.04

**Additional context**
+ I can fix it by deleting the signatures.json in /.mythril, and mythril goes well then.
+ I open the signatures.json to find the error, I notice the problem is the "0x" is missing in column 254411 in my signatures.json. (sorry I have deleted it, so I can't give more details)
+ What is signatures.json??

JoranHonig:
@TonyZhouCH Thanks for the report, could you upload you signatures.json file? That might help us a lot. The signatures file is a mapping from hexadecimal function identifiers to their function description. You might find the following page interesting: https://ethereum.stackexchange.com/questions/7602/how-does-the-evm-find-the-entry-of-a-called-function

TonyZhouCH:
@JoranHonig  Hi, I have uploaded two signatures.json, which are both wrong. Hope this helps.

[signatures.zip](https://github.com/ConsenSys/mythril/files/2133399/signatures.zip)

shashank-srikant:
I faced a similar issue. 

Although the symptom's the same, my issue stemmed from the fact that i was running Mythril in parallel on several cores. Write/Update-operations to the common `signatures.json` file seemed to cause content inconsistency, leading to a invalid JSON format.

As soon as i remove parallelism and run this on a single core, the issue doesn't seem to surface.

TonyZhouCH:
@shashank-srikant  It seems your explanation is very convincing，beacuse I run mythril in parallel. Thanks a lot!

JoranHonig:
Thanks for the discussion, I believe we don't currently support concurrent runs of mythril. But imo this would be a super useful feature, with some important usecases. We'll probably pick this up in the coming weeks :+1:

JoranHonig:
Should be fixed with #308

PR #308, opened 2018-07-03
Feature: online signature hash lookup and signatures refactor

tintinweb:
Attempts to refactor the support/signatures modules to provide a `Signatures()` class which loads signatures from a specified path or from the default location. Optionally queries an online database to resolve function signatures.

The online database is https://4byte.directory (>65k sigs). I maintain a project at [smart-contract-sanctuary](https://github.com/tintinweb/smart-contract-sanctuary) that dumps contracts from etherscan.io and submits them to https://4byte.directory - this should even increase the coverage in the future.

As noted as a todo in `signatures.py` once this PR has been merged I'll submit another that moves the signature utility functionality to this class in order to have one interface to signatures.

**Dependencies:**
https://github.com/tintinweb/ethereum-input-decoder - (high level interface to eth_abi and access to 4byte.directory)

**Usage:**
```python
signatures = Signatures(enable_online_lookkup=True)  
try:
    signatures.open()  # open from default locations
except FileNotFoundError:
    logging.info("Missing function signature file. Resolving of function names from disabled.")

print(signatures.get(sighash))  # returns ["signature1","signature2"] or throws KeyError
```

cheers,
tin

### DIFF
--- a/mythril/support/signatures.py
+++ b/mythril/support/signatures.py
@@ -99,7 +99,7 @@ class SignatureDb(object):
             logging.debug("Signatures: file not found: %s" % path)
             raise FileNotFoundError("Missing function signature file. Resolving of function names disabled.")
 
-        self.signatures_file_lock = SimpleFileLock(self.signatures_file)  # lock file to prevent concurrency issues
+        self.signatures_file_lock = self.signatures_file_lock or SimpleFileLock(self.signatures_file)  # lock file to prevent concurrency issues
         self.signatures_file_lock.aquire()  # try to aquire it within the next 10s
 
         with open(path, 'r') as f:
@@ -110,7 +110,7 @@ class SignatureDb(object):
         # normalize it to {sighash:list(signatures,...)}
         for sighash, funcsig in sigs.items():
             if isinstance(funcsig, list):
-                self.signatures = sigs  # keep original todo: tintinweb - super hacky. make sure signatures.json is initially in correct format fixme
+                self.signatures = sigs
                 break  # already normalized
             self.signatures.setdefault(sighash, [])
             self.signatures[sighash].append(funcsig)
@@ -126,7 +126,7 @@ class SignatureDb(object):
         :return: self
         """
         path = path or self.signatures_file
-        self.signatures_file_lock = SimpleFileLock(path)  # lock file to prevent concurrency issues
+        self.signatures_file_lock = self.signatures_file_lock or SimpleFileLock(path)  # lock file to prevent concurrency issues
         self.signatures_file_lock.aquire()  # try to aquire it within the next 10s
 
         if sync and os.path.exists(path):
--- a/mythril/support/signatures.py
+++ b/mythril/support/signatures.py
@@ -56,9 +56,9 @@ def add_signatures_from_file(file, sigs={}):
 
 class Signatures(object):
 
-    def __init__(self, enable_online_lookkup=True):
+    def __init__(self, enable_online_lookup=True):
         self.signatures = {}  # signatures in-mem cache
-        self.enable_online_lookup =enable_online_lookkup  # enable online funcsig resolving
+        self.enable_online_lookup =enable_online_lookup  # enable online funcsig resolving
 
     def open(self, path=None):
         if not path:
@@ -91,14 +91,17 @@ class Signatures(object):
         :return: list of function signatures
         """
         if not self.signatures.get(sighash) and self.enable_online_lookup:
-            self.signatures[sighash] = Signatures.lookup_online(sighash)  # might return multiple sigs
-        return self.signatures.get(sighash)
+            funcsigs = Signatures.lookup_online(sighash)  # might return multiple sigs
+            if funcsigs:
+                # only store if we get at least one result
+                self.signatures[sighash] = funcsigs
+        return self.signatures[sighash]  # raise keyerror
 
 
     @staticmethod
     def lookup_online(sighash):
         """
-        Lookup function signatures from 4bytes.directory.
+        Lookup function signatures from 4byte.directory.
         //tintinweb: the smart-contract-sanctuary project dumps contracts from etherscan.io and feeds them into
                      4bytes.directory.
                      https://github.com/tintinweb/smart-contract-sanctuary
--- a/mythril/support/signatures.py
+++ b/mythril/support/signatures.py
@@ -1,7 +1,17 @@
 import re
+import os
+import logging
+import json
 from ethereum import utils
 
+try:
+    # load if available but do not fail
+    import ethereum_input_decoder
+except ImportError:
+    ethereum_input_decoder = None
 
+
+# TODO: tintinweb: move this and signature functionality from mythril.py to class Signatures to have one single interface.
 def add_signatures_from_file(file, sigs={}):
 
     funcs = []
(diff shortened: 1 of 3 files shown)


tintinweb:
well, the question is if the online lookup even makes sense since we should always have the source code at hand. if that's the case then I'll remove the lookup part and just refactor the signatures class with this pr :)

JoranHonig:
Hi @tintinweb, thanks for the pr! 
I'm hesitant about adding the dependency on 4byte.directory and your project (even though they seem awesome), they are also relatively small projects which might cause a higher risk that the projects get discontinued. This is more of an maintainability issue than a problem with your idea/implementation :sweat_smile:

tintinweb:
Hi @JoranHonig,

I completely understand your concern as I don't like getting new deps to my projects either :relaxed: The maintainability shouldn't be much of a blocker though. the code is versioned and published on pypi, its not rocket-science and the idea was to avoid duplicating code as other parts of the input-decoder might be interesting for mythril as well - maybe at a later stage.
The online lookup is anyway only relevant for contract analysis where solidity code is not available (disassembler feature). It comes in handy see for example [ethereum-dasm output example](https://github.com/tintinweb/ethereum-dasm)

so, two ways to proceed:
* I can remove the dependency and directly integrate an interface to 4byte.directory (best-effort, nice to have if it resolves funcsigs) if the dependency to the external site is not a blocker
* or remove the online resolve part and just refactor the signature code as initially noted

cheers
tin

p0n1:
Cool feature!

> the question is if the online lookup even makes sense since we should always have the source code at hand.

Mythril can connect to INFURA mainnet to scan contracts with byte code only.

like this `myth -xia 0x5c436ff914c458983414019195e0f4ecbef9e6dd`

JoranHonig:
@tintinweb In your lib, what happens if 4bytes is down or slow? It seems that it will throw exceptions specific to the requests library (is that correct?).

If thats the case, and we wan't to handle exceptions coming from your lib the disassembly module would also need to depend on this library. Would it be a good idea to define your own RequestFailedError or something? That way we can build the disassembly module without it being aware that your library does http requests using the requests library (which might change). 

In the current pr there doedn't yet seem to be any error handling, I think its a good thing to do to be safe in case 4bytes.directory stops working/gets overloaded.
 
;tldr I think some error handling wil help a lot with safeguarding against project failure / changes

tintinweb:
Good point @JoranHonig. Definitely a good idea to add error handling and throw abstracted meaningful exceptions. I'll add that and will also expose e.g. the request timeout and temporarily flag the service as unavailable to make sure it does not stall the overall analysis in case something is odd with 4byte.dir  :+1:.

JoranHonig:
@tintinweb Awesome, that would help a lot!

We are thinking about using leveldb for signature storage (and maybe also other persistence) vs a json file (which has given some concurrency issues #294), if you wan't you could also include this in your refactor. Otherwise we'll probably pick this up in a separate pr

tintinweb:
@JoranHonig Agree a database would also better scale than the json file. However I am not sure if leveldb can store lists. For the time being I've tried to add a simple file-based locking mechanism which is obviously not working correctly (see ci-unittest result :)). In general I'd like to keep this PR reviewable and clean. I'll have a look at the locking/leveldb things later.

I've tested the disassembler feature and the online funcsig resolving works (both if the lib is available or not). However, the easm printer right now does not print the function signature information so it's only available internally (func_to_addr, addr_to_func). We can easily resolve and add this information to the disassembly if needed.

* the online funcsig resolver is optional now as the dependencie is not added to the package deps.
* the disassembler resolves funcsig_to_address but does not use it in the output right now (added a few todo's to the code)
* timeout and proxies are now available in the online resolver
* the online resolver throws custom exceptions
* if the online resolver is unavailable it is flagged as such to not waste more time
* new interface for SignatureDb
* the unit-tests are currently failing

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+    # load if available but do not fail
+    import ethereum_input_decoder
+    from ethereum_input_decoder.decoder import FourByteDirectoryOnlineLookupError
+except ImportError:
+    # fake it :)
+    ethereum_input_decoder = None
+    FourByteDirectoryOnlineLookupError = Exception
+
+
+class SimpleFileLock(object):
```
norhh: hi @tintinweb , we can use some library here right?

norhh:
hi @tintinweb ,to store list in leveldb as far as i know one way is to dump a "[sig1, sig2, ....]"  to leveldb , or the other way would be to append sighash with a unique value and make it a key(and also a key without any appends of unique value will store the size of the list), then it is possible it iterate from 1 to size. and get key_1 to key_size values from leveldb. both these methods have their own set of limitations.

mythril/support/signatures.py
tintinweb: @norhh This is a quick shot on providing a solution for the concurrency issue until the switch is made to a real database (leveldb was mentioned earlier). The idea was to avoid adding another dependency. Free to propose something that works cross-platform.

norhh: quicker shot would have been to use a library as it might anyway get replaced by the leveldb so not a much of longterm dependency 😄 anyway it doesn't matter now.

JoranHonig: In cases such as this I also would be cautious to add another dependency, it does not have all the same problems as a long term dependency might have.
One thing that could happen (specifically with short term dependencies) is us forgetting to remove the dependency when its not needed.

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+
+        :param path: specific path to signatures.json; default mythril location if not specified
+        :return: self
+        """
+        if not path:
+            #  try default locations
+            try:
+                mythril_dir = os.environ['MYTHRIL_DIR']
+            except KeyError:
+                mythril_dir = os.path.join(os.path.expanduser('~'), ".mythril")
```
JoranHonig: This doesn't work for all os systems, a good example of how to do this can be found in mythril.py: _init_config()

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+
+        with open(path, 'r') as f:
+            sigs = json.load(f)
+
+        self.signatures_file_lock.release()  # release lock
+
+        # normalize it to {sighash:list(signatures,...)}
+        for sighash, funcsig in sigs.items():
+            if isinstance(funcsig, list):
+                self.signatures = sigs  # keep original todo: tintinweb - super hacky. make sure signatures.json is initially in correct format fixme
```
JoranHonig: Maybe remove/change the todo. As this might not make sense with a new persistence method

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+    def write(self, path=None, sync=True):
+        """
+        Write signatures database as json to file
+
+        :param path: specify path otherwise update the file that was loaded with open()
+        :param sync: lock signature file, load contents and merge it into memcached sighash db, then save it
+        :return: self
+        """
+        path = path or self.signatures_file
+        self.signatures_file_lock = SimpleFileLock(path)  # lock file to prevent concurrency issues
```
JoranHonig: maybe only initialize if signatures_file_lock is None

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+                mythril_dir = os.path.join(os.path.expanduser('~'), ".mythril")
+            path = os.path.join(mythril_dir, 'signatures.json')
+
+        self.signatures_file = path  # store early to allow error handling to access the place we tried to load the file
+
+        if not os.path.exists(path):
+            logging.debug("Signatures: file not found: %s" % path)
+            raise FileNotFoundError("Missing function signature file. Resolving of function names disabled.")
+
+        self.signatures_file_lock = SimpleFileLock(self.signatures_file)  # lock file to prevent concurrency issues
```
JoranHonig: maybe only initialize if signatures_file_lock is None

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+            logging.debug("Signatures: performing online lookup for sighash %r" % sighash)
+            try:
+                funcsigs = SignatureDb.lookup_online(sighash, timeout=timeout)  # might return multiple sigs
+                if funcsigs:
+                    # only store if we get at least one result
+                    self.signatures[sighash] = funcsigs
+                else:
+                    # miss
+                    self.online_lookup_miss.add(sighash)
+            except FourByteDirectoryOnlineLookupError as fbdole:
```
JoranHonig: Awesome! This is exactly what I meant :+1:

JoranHonig:
Good work! @tintinweb, most of my comments don't necessarily require a change

mythril/support/signatures.py
```diff
@@ -1,44 +1,244 @@
...
+        :param sighash: function signature hash as hexstr
+        :param timeout: optional timeout for online lookup
+        :param proxies: optional proxy servers for online lookup
+        :return: a list of matching function signatures for this hash
+        """
+        if not ethereum_input_decoder:
+            return None
+        return list(ethereum_input_decoder.decoder.FourByteDirectory.lookup_signatures(sighash,
+                                                                                       timeout=timeout,
+                                                                                       proxies=proxies))
```
norhh: i think this will straight away assume proxy is None, can you pass the parameters for proxy? like use the default http_proxy kind of parameters in their respective os. or make your library use the system proxy as default, that would be cool.

tintinweb: @norhh `requests` by default uses the system proxy from env if a proxy wasn't explicitly specified. That's why we pass `{}` or `None` (both equivalent) to maintain the default behavior see [github/requests](https://github.com/requests/requests/blob/master/requests/sessions.py#L447).

mythril/support/signatures.py
tintinweb: shouldn't make much of a difference besides GCing the old lock object and creating a new instance. Since `db.write()` can be called without ever calling `db.open()` I'll change this to use an existing instance if available.

mythril/support/signatures.py
tintinweb: this is more or less a reminder that we need to change the default `signatures.json` format to `{hash:[funcsig,...]}`. I'll remove the todo as - as you noted - this will have to be refactored anyway when switching to a new db backend.

mythril/support/signatures.py
tintinweb: this was basically the original code that was used in the `disassembler.py`. we should probably make the path guessing logic available in a utility module in order to avoid code duplications. Also the code in `_init_config()` does not appear to match the code in `_init_mythril_dir()`.
I would suggest to change this after merging the PR, providing a utililty function that returns the correct path and use this wherever path guessing is currently used. I wouldn't want to add this to this PR for the sake of readability.

tintinweb:
@JoranHonig addressed the parts mentioned in the PR comments. Unittests would definitely make sense but I guess as the sigdb is not a critical part we could add that at a later stage.

The ci-unittests error is odd. Looks like an environmental issue? Any clues? (my last commit did not really change anything that could lead to that error)

norhh:
@tintinweb  at first glance it seems some rlp dependency issue, most probably ALL_BYTES was removed from rlp.utils .

norhh:
@tintinweb now it should work

norhh:
hey @tintinweb  can you try merging with master and trying again? Then it should work

tintinweb:
* rebased off master as per request @norhh - seems to work now, thx

norhh:
LGTM, @JoranHonig  anything from your side?

JoranHonig:
Looks good, @tintinweb thx for the contribution!

PR #328, opened 2018-07-10
Hook implementation

### LINKED ISSUE #264: Add callback hook support to laser
## Description
The current modules execute their analysis after laser has finished symbolic execution. Ideally we want to be able to do this concurrently to save time and duplicate effort.

blocked by #222

JoranHonig:
Initial callback/instruction hook implementation

fixes #264


JoranHonig:
I'm not sure that we actually need the decorator, its just a fun construct that allows the following:

```
@LaserSVM.hook("ADD")
def add_instruction_hook(global_state):
    pass
```

PR #331, opened 2018-07-10
Implement coverage measurement

### LINKED ISSUE #256: Code coverage?
Is there a way of seeing the percentage of bytecode which was executed after an analysis, aka the code coverage?

JoranHonig:
fixes #256

### DIFF
--- a/mythril/laser/ethereum/svm.py
+++ b/mythril/laser/ethereum/svm.py
@@ -138,5 +138,5 @@ class LaserEVM:
 
     @property
     def coverage(self):
-        return reduce(lambda sum, val: sum + 1 if val else sum, self.instructions_covered) / float(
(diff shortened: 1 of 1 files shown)


JoranHonig:
sum is a builtin, I'll change it to sum_

mythril/laser/ethereum/svm.py
```diff
@@ -63,6 +66,7 @@ def sym_exec(self, main_address):
         self._sym_exec()
 
         logging.info("Execution complete")
+        logging.info("Achieved {0:.3g}% coverage".format(self.coverage))
```
norhh: this won't be visible to the user right?

JoranHonig: If you do -v1 or v2 then you can see it. This pr was only about the functionality, not really about presentation

PR #335, opened 2018-07-11
Search strategy selection and timeout implementation

### LINKED ISSUE #186: slow analysis speed on bytes-typed public storage variable variable
Running Mythril on the following program takes 41 seconds, which seems slow for Mythril when compared to its speed on more complex programs. Interestingly, if the `public` modifier is removed, the analysis takes only 0.5 seconds.

```
contract Benchmark {
    bytes public x;
}
```

```
vagrant@vagrant:~$ time myth -x host/benchmarks/perf1.sol 

real	0m40.484s
user	0m40.240s
sys	0m0.192s
```

JoranHonig:
It might make sense to in a future pr implement the following:
- Make max-depth vs time constraints optional and implement cleaner instantation

Other than that this pr allows the user to select a search strategy and an execution timeout, the combination of which should allow the user to select circumvent some problems

closes #186


b-mueller:
Works in my tests. Seems like this will be really useful once we have refactored tests to use callbacks :)

PR #342, opened 2018-07-15
Refactor search

b-mueller:
There is no more reason to restrict search to >0 balance contracts (this was originally done to keep the contract database small).

This PR removes the feature. We can re-introduce a "min_balance" search condition later if needed.

Also: 

- Output a progress message every 1,000 contracts (with -v1)
- Remove calculation of MD5 hash for every contract (was an artifact from before)

### DIFF
--- a/mythril/leveldb/client.py
+++ b/mythril/leveldb/client.py
@@ -48,9 +48,6 @@ class EthLevelDB(object):
         for account in self._get_head_state().get_all_accounts():
             if account.code is not None:
                 code = _encode_hex(account.code)
-                # md5 = hashlib.md5()
-                # md5.update(code.encode('UTF-8'))
-                # contract_hash = md5.digest()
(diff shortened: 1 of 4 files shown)


mythril/leveldb/client.py
```diff
@@ -40,27 +41,35 @@ def __init__(self, path):
...
         '''
         for account in self._get_head_state().get_all_accounts():
-            if account.code is not None and (search_all or account.balance != 0):
+            if account.code is not None:
                 code = _encode_hex(account.code)
-                md5 = hashlib.md5()
-                md5.update(code.encode('UTF-8'))
-                contract_hash = md5.digest()
-                contract = ETHContract(code, name=contract_hash.hex())
+                # md5 = hashlib.md5()
```
JoranHonig: Why not delete this?

PR #348, opened 2018-07-17
mythril & cli: fix missing strategy param

p0n1:
```bash
mythdev -g test.html test.sol
```
This leads to an error.

`TypeError: __init__() missing 1 required positional argument: 'strategy'`

This pull request should fix this.


norhh:
Great, @p0n1 thanks for the fix.

PR #354, opened 2018-07-19
Addin _ to codecopy function

LoCorVin:
Missing _ will not find the appropriate function when handling the codecopy instruction


JoranHonig:
Thanks!

PR #355, opened 2018-07-19
Implements initial transaction based symbolic execution

JoranHonig:


mythril/laser/ethereum/transaction.py
```diff
@@ -0,0 +1,59 @@
+import logging
+from mythril.laser.ethereum.state import GlobalState, Environment, CalldataType
+from mythril.laser.ethereum.cfg import Node, Edge, JumpType
+from z3 import BitVec
+
+
+class CallTransaction:
```
b-mueller: I think `MessageCall` would be a  more accurate name for this class.

mythril/laser/ethereum/transaction.py
```diff
@@ -0,0 +1,59 @@
...
+        self.call_value = BitVec("callvalue", 256)
+        self.origin = BitVec("origin", 256)
+
+        self.open_states = None
+
+    @property
+    def has_ran(self):
+        return self.open_states is None
+
+    def run(self, open_world_states: list, evm):
```
b-mueller: Does "open world states" refer to the *opening* states?

mythril/laser/ethereum/transaction.py
JoranHonig: I'll change it

mythril/laser/ethereum/transaction.py
JoranHonig: I haven't implemented the functionality behind this part yet. My idea was to store a reference to the final world states of a transaction.
So a transaction would result in say 10 terminating states, the world states for those global states would then be referenced in the open_states variable of the transaction

JoranHonig: So its more like, the  "valid" states we can continue execution from

JoranHonig: Yeah I had indeed already thought of that ^^, that should save a lot of time

mythril/laser/ethereum/transaction.py
```diff
@@ -0,0 +1,59 @@
...
+        self.callee_address = callee_address
+        self.caller = BitVec("caller", 256)
+        self.gas_price = BitVec("gasprice", 256)
+        self.call_value = BitVec("callvalue", 256)
+        self.origin = BitVec("origin", 256)
+
+        self.open_states = None
+
+    @property
+    def has_ran(self):
```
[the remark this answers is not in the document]
JoranHonig: it should be is not None ^^'

mythril/laser/ethereum/transaction.py
```diff
@@ -0,0 +1,59 @@
...
+        return self.open_states is None
+
+    def run(self, open_world_states: list, evm):
+        """ Runs this transaction on the evm starting from the open world states """
+        # Consume the open states
+        open_states = open_world_states[:]
+        del open_world_states[:]
+
+        for open_world_state in open_states:
```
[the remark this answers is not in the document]
JoranHonig: I think then it might become hard to keep track of what transactions there are. 
I think both methods are (now vs your suggestion) are reasonable

JoranHonig:
@b-mueller Is it okay to merge now?

b-mueller:
@JoranHonig yep!

JoranHonig:
@b-mueller Cool, then I'll get started on contract creation transactions

PR #357, opened 2018-07-21
Enhance/340

norhh:
fixes 340


mythril/mythril.py
```diff
@@ -118,40 +119,69 @@ def _init_mythril_dir(self):
...
+        config.set('defaults', "#– Linux: ~/.ethereum/geth/chaindata")
+        config.set('defaults', "#– Windows: %USERPROFILE%\\AppData\\Roaming\\Ethereum\\geth\\chaindata")
+        config.set('defaults', 'leveldb_dir', leveldb_fallback_dir)
+        return config
+
+    @staticmethod
+    def _add_dynamic_loading_option(config):
+        config.set('defaults', '#– To connect to Infura use dynamic_loading: infura')
+        config.set('defaults', '#– To connect to Ipc use dynamic_loading: ipc')
+        config.set('defaults', '#– To connect to Rpc use '
```
norhh: Should we be adding comments for every option or can we just mention that properly in the wiki?

mythril/mythril.py
```diff
@@ -118,40 +119,69 @@ def _init_mythril_dir(self):
...
+            config.write(fp)
+
+        config.read(self.config_path, 'utf-8')
+        leveldb_dir = config.get('defaults', 'leveldb_dir', fallback=leveldb_fallback_dir)
         return os.path.expanduser(leveldb_dir)
 
+    @staticmethod
+    def _add_default_options(config):
+        config.add_section('defaults')
+        return config
```
JoranHonig: you don't need the returns for these functions right?

mythril/mythril.py
JoranHonig: I think it doesn't hurt to also have them here

mythril/mythril.py
```diff
@@ -226,6 +256,23 @@ def set_api_rpc_localhost(self):
         self.eth = EthJsonRpc('localhost', 8545)
         logging.info("Using default RPC settings: http://localhost:8545")
 
+    def set_api_from_config_path(self):
+        config = ConfigParser(allow_no_value=False)
+        config.optionxform = str
+        config.read(self.config_path, 'utf-8')
+        if config.has_option('defaults', 'dynamic_loading'):
```
JoranHonig: You could change it to the following one liner:
```python
dynamic_loading = config.get('defaults', 'dynamic_loading') if config.has_option('defaults', 'dynamic_loading') else 'infura'
```

JoranHonig: Not that the current situation isn't good now :sweat_smile:

mythril/mythril.py
```diff
@@ -118,40 +119,69 @@ def _init_mythril_dir(self):
...
+        if not config.has_option('defaults', 'leveldb_dir'):
+            config = self._add_leveldb_option(config, leveldb_fallback_dir)
+
+        if not config.has_option('defaults', 'dynamic_loading'):
+            config = self._add_dynamic_loading_option(config)
+
+        with codecs.open(self.config_path, 'w', 'utf-8') as fp:
+            config.write(fp)
+
+        config.read(self.config_path, 'utf-8')
```
JoranHonig: why do we need this step?

mythril/mythril.py
```diff
@@ -118,40 +119,69 @@ def _init_mythril_dir(self):
         return mythril_dir
 
     def _init_config(self):
```
JoranHonig: The return parameter of this function is the leveldb dir, this is not reflected in the signature or function name, it seems to mee that the getting of the leveldb should be done in a separate function. But for now can you make the docs reflect that this happens, and create an issue to fix this if you agree with me ^^.

mythril/mythril.py
norhh: oops

mythril/mythril.py
norhh: But that would be quite a long 1 liner.

mythril/mythril.py
norhh: yes.

mythril/mythril.py
norhh: i was actually thinking of separating leveldb part, setting up of the config file in a different pr.

mythril/interfaces/cli.py
```diff
@@ -164,7 +164,7 @@ def main():
             address, _ = mythril.load_from_address(args.address)
         elif args.solidity_file:
             # Compile Solidity source file(s)
-            #if args.graph and len(args.solidity_file) > 1:
+            # if args.graph and len(args.solidity_file) > 1:
```
b-mueller: Why is this commented out? Does generating call graphs from multiple files work?

norhh: After a small glance at the code it doesn't seem that graph would work for multiple files, but this was commented out before this pr, i just added a space before '#' to make it pep-8 compliant :smiley: .We may probably need to uncomment this.

mythril/interfaces/cli.py
```diff
@@ -91,11 +91,6 @@ def main():
         print("Mythril version {}".format(VERSION))
         sys.exit()
 
-    # -- args sanity checks --
-    # Detect unsupported combinations of command line args
-
-    if args.dynld and not args.address:
-        exit_with_error(args.outform, "Dynamic loader can be used in on-chain analysis mode only (-a).")
```
norhh: @b-mueller should we be allowing this case?

mythril/interfaces/cli.py
b-mueller: Not sure who commented this out and why. Let's uncomment it?

mythril/mythril.py
JoranHonig: :+1: sure,

mythril/interfaces/cli.py
norhh: Removed

PR #378, opened 2018-07-24
Fix function signature search

b-mueller:


JoranHonig:
Is this ready for review?

b-mueller:
Yep. Some problem with integration tests but those are not related to this PR.

birdofpreyru:
@b-mueller that's the problem with integration tests themselves - I rebuilt test image to include changes for latest update of API, but that's does not yet work as supposed.

PR #380, opened 2018-07-25
Use solidity filename for truffle

### LINKED ISSUE #356: Wrong file name listed for a contract in a truffle project
**Describe the bug**
On an error (IntegerOverflow), I am seeing the wrong filename, the filename of the truffle project JSON rather than filename of the contract.

**To Reproduce**

Create a truffle project

```
$  truffle init IntegerOverflow
$ cd IntegerOverflow
```
create IntegerOverflow.sol: 

```
pragma solidity ^0.4.23;

contract IntegerOverflow {
    uint public count = 1;

    function run(uint256 input) public {
        count *= input;
    }
}
```
Run: 

```
$ myth -d -t .   # Note -d option is ignored.
# Analysis result for IntegerOverflow:

==== Integer Overflow ====
Type: Warning
Contract: IntegerOverflow
Function name: _function_0xa444f5e9
PC address: 174
A possible integer overflow exists in the function `_function_0xa444f5e9`.
The addition or multiplication may result in a value higher than the maximum representable integer.
--------------------
In file: IntegerOverflow.json:7
                         ^^^^

count *= input

--------------------
```

**Expected output***:

```
In file: IntegerOverflow.sol:7
```

Note that using JSON output `-o json` doesn't change the wrong data.

**Desktop (please complete the following information):**
 - OS: Ubuntu
 - Version: bionic

norhh:
fixes #356

### DIFF
--- a/mythril/support/truffle.py
+++ b/mythril/support/truffle.py
@@ -1,5 +1,5 @@
 import os
-from pathlib import Path
+from pathlib import PurePath
 import re
 import sys
 import json
@@ -32,7 +32,7 @@ def analyze_truffle_project(args):
             try:
                 name = contractdata['contractName']
                 bytecode = contractdata['deployedBytecode']
-                filename = Path(contractdata['sourcePath']).name
+                filename = PurePath(contractdata['sourcePath']).name
             except:
                 print("Unable to parse contract data. Please use Truffle 4 to compile your project.")
                 sys.exit()
--- a/mythril/support/truffle.py
+++ b/mythril/support/truffle.py
@@ -1,4 +1,5 @@
 import os
+from pathlib import Path
 import re
 import sys
 import json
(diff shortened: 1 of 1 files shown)


norhh:
hey @rocky can you check whether this commit fixes #356 or not?

mythril/support/truffle.py
```diff
@@ -31,6 +31,7 @@ def analyze_truffle_project(args):
             try:
                 name = contractdata['contractName']
                 bytecode = contractdata['deployedBytecode']
+                filename = contractdata['sourcePath'].split('/')[-1]
```
rocky: Looking at some other code I learned about  [pathlib](https://docs.python.org/3.5/library/pathlib.html)  in Python 3.5+. 

```python
import pathlib
>>> pathlib.Path("/a/b/c/foo.bar").name
'foo.bar'
```

Old Python/Unix folks know this function/command as `basename`.

norhh: Does the pathlib work for windows?( i think the slashes are opposite for windows if thats the case the split i wrote will fail)

rocky: Excellent point! It looks like you've checked .

PR #375, opened 2018-07-24
Check on active account vs environment (which is shallow copied)

JoranHonig:
Fixes #367

PR #384, opened 2018-07-28
Fixes a bug that causes the storage of the active account to act as a…

LoCorVin:
… singleton

It is not enough to deepcopy the storage, as all copies of that global state will still point to the same environment, pointing to the same active account, pointing to the same now overwritten active account. I noticed this because after processing a STOP, the next instruction CALLVALUE of the alternative branch had the same storage content as stop.

One solution seems to be the proposed change actually deepcopying the entire environment, but it is also necessary to ensure that when branching mstate gets a new instance.

### DIFF
--- a/mythril/laser/ethereum/instructions.py
+++ b/mythril/laser/ethereum/instructions.py
@@ -709,7 +709,11 @@ class Instruction:
         try:
             # Create a fresh copy of the account object before modifying storage
 
-            global_state.environment = deepcopy(global_state.environment)
+            for k in global_state.accounts:
+                if global_state.accounts[k] == global_state.environment.active_account:
+                    global_state.accounts[k] = deepcopy(global_state.accounts[k])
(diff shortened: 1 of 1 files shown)


norhh:
Nice job spotting it. I think joran made some changes in #384 which are in some way related to this. But yes i think we need to entirely copy the state in the jumpi, but i think we should not do a deep copy of node which happens as a result of deepcopying the global state, I am not totally sure. @JoranHonig  any comments on this.

LoCorVin:
Agreed, I really like the idea to do be as lazy when copying as possible to reduce the overall overhead. But my change might be only a quickfix for my problem.

JoranHonig:
Yeah, this is the same problem that got fixed in #375, there is some specific behavior necessary for the copy which is why we use the override. #375 has been merged, @LoCorVin could you check if that also solves your problem?

LoCorVin:
Yes it solves the problem, thanks guys.

PR #385, opened 2018-07-28
Force determinism in iterating over dictionary keys...

rocky:
and replace an "import *" with something specific


rocky:
This is a improvement, but it looks like there is more of this ilk that is needed. Will address later.

PR #391, opened 2018-07-30
Bugfix/fix fallback name

norhh:
remove simplifying bool in jumpi and also change fallback name.
As rocky mentioned in #390  someone using fallback as a function  it might cause a confusion. As of now i am thinking of using fallback instead of fallback() may do the work but i am not sure of it, so any suggestions?Another option as suggested by rocky is changing description of each issue for fallback seperately:- if thats the case then any suggestions on the display text for fallback.

### DIFF
--- a/mythril/laser/ethereum/instructions.py
+++ b/mythril/laser/ethereum/instructions.py
@@ -767,25 +767,25 @@ class Instruction:
         instr = disassembly.instruction_list[index]
 
         # True case
-        condi = condition if type(condition) == BoolRef else condition != 0
+        condi = simplify(condition) if type(condition) == BoolRef else condition != 0
         if instr['opcode'] == "JUMPDEST":
-            if (type(condi) == bool and condi) or (type(condi) == BoolRef and not is_false(simplify(condi))):
+            if (type(condi) == bool and condi) or (type(condi) == BoolRef and not is_false(condi)):
                 new_state = copy(global_state)
                 new_state.mstate.pc = index
                 new_state.mstate.depth += 1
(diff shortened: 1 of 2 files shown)


JoranHonig:
why remove the simplify?

rocky:
> For now I'd say that just 'fallback' is fine

Using two words like "unamed fallback" will make more clear than dropping parenthesis that this is not a function name.

norhh:
If i change it to unnamed fallback this is how it will look

norhh:
After changing it to unnamed fallback .
```
==== Integer Overflow ====
Type: Warning
Contract: Crowdfunding
Function name: unnamed fallback
PC address: 501
A possible integer overflow exists in the function `unnamed fallback`.
The addition or multiplication may result in a value higher than the maximum representable integer.
--------------------
In file: ether_send.sol:24

balances[msg.sender] += msg.value

--------------------
```
This is fine right?

b-mueller:
> For now I'd say that just 'fallback' is fine

I agree."fallback function" is official terminology for the code that is executed if no function hash matches. We can expect Ethereum devs to know what a fallback function is. "Unknown fallback" sounds weird.

rocky:
It was "unnamed" not "unknown". I got that phrase from https://www.bitdegree.org/learn/solidity-fallback-functions/

> In Solidity, a contract may have precisely one unnamed function, which cannot have arguments, nor return anything.

That said, if others find "unnamed fallback" confusing , drop the "unnamed". I don't think it will be that big a deal because wrappers will have other ways to indicate what we're talking about.

PR #392, opened 2018-07-31
Refactor pr

JoranHonig:
All modules that provide interfacing with ethereum clients etc are moved to mythril.ethereum.interface, to group them together

Ether was renamed contract, I feel that this name better describes the content of that module

asm was moved to the disassembler module as it fits better there looking at the dependencies


JoranHonig:
Sadly pycharm didn't do the refactors for the tests, before manually fixing the tests I'd like to get some feedback on the refactors themselves

Issue #397, opened 2018-08-01
Assertion error when run on some contracts with -l


norhh:
## Description

Specifically happens when -l is used.

## How to Reproduce
```console
$ myth -xila 0x8f46325b00536da3714eb8646b10ed653d5b17f2
Traceback (most recent call last):
  File "./myth", line 9, in <module>
    mythril.interfaces.cli.main()
  File "/home/nikhil/MYTHRIL/mythril/mythril/interfaces/cli.py", line 209, in main
    max_depth=args.max_depth, execution_timeout=args.execution_timeout)
  File "/home/nikhil/MYTHRIL/mythril/mythril/mythril.py", line 382, in fire_lasers
    issues = fire_lasers(sym, modules)
  File "/home/nikhil/MYTHRIL/mythril/mythril/analysis/security.py", line 20, in fire_lasers
    issues += module.execute(statespace)
  File "/home/nikhil/MYTHRIL/mythril/mythril/analysis/modules/integer.py", line 34, in execute
    issues += _check_integer_overflow(statespace, state, node)
  File "/home/nikhil/MYTHRIL/mythril/mythril/analysis/modules/integer.py", line 84, in _check_integer_overflow
    if not _verify_integer_overflow(statespace, node, expr, state, model, constraint, op0, op1):
  File "/home/nikhil/MYTHRIL/mythril/mythril/analysis/modules/integer.py", line 103, in _verify_integer_overflow
    interesting_usages = _search_children(statespace, node, expr, constraint=[constraint], index=node.states.index(state))
  File "/home/nikhil/MYTHRIL/mythril/mythril/analysis/modules/integer.py", line 233, in _search_children
    taint_result = TaintRunner.execute(statespace, node, state, initial_stack=taint_stack)
  File "/home/nikhil/MYTHRIL/mythril/mythril/laser/ethereum/taint_analysis.py", line 107, in execute
    records = TaintRunner.execute_node(node, record, index)
  File "/home/nikhil/MYTHRIL/mythril/mythril/laser/ethereum/taint_analysis.py", line 139, in execute_node
    records.append(TaintRunner.execute_state(records[-1], current_state))
  File "/home/nikhil/MYTHRIL/mythril/mythril/laser/ethereum/taint_analysis.py", line 144, in execute_state
    assert len(state.mstate.stack) == len(record.stack)
AssertionError

...
$
```
## Expected behavior

Should execute properly

## Environment

- Mythril version: latest master
[...]

PR #404, opened 2018-08-01
Transaction based message calling

JoranHonig:
In this pr I changed how calls are handled. These are now modeled as message call transactions (which they are :))

Since the mstate is specific for each transaction, the depth is now also managed for each transaction separately.

Transaction creation and termination now happens with the use of exceptions, the transaction creation ops now also have a "post" execution implementation. This mutator is executed after the called transaction is finished.


norhh:
Can you do space optimization by an chance, as It killed my laptop when run on address mentioned in #397 ,

JoranHonig:
@norhh Ah this probably has to do with:
```
Since the mstate is specific for each transaction, the depth is now also managed for each transaction separately.
```

Try setting ```--execution-timeout 30``` or something like that

mythril/laser/ethereum/instructions.py
```diff
@@ -953,79 +942,181 @@ def call_(self, global_state):
...
+        if global_state.last_return_data is None:
+            # Put return value on stack
+            return_value = BitVec("retval_" + str(instr['address']), 256)
+            global_state.mstate.stack.append(return_value)
+            global_state.mstate.constraints.append(return_value == 0)
+
+            return [global_state]
 
+            # Copy memory
+        global_state.mstate.mem_extend(memory_out_offset,
```
norhh: It's possible for memory_out_offset to be some bitvec stuff right?

mythril/laser/ethereum/instructions.py
```diff
@@ -953,79 +942,181 @@ def call_(self, global_state):
...
+                                       min(memory_out_size.as_long(), len(global_state.last_return_data)))
+        for i in range(min(memory_out_size.as_long(), len(global_state.last_return_data))):
+            global_state.mstate.memory[i + memory_out_offset] = global_state.last_return_data[i]
```
norhh: Also possible here

b-mueller:
I started playing with calls, here's my little test program:

```
from mythril.disassembler.disassembly import Disassembly
from mythril.laser.ethereum import svm
from mythril.analysis.callgraph import generate_graph
import logging


logging.basicConfig(level=logging.INFO)

caller_code = Disassembly("6080604052348015600f57600080fd5b5073deadbeefdeadbeefdeadbeefdeadbeefdeadbeef73ffffffffffffffffffffffffffffffffffffffff166389627e13336040518263ffffffff167c0100000000000000000000000000000000000000000000000000000000028152600401808273ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001915050602060405180830381600087803b15801560be57600080fd5b505af115801560d1573d6000803e3d6000fd5b505050506040513d602081101560e657600080fd5b8101908080519060200190929190505050500000a165627a7a72305820fdb1e90f0d9775c94820e516970e0d41380a94624fa963c556145e8fb645d4c90029")
caller_address = "0xaffeaffeaffeaffeaffeaffeaffeaffeaffeaffe"

callee_code = Disassembly("608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff16806389627e13146044575b600080fd5b348015604f57600080fd5b506082600480360381019080803573ffffffffffffffffffffffffffffffffffffffff1690602001909291905050506084565b005b8073ffffffffffffffffffffffffffffffffffffffff166108fc3073ffffffffffffffffffffffffffffffffffffffff16319081150290604051600060405180830381858888f1935050505015801560e0573d6000803e3d6000fd5b50505600a165627a7a72305820a6b1335d6f994632bc9a7092d0eaa425de3dea05e015af8a94ad70b3969e117a0029")
callee_address = "0xdeadbeefdeadbeefdeadbeefdeadbeefdeadbeef"

caller_account = svm.Account(caller_address, caller_code, contract_name="Caller")
callee_account = svm.Account(callee_address, callee_code, contract_name="Callee")

accounts = {
    caller_address: caller_account,
    callee_address: callee_account
}

laser = svm.LaserEVM(accounts)
laser.sym_exec(caller_address)

html = generate_graph(laser)

print(html)
```

Running this produced the error:

```
  File "/Users/bernhardmueller/Projects/mythril/mythril/laser/ethereum/svm.py", line 120, in execute_state
    global_state.accounts[return_global_state.environment.active_account.contract_name]
KeyError: 'Caller'
```

Shouldn't accounts be indexed by address instead of name? What role does contract_name play?

b-mueller:
Good opportunity to clean up the class interface? I'm also getting confused by the existence of "SymExecWrapper". I remember I thought it was a good idea long ago, but not sure if that's true anymore :)

b-mueller:
FYI I created a separate issue: #413

norhh:
Also i think it would be good to reduce the default expected timeout, some people may run it on their pc(which may not be good) and if it fails they may think they can't run it and leave it that way.
Or
another option is we can add that to FAQ.

mythril/laser/ethereum/instructions.py
JoranHonig: fixed

mythril/laser/ethereum/instructions.py
JoranHonig: fixed

JoranHonig:
@norhh Yeah, I think it might be a good idea to put this in the readme.md actually, since otherwise we might risk users not seeing this information.

b-mueller:
@JoranHonig works now! It also creates the correct graph, only the nodes of different accounts don't get a different color like before.

mythril/laser/ethereum/instructions.py
```diff
@@ -953,79 +942,197 @@ def call_(self, global_state):
...
+            memory_out_size = util.get_concrete_int(memory_out_size) if isinstance(memory_out_size,
+                                                                                   ExprRef) else memory_out_size
+        except AttributeError:
+            global_state.mstate.stack.append(BitVec("retval_" + str(instr['address']), 256))
+            return [global_state]
+
+            # Copy memory
+        global_state.mstate.mem_extend(memory_out_offset,
+                                       min(memory_out_size.as_long(), len(global_state.last_return_data)))
+        for i in range(min(memory_out_size.as_long(), len(global_state.last_return_data))):
```
norhh: same here

mythril/laser/ethereum/instructions.py
```diff
@@ -953,79 +942,197 @@ def call_(self, global_state):
...
+                                                                                       ExprRef) else memory_out_offset
+            memory_out_size = util.get_concrete_int(memory_out_size) if isinstance(memory_out_size,
+                                                                                   ExprRef) else memory_out_size
+        except AttributeError:
+            global_state.mstate.stack.append(BitVec("retval_" + str(instr['address']), 256))
+            return [global_state]
+
+            # Copy memory
+        global_state.mstate.mem_extend(memory_out_offset,
+                                       min(memory_out_size.as_long(), len(global_state.last_return_data)))
```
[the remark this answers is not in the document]
JoranHonig: The try catch should handle that

mythril/laser/ethereum/instructions.py
JoranHonig: The try catch should handle that

mythril/laser/ethereum/instructions.py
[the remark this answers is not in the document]
norhh: i mean there is no as_long() for integers right?In that case delegatecall_post throws an error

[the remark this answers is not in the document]
JoranHonig: its always an integer ^^', should be fixed now

PR #407, opened 2018-08-03
Remove reference to ipc

### LINKED ISSUE #406: Namespace has no ipc
Ipc has been removed from the options in cli.py, however it is still "used" on line 115.

This causes an AttributeError, to fix this the reference can just be removed

JoranHonig:
fixes #406

### DIFF
--- a/mythril/interfaces/cli.py
+++ b/mythril/interfaces/cli.py
@@ -107,12 +107,12 @@ def main():
 
     try:
         # the mythril object should be our main interface
-        # infura = None, rpc = None, rpctls = None, ipc = None,
+        # infura = None, rpc = None, rpctls = None
         # solc_args = None, dynld = None, max_recursion_depth = 12):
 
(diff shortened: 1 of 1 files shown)


mythril/interfaces/cli.py
```diff
@@ -107,12 +107,12 @@ def main():
...
     try:
         # the mythril object should be our main interface
-        # infura = None, rpc = None, rpctls = None, ipc = None,
+        # infura = None, rpc = None, rpctls = None
         # solc_args = None, dynld = None, max_recursion_depth = 12):
 
         mythril = Mythril(solv=args.solv, dynld=args.dynld,
                           solc_args=args.solc_args)
-        if args.dynld and not (args.ipc or args.rpc or args.i):
+        if args.dynld and not (args.rpc or args.i):
```
norhh: there is also an args.ipc in line 124 or somewhere near that.

PR #416, opened 2018-08-06
List 'markdown' option in command line help

b-mueller:


rocky:
One thing I noticed is that the issue messages are given in markdown format with respect to backticks whether or not the `--markdown` flag is given.  See for example the formating `assert` and `require` in [assert_minimal](https://ethereumanalysisbenchmarks.github.io/Suhabe/index.html#assert_minimal).

If this were not markdown/rst it would be "assert" or perhaps `assert'

I'm not suggesting we necessarily change this, just noting that it is.

PR #419, opened 2018-08-06
Contract creation transaction

### LINKED ISSUE #191: Implement multi-transactional analysis
Mythril reports an assertion violation on this benchmark:
```
pragma solidity ^0.4.19;

contract Benchmark {
    uint256 private param;

    function Benchmark(uint256 _param) public {
        require(_param > 0);
        param = _param;
    }

    function run() {
        assert(param > 0);
    }
}
```

child of #203

JoranHonig:
This pr implements a contract creation transaction in addition to the logic needed in laserEVM to start executing a contract with creation code.

closes #191

### DIFF
--- a/mythril/laser/ethereum/state.py
+++ b/mythril/laser/ethereum/state.py
@@ -95,7 +95,6 @@ class MachineState:
         self.pc = 0
         self.stack = []
         self.memory = []
-        self.memory_size = 0
         self.gas = gas
         self.constraints = []
         self.depth = 0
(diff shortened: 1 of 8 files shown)


JoranHonig:
In #404 @b-mueller mentioned the absence of different colours in the callgraph. They are back with this pr
Its now fix with the following addition
```
            if return_global_state is None:
                e.global_state.world_state.node = global_state.node <- this one
                self.open_states.append(e.global_state.world_state)
                new_global_states = []
```

mythril/laser/ethereum/state.py
```diff
@@ -115,22 +106,15 @@ def mem_extend(self, start, size):
...
-                    self.memory.append(0)
-                    n_append -= 1
+        if size and start + size > len(self.memory):
+            n_append = start + size - len(self.memory)
 
-                # FIXME: this does not seem right
-                self.memory_size = size
+            while n_append > 0:
+                self.memory.append(0)
+                n_append -= 1
```
norhh: we can write it as self.memory += n_append*[0] . It would be simple.

JoranHonig: Fixed, and added a nice test

norhh:
When i ran this on https://etherscan.io/address/0xf6da07a8ffb5683314715d8c7a1a5c952b4e74ca#code by copy pasting it to a file, it threw an attribute error,can you check this?

JoranHonig:
@norhh Great catch. If the constructor "fails" then no contract is created. I didn't yet implement something that would handle that. Should be fixed now

PR #430, opened 2018-08-13
Add better lock support using sys calls

### LINKED ISSUE #409: Make the circleci deterministically pass the tests for py35
## Description

Currently tests results are non deterministic for py35.

## Locations of errors
tests/report_test.py::test_json_report
tests/report_test.py::test_markdown_report
tests/report_test.py::test_text_report

one possible error set: https://circleci.com/gh/ConsenSys/mythril/704
other possible error set:https://circleci.com/gh/ConsenSys/mythril/708

norhh:
fixes #409
Uses sys calls over file locking.

### DIFF
--- a/mythril/support/signatures.py
+++ b/mythril/support/signatures.py
@@ -24,46 +24,34 @@ except ImportError:
     FourByteDirectoryOnlineLookupError = Exception
 
 
-class SimpleFileLock(object):
-    # todo: replace with something more reliable. this is a quick shot on concurrency and might not work in all cases
+try:
+    # Posix based file locking (Linux, Ubuntu, MacOS, etc.)
+    import fcntl
 
-    def __init__(self, path):
-        self.path = path
-        self.lockfile = pathlib.Path("%s.lck" % path)
-        self.locked = False
+    def lock_file(f, exclusive=False):
+        if f.mode == 'r' and exclusive:
+            raise Exception('Please use non exclusive mode for reading')
+        flag = fcntl.LOCK_EX if exclusive else fcntl.LOCK_SH
+        fcntl.lockf(f, flag)
 
-    def aquire(self, timeout=5):
-        if self.locked:
-            raise Exception("SimpleFileLock: lock already aquired")
+    def unlock_file(f):
+        return
 
-        t_end = time.time() + timeout
-        while time.time() < t_end:
-            # try to aquire lock
-            try:
-                self.lockfile.touch(mode=0o0000, exist_ok=False)  # touch the lockfile
-                # lockfile does not exist. we have a lock now
-                self.locked = True
-                return
-            except FileExistsError as fee:
-                # check if lockfile date exceeds age and cleanup lock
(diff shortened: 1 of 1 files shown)


norhh:
should fix it as it has sustained multiple [tests](https://circleci.com/gh/ConsenSys/mythril/tree/pull%2F430)

JoranHonig:
Awesome! Let me run a few more and check the code :+1:

mythril/support/signatures.py
```diff
@@ -128,21 +113,22 @@ def write(self, path=None, sync=True):
...
         path = path or self.signatures_file
-        self.signatures_file_lock = self.signatures_file_lock or SimpleFileLock(path)  # lock file to prevent concurrency issues
-        self.signatures_file_lock.aquire()  # try to aquire it within the next 10s
 
         if sync and os.path.exists(path):
             # reload and save if file exists
-            with open(path, 'r') as f:
+            with open(path, "r") as f:
+                lock_file(f)
                 sigs = json.load(f)
```
JoranHonig: I think we also didn't do this before but Ideally we should have the following: 
```
lock_file(f)
try:
    sigs = json.load(f)
finally:
    unlock_file(f)
```

I'm not sure how the locking libraries work in this case, but this would ensure that the lock is released.

mythril/support/signatures.py
```diff
@@ -24,46 +24,34 @@
...
+        return os.path.getsize(os.path.realpath(f.name))
 
-        try:
-            self.lockfile.unlink()  # might throw if we force unlock and the file gets removed in the meantime. TOCTOU
-        except FileNotFoundError as fnfe:
-            logging.warning("SimpleFileLock: release(force=%s) on unavailable file. race? %r" % (force, fnfe))
+    def lock_file(f, exclusive=False):
+        if f.mode == 'r' and exclusive:
+            raise Exception('Please use non exclusive mode for reading')
+        msvcrt.locking(f.fileno(), msvcrt.LK_RLCK, file_size(f))
```
JoranHonig: There seems to be no difference between exclusive and non exclusive locks, that shouldn't be right right?

norhh: Thats for windows, i didn't find any flag for shared locks, so i put it as a TODO.

mythril/support/signatures.py
norhh: in unix, if the file is closed then the lock will be closed.But yes thats a better way to do it.

b-mueller:
That's great, I often had issues with old signatures.json lock file!

mythril/support/signatures.py
JoranHonig: I'm not 100% sure but I think by default in windows you have an "exclusive" lock, and you can choose to make it shared. (Don't know how the lib implements that though :sweat_smile: )

norhh: The library doesn't seem to have any shared lock.Something more should be done for it. and also there doesn't seem to be any performance difference with either of the locks, so i thought it might be better to leave it for now.

PR #437, opened 2018-08-14
Do multiple transactions

JoranHonig:
Executes 2 transactions by default, in the future we might want to do this based on coverage

Additionally adds hinting if no contract was created during the contract creation transaction execution


norhh:
can you explain why 2 transactions are necessary there?Also a comment there might be of some help for the future.

JoranHonig:
@norhh Yeah sure. There might be situations where a contract is not directly exploitable from its initial state. Because we now don't have over approximated storage anymore we won't be able to cover a full contract with one transaction pass.
Note that just doing 2 transactions is a very weak solution to this problem (but very necessary in the short term), and in the future should be replaced  with logic that works based on coverage/permitted resources

PR #441, opened 2018-08-14
Fix integration tests

norhh:
create signatures file if it doesn't exists.


norhh:
seems like integration tests have passed, but this is quite a hacky approach.Might be good to use some sort of db or a third party lib which does atomic file operations over syscalls.

PR #465, opened 2018-08-20
Concolic exec and tests

JoranHonig:
This pr includes a bit of logic to make concolic execution easy, this is ultimately to help the integration of evm testsuite. 
Which this pr also does. The initial framework has been setup and all the addition tests have been added

closes #442

PR #466, opened 2018-08-20
Features/arithmetic test

JoranHonig:
Adds arithmetic tests to repo and fixes the problems found because of them, builds on top of #465


JoranHonig:
The exponent tests are excluded for now as this isn't fully Implemented yet

PR #468, opened 2018-08-21
Add support for loading bytecode from file

lexibrent:
Many shells have a limit on the length of command-line arguments, so including the full bytecode of a large contract to be analysed on the command line may be impossible. Hence, I have added an option for loading the bytecode directly from a file.


norhh:
Thanks a lot, I wanted to have this option as it simplifies doing lot of things.

Issue #446, opened 2018-08-16
get_sourcecode_info in contract needs to do this for the contract creation to after recent extension to contract creation analysis


LoCorVin:
## Description

Lookup of sourcecode info may fail or return the wrong result when an issue was found in the constructor code. To Fix this the srcmap in addition to srcmap-runtime has to be added to the contract and the lookup then needs information on whether or not to look in the creation or transactional mapping in addition to the address.

## How to Reproduce

Contract: minmal.sol

```console
pragma solidity ^0.4.24;

contract Minimal {

	constructor(uint8 var1){
		assert(var1>0);
	}
}
```

```console
$ myth -x minimal.sol
Traceback (most recent call last):
  File "../../mythril/myth", line 9, in <module>
    mythril.interfaces.cli.main()
  File "/home/eltano/mythril/mythril/interfaces/cli.py", line 209, in main
    max_depth=args.max_depth, execution_timeout=args.execution_timeout)
  File "/home/eltano/mythril/mythril/mythril.py", line 385, in fire_lasers
    issue.add_code_info(contract)
  File "/home/eltano/mythril/mythril/analysis/report.py", line 38, in add_code_info
    codeinfo = contract.get_source_info(self.address)
  File "/home/eltano/mythril/mythril/ether/soliditycontract.py", line 105, in get_source_info
    solidity_file = self.solidity_files[self.mappings[index].solidity_file_idx]
TypeError: list indices must be integers or slices, not NoneType
...
$
```

## Expected behavior

Return an issue of an assert exception code that can be reached.

PR #480, opened 2018-08-26
Workaround for Suhabe assert_constructor.sol...

rocky:
File index for instruction address not found. But is there only *one* source file.

### DIFF
--- a/mythril/ether/soliditycontract.py
+++ b/mythril/ether/soliditycontract.py
@@ -102,6 +102,13 @@ class SolidityContract(ETHContract):
 
         index = helper.get_instruction_index(self.disassembly.instruction_list, address)
 
+        if index is None:
+            # File index for instruction address not found. But is there only *one* source file?
+            file_idx_set = {m.solidity_file_idx for m in self.mappings}
+            if len(file_idx_set) == 1:
+                # Yes, so use the since file mentioned.
+                index = file_idx_set.pop()
+
         solidity_file = self.solidity_files[self.mappings[index].solidity_file_idx]
 
         filename = solidity_file.filename
--- a/mythril/ether/soliditycontract.py
+++ b/mythril/ether/soliditycontract.py
@@ -106,7 +106,7 @@ class SolidityContract(ETHContract):
             # File index for instruction address not found. But is there only *one* source file?
             file_idx_set = {m.solidity_file_idx for m in self.mappings}
             if len(file_idx_set) == 1:
-                # Yes, so use the since file mentioned.
+                # Yes, so use the single file mentioned.
                 index = file_idx_set.pop()
 
         solidity_file = self.solidity_files[self.mappings[index].solidity_file_idx]


norhh:
Oops, i was just about to put the PR for #446 .

rocky:
> Oops, i was just about to put the PR for #446
 
Still, this kind of thing can happen, so it adds robustness to the code.

mythril/ether/soliditycontract.py
```diff
@@ -102,6 +102,13 @@ def get_source_info(self, address):
 
         index = helper.get_instruction_index(self.disassembly.instruction_list, address)
 
+        if index is None:
+            # File index for instruction address not found. But is there only *one* source file?
+            file_idx_set = {m.solidity_file_idx for m in self.mappings}
+            if len(file_idx_set) == 1:
+                # Yes, so use the since file mentioned.
+                index = file_idx_set.pop()
```
norhh: this always returns 0 right?

rocky: It probalby returns 0 - but I think it not wise to make that kind of assumption when it isn't necessary.

Semantically, if the address isn't explicitly found, and there is only one file index, then use that file index. That you some know that file indices start from 0 and that there coudn't have been two file indices where one having index 0 was deleted are additional assumptions.

rocky: I see a typo "since file" -> single file index"

norhh: oops , the index there means the address index in the instructions list.

norhh: oops, the index there means the index of the instruction in the assembly code list.(and yep that is counter intuitive).

mythril/ether/soliditycontract.py
```diff
@@ -102,6 +102,13 @@ def get_source_info(self, address):
 
         index = helper.get_instruction_index(self.disassembly.instruction_list, address)
 
+        if index is None:
+            # File index for instruction address not found. But is there only *one* source file?
+            file_idx_set = {m.solidity_file_idx for m in self.mappings}
+            if len(file_idx_set) == 1:
+                # Yes, so use the single file mentioned.
+                index = file_idx_set.pop()
```
norhh: the index here means instruction index.

rocky: I believe it is  a _file_ index, not an instruction index. Right?

rocky: Hmm. I suppose we could look for an instruction that has the right file index, and since there is only one file index, it should be easy to find. But all of this feels a little convoluted because were looking for an instruction index just to be able to reuse the instruction-index-based code below. 

Again, if you can think of a better way, that'd be perfect

JoranHonig:
This pr is getting a bit dated, and the code around it has changed a bit. Is it okay if we close this pr and I set up an issue that describes the problem to fix:
```
Currently a SolidityContract object only gathers/stores mappings for a single source file, in order to do correct issue reporting for multiple file contracts we need to add support for this in the SolidityContract class.
```

b-mueller:
Yep

Issue #473, opened 2018-08-21
Error getting source


rocky:
## Description

mythril gives a traceback in running:

[benchmarks/Suhabe/benchmarks/assert_constructor.sol](https://github.com/ConsenSys/evm-analyzer-benchmark-suite/blob/948ff12ef857ed84e252d3f1701cc831b86124bf/benchmarks/assert_constructor.sol)

Below are details. 

(This is a virtue of regularly running the benchmark suite.)

[Pull request 472](https://github.com/ConsenSys/mythril/pull/472) was opened to tweak the error message but, the traceback still remains. 

Benchmark attribute_store.sol seems to have a similar problem. If fixing this doesn't fix that, I'll open a separate issue.

## How to Reproduce

```console
$ myth -x benchmarks/Suhabe/benchmarks/assert_constructor.sol 
No contract was created during the execution of contract creation Try to increase the resources for creation exection (max-depth or create_timeout)
Traceback (most recent call last):
  File "/home/rocky/.pyenv/versions/3.6.6/bin/myth", line 11, in <module>
    load_entry_point('mythril', 'console_scripts', 'myth')()
  File "/src/external-vcs/github/consensys/mythril/mythril/interfaces/cli.py", line 218, in main
    create_timeout=args.create_timeout)
  File "/src/external-vcs/github/consensys/mythril/mythril/mythril.py", line 374, in fire_lasers
    issue.add_code_info(contract)
  File "/src/external-vcs/github/consensys/mythril/mythril/analysis/report.py", line 38, in add_code_info
    codeinfo = contract.get_source_info(self.address)
  File "/src/external-vcs/github/consensys/mythril/mythril/ether/soliditycontract.py", line 105, in get_source_info
    solidity_file = self.solidity_files[self.mappings[index].solidity_file_idx]
TypeError: list indices must be integers or slices, not NoneType
```

## Expected behavior

No traceback

## Environment

- mythril  b5afa9ff1aa2b5dc8863d29aa9e0a24b34eb4747so
- Version: 0.4.24+commit.e67f0147.Linux.g++
- Python 3.6.6
- Ubuntu bionic

norhh:
This is related to #446

rocky:
A possible solution for this is https://github.com/ConsenSys/mythril/pull/480 .

PR #481, opened 2018-08-26
Implement correct mapping for contract creation code

### LINKED ISSUE #446: get_sourcecode_info in contract needs to do this for the contract creation to after recent extension to contract creation analysis

norhh:
fixes #446 , #473


mythril/ether/ethcontract.py
```diff
@@ -17,7 +17,8 @@ def __init__(self, code, creation_code="", name="Unknown", enable_online_lookup=
         code = re.sub(r'(_+.*_+)', 'aa' * 20, code)
 
         self.code = code
-        self.disassembly = Disassembly(self.code, enable_online_lookup=enable_online_lookup)
+        self.disassembly = Disassembly(code, enable_online_lookup=enable_online_lookup)
+        self.creation_disassemble = Disassembly(creation_code, enable_online_lookup=enable_online_lookup)
```
JoranHonig: disassembly

JoranHonig: An eth contract is used to model contracts that we don't have the construction code for, so we wouldn't need this here I think

norhh: But SolidityContract takes the disassembly from ETHContract right?

JoranHonig: Yes, a solidity contract has creation code and is an ETHContract. The ethcontract baseclass takes case of the runtime disassembly. 
However the ETHContract is not a solidity contract and (generally) does not have creation code. The current implementation does have it sometimes, but by default it is ""

I do see that ETHContract already stored the creation_code, so your extension is not illogical.
Lets keep it like this for now

norhh:
@JoranHonig  any additional changes required here?

mythril/ether/ethcontract.py
JoranHonig: shouldn't it be self.creation_disassembly?

norhh: Oops I thought I changed it

norhh: Wait! It’s changed right?

b-mueller:
@JoranHonig I think this can be merged?

PR #492, opened 2018-08-31
Put issue instructions in comments...

rocky:
so users don't have to delete them. In bug-report.md, put possibly
the more likley thing to do at the top.

I know this seems silly, but some users (i.e. me) can spend a lot of time writing issues. So anything to simplify or improve things help a lot.


norhh:
Awesome, I wanted to do that

PR #494, opened 2018-09-02
Add EVM Push, Dup and Swap tests

norhh:


mythril/laser/ethereum/state.py
```diff
@@ -141,8 +141,11 @@ def mem_extend(self, start, size):
         :param start: Start of memory extension
         :param size: Size of memory extension
         """
-        self.memory += [0] * max(0, start + size - self.memory_size)
-
+        if self.memory_size > start + size:
```
JoranHonig: Why this change?

tests/laser/evm_testsuite/evm_test.py
```diff
@@ -51,20 +52,28 @@ def test_vmtest(test_name: str, pre_condition: dict, action: dict, post_conditio
...
+            caller_address=action['caller'],
+            origin_address=action['origin'],
+            code=action['code'][2:],
+            gas=action['gas'],
+            data=binascii.a2b_hex(action['data'][2:]),
+            gas_price=int(action['gasPrice'], 16),
+            value=int(action['value'], 16)
+        )
+    except VmException:
+        return
```
JoranHonig: Is this correct? This just ignores the test if it fails right?

tests/laser/evm_testsuite/evm_test.py
```diff
@@ -76,4 +85,5 @@ def test_vmtest(test_name: str, pre_condition: dict, action: dict, post_conditio
         for index, value in details['storage'].items():
             expected = int(value, 16)
             actual = get_concrete_int(account.storage[int(index, 16)])
+            print(actual, expected)
```
JoranHonig: Remove this print statement

mythril/laser/ethereum/state.py
norhh: thats seems faster than the older method.

tests/laser/evm_testsuite/evm_test.py
norhh: some tests check how errors are dealt so thats why, also i think i should check whether the test expects for the error in the exception catch.

tests/laser/evm_testsuite/evm_test.py
norhh: argh

norhh:
After a look at yellow paper, it seems push has a bit of a problem which isn't caught by tests as they are not checking the stack in a case,

tests/laser/evm_testsuite/evm_test.py
norhh: removed

tests/laser/evm_testsuite/evm_test.py
norhh: changed

mythril/laser/ethereum/instructions.py
```diff
@@ -92,15 +91,15 @@ def swap_(self, global_state):
             stack = global_state.mstate.stack
             stack[-depth - 1], stack[-1] = stack[-1], stack[-depth - 1]
         except IndexError:
-            raise StackUnderflowException()
+            raise VmException('Stack underflow exception')
```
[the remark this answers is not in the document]
norhh: changed

mythril/laser/ethereum/instructions.py
```diff
@@ -68,10 +68,17 @@ def jumpdest_(self, global_state):
...
     @instruction
     def push_(self, global_state):
+        push_instruction = global_state.get_current_instruction()
+        push_value = push_instruction['argument'][2:]
         try:
-            value = BitVecVal(int(global_state.get_current_instruction()['argument'][2:], 16), 256)
+            no_of_bytes = int(push_instruction['opcode'][4:])
         except ValueError:
-            raise VmException('Invalid Argument for PUSH')
+            raise VmException('Invalid Push instruction')
```
JoranHonig: will this ever happen?

mythril/laser/ethereum/instructions.py
```diff
@@ -68,10 +68,17 @@ def jumpdest_(self, global_state):
 
     @instruction
     def push_(self, global_state):
```
JoranHonig: I think this implementation is getting a bit too complicated. 
I think this can be fixed in a more simple way using Concat

mythril/laser/ethereum/instructions.py
norhh: nope, unless we want to input disassembled code in the future.

mythril/laser/ethereum/instructions.py
norhh: Concat is same as doing a "0" * len right?first i will try to simplify in a typical way and then lets see

mythril/laser/ethereum/state.py
JoranHonig: Is it much faster? I liked the cleanness of the previous implementation, and if its just a small performance improvement then I'd like to keep it like that

norhh: I am not sure it is big or small but it is around 20%

norhh: also this method somehow survived some of the sha3 tests while the former killed my ram, but those tests were meant for gas exhaustion so it might not matter.

JoranHonig:
It seems that the merge with master failed

norhh:
integration tests failed due to some TLS errors

PR #495, opened 2018-09-02
Fix issue 488: Broken CFG HTML output

dmuhs:
The problem here is that the added `<!DOCTYPE html>` makes the browser switch from [quirks mode](https://en.wikipedia.org/wiki/Quirks_mode) to standards mode. While that's what you usually want when declaring a doctype, it can break some layout logic.
In standards mode, elements like `html` and `body` don't occupy the whole page anymore but instead wrap around the smallest element. As only the network canvas had an explicit height defined (but it was calculated and fairly small), the body wrapped around the tiny canvas and all other relative heights took that as a reference point - making the whole layout small.

The solution is to explicitly assign a relative height to `html` and `body` tags and then scale up the UI elements relative to them. Also removed a redundant p-tag around the canvas `div`, as it's already a block element.

### DIFF
--- a/mythril/analysis/templates/callgraph.html
+++ b/mythril/analysis/templates/callgraph.html
@@ -15,6 +15,9 @@
             color: #ffffff;
             font-size: 10px;
         }
+        html, body {
+          height: 95%;
+        }
     </style>
     {% else %}
     <style type="text/css">
@@ -26,6 +29,9 @@
             color: #000000;
             font-size: 10px;
             font-family: "courier new";
+        }
+        html, body {
+          height: 95%;
         }
     </style>
     {% endif %}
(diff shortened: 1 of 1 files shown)


norhh:
hi @dmuhs thanks a lot for the PR.

mythril/analysis/templates/callgraph.html
```diff
@@ -8,17 +8,22 @@
...
         #mynetwork {
+          height: 100%;
             background-color: #232625;
         }
         body {
             background-color: #232625;
             color: #ffffff;
             font-size: 10px;
         }
+        html, body {
```
norhh: These can be moved to outside the if else, but in this case it might not affect that much as it's just 1 condition.

PR #502, opened 2018-09-09
Fix Docker container crash with UnicodeEncodeError

### LINKED ISSUE #479: Mythril crashes with Unicode error on docker
## Description

Mythrl docker image crashes with unicode error for [this](https://etherscan.io/address/0x39e8082b00c9f19ebf553e40feb7cf8459acc693#code) contract. may be related to #448 but in this case for docker it is crashing at a different location.

## How to Reproduce
```console
$ sudo docker run -v $(pwd):/tmp mythril/myth -x /tmp/crash.sol
  File "/usr/local/bin/myth", line 4, in <module>
    __import__('pkg_resources').run_script('mythril==0.18.11', 'myth')
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 658, in run_script
    self.require(requires)[0].run_script(script_name, ns)
  File "/usr/lib/python3/dist-packages/pkg_resources/__init__.py", line 1438, in run_script
    exec(code, namespace, namespace)
  File "/usr/local/lib/python3.6/dist-packages/mythril-0.18.11-py3.6.egg/EGG-INFO/scripts/myth", line 9, in <module>
    mythril.interfaces.cli.main()
  File "/usr/local/lib/python3.6/dist-packages/mythril-0.18.11-py3.6.egg/mythril/interfaces/cli.py", line 212, in main
    print(outputs[args.outform])
UnicodeEncodeError: 'ascii' codec can't encode characters in position 439-447: ordinal not in range(128)
...
$
```

## Expected behavior

It should not crash.

## Environment
Please modify for your setup

- Mythril version: 0.18.11

dmuhs:
Fixes #479 

The report-generated output from the text and markdown functions returned text in UTF-8 encoding. The `print` function has an implicit encoding in ASCII. It tried to decode the bytes passed to it and crashed. The solution was to explicitly encode the input as UTF-8. This should also work when Mythril is run locally.

To make extra sure, I have set `PYTHONIOENCODING=UTF-8` in the Docker container's environment, which overrides the implicit print behaviour and encodes to UTF-8 by default.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -22,4 +22,6 @@ RUN apt-get update \
   && pip3 install -r requirements.txt \
   && python setup.py install
 
+ENV PYTHONIOENCODING=UTF-8
+
 ENTRYPOINT ["/usr/local/bin/myth"]
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,7 @@ COPY . /opt/mythril
 RUN apt-get update \
   && apt-get install -y \
      build-essential \
+     locales \
      python-pip-whl=9.0.1-2 \
      python3-pip=9.0.1-2 \
      python3-setuptools \
@@ -22,6 +23,9 @@ RUN apt-get update \
   && pip3 install -r requirements.txt \
   && python setup.py install
 
-ENV PYTHONIOENCODING=UTF-8
+RUN locale-gen en_US.UTF-8
+ENV LANG en_US.UTF-8
+ENV LANGUAGE en_US.en
+ENV LC_ALL en_US.UTF-8
 
 ENTRYPOINT ["/usr/local/bin/myth"]
--- a/mythril/interfaces/cli.py
+++ b/mythril/interfaces/cli.py
@@ -221,7 +221,7 @@ def main():
                     'text': report.as_text(),
                     'markdown': report.as_markdown()
                 }
-                print(outputs[args.outform].encode('utf-8'))
+                print(outputs[args.outform])
 
         elif args.statespace_json:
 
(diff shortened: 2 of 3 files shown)


norhh:
I think just in the docker, python is expecting ascii characters as input and output, else we might have encountered it in the non docker version which didn't happen(can you also confirm it as my pc's env might be different where it is working) . I think print defaults to unicode in python 3.5+ .

dmuhs:
After doing some more checking, Python seems to get the output encoding from the locales. Instead of just setting the environment variables specific to stdout, a better alternative would be to change the Dockerfile's locale settings to UTF-8. This will then not only switch Python's IO encoding, but also other components like the default filesystem encoding (allowing for UTF-8 characters in file names).

The current locale settings are:
```
root@37ccc196ab42:/# locale
LANG=
LANGUAGE=
LC_CTYPE="POSIX"
LC_NUMERIC="POSIX"
LC_TIME="POSIX"
LC_COLLATE="POSIX"
LC_MONETARY="POSIX"
LC_MESSAGES="POSIX"
LC_PAPER="POSIX"
LC_NAME="POSIX"
LC_ADDRESS="POSIX"
LC_TELEPHONE="POSIX"
LC_MEASUREMENT="POSIX"
LC_IDENTIFICATION="POSIX"
LC_ALL=
```
which also seems to be the reason why stdout encoding defaults to ASCII. On most dev machines this should not happen, as having an UTF-8 encoding in `LANG` or `LC_CTYPE` is fairly common.

This means that the encode statement in the above print function is redundant (and as @norhh pointed out, yes, it would destroy special characters like \t and \n). The best solution here seems to add the `locales`package as a dependency and then run `locale-gen` during the build.

dmuhs:
Just set the container locales. I have encountered a build error regarding the `eth-typing` dependency and fixed it by setting the required version range (as required by the `eth-utils` package) in the requirements file.

PR #503, opened 2018-09-10
Machinestack refactor

### LINKED ISSUE #491: Detect EVM stack overflow?
## Description

Detect EVM stack overflow

## Background

This was something mentioned in passing by Gonçalo.

norhh:
refactor machine stack and also fixes #491 
Edit: Also has a dependency on PR #494 (not a hard dependency but just to check it didn't mess up any tests)Only the last 2 commits are related to this PR(and also branch)


mythril/laser/ethereum/state.py
```diff
@@ -151,7 +188,7 @@ def memory_write(self, offset, data):
     def pop(self, amount=1):
```
JoranHonig: Does this function still work with stack being a Machine Stack

mythril/laser/ethereum/state.py
```diff
@@ -122,14 +126,44 @@ def as_dict(self):
                     calldata_type=self.calldata_type)
 
 
+class MachineStack(list):
```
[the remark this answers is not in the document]
norhh: done

mythril/laser/ethereum/exceptions.py
```diff
@@ -0,0 +1,14 @@
+class VmException(Exception):
```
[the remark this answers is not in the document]
norhh: done

mythril/laser/ethereum/instructions.py
```diff
@@ -13,14 +13,17 @@
...
-from mythril.laser.ethereum.exceptions import VmException, StackUnderflowException
+from mythril.laser.ethereum.evm_exceptions import VmException, StackUnderflowException
 from mythril.laser.ethereum.keccak import KeccakFunctionManager
 
 TT256 = 2 ** 256
 TT256M1 = 2 ** 256 - 1
 
 keccak_function_manager = KeccakFunctionManager()
 
+class StopSignal(Exception):
```
norhh: Is this being used?

JoranHonig: I don't think so, this is probably from before the TransactionEndSignal

JoranHonig: Confirmed, no current uses

norhh: removed :+1:

mythril/laser/ethereum/state.py
```diff
@@ -124,14 +126,64 @@ def as_dict(self):
...
+
 class MachineState:
     """
     MachineState represents current machine state also referenced to as \mu
     """
     def __init__(self, gas):
         """ Constructor for machineState """
         self.pc = 0
-        self.stack = []
+        self.stack = MachineStack([])
```
JoranHonig: We could just use the default variable value here right?

mythril/laser/ethereum/state.py
```diff
@@ -156,7 +208,7 @@ def memory_write(self, offset, data):
     def pop(self, amount=1):
         """ Pops amount elements from the stack"""
         if amount >= len(self.stack):
-            raise IndexError()
+            raise StackUnderflowException
```
JoranHonig: ```StackUnderflowException()```?

mythril/laser/ethereum/evm_exceptions.py
```diff
@@ -6,5 +6,8 @@ class StackUnderflowException(VmException):
     pass
 
 
-class StopSignal(Exception):
+class StackOverflowException(VmException):
```
JoranHonig: Can maybe make this:
```
StackOverflowException(VmException, IndexError)
```
That way the analysis modules and svm.py won't have to depend on the StackOverflowException, but can just keep on working with IndexErrors

tests/laser/state/mstack_test.py
```diff
@@ -0,0 +1,68 @@
...
+        mstack = MachineStack([])
+        mstack.append(0)
+        assert(mstack == [0])
+
+        for i in range(mstack.STACK_LIMIT-1):
+            mstack.append(1)
+
+        try:
+            mstack.append(1000)
+        except StackOverflowException:
```
JoranHonig: You can use pytest.raises here, like in mstate_tests

JoranHonig: Oh, these tests aren't written using pytest. I think pytests provides some really nice features, like the one mentioned here.
Let me know if you disagree, but lets write all the (new) tests using pytest (because uniformity + functionality)

tests/laser/state/mstack_test.py
```diff
@@ -0,0 +1,68 @@
+from mythril.laser.ethereum.state import MachineStack
+from mythril.laser.ethereum.evm_exceptions import *
+from tests import BaseTestCase
+
+
+class MachineStackTest(BaseTestCase):
```
JoranHonig: Can you structure the unit tests in the following way (when possible):
# Arrange
# Act
# Assert

Not saying that this is perfect, but it will allow us to stick to an easy format that is simple to read and semi-selfdocumenting. Plus in this case you see that a lot of the testcases actually have multiple testcases, I think it makes more sense to have a separate unittest for each test case

mythril/laser/ethereum/state.py
norhh: there is no need for an argument to be passed to the `StackUnderflowException` right? so i don't think it will be necessary here.

tests/laser/state/mstack_test.py
norhh: looks like pytest.raises works without any pytest stuff. I think using either of one is fine(both have their own list of good and bad), so let's stick to a convention and start using that from now on (pytest)

mythril/laser/ethereum/state.py
norhh: removed

mythril/laser/ethereum/instructions.py
```diff
@@ -861,7 +845,7 @@ def jump_(self, global_state):
         except AttributeError:
             logging.debug("Invalid jump argument (symbolic address)")
             return []
-        except IndexError:  # Stack Underflow
+        except StackUnderflowException:
```
JoranHonig: I'm going to do my best to merge https://github.com/ConsenSys/mythril/pull/514/ in a minute or so. 
In which case we can remove this except :+1:

tests/laser/state/mstack_test.py
JoranHonig: Good job :+1:

PR #512, opened 2018-09-17
Suicide implementation

JoranHonig:
This pr implements the suicide opcode


mythril/laser/ethereum/state.py
```diff
@@ -59,14 +60,16 @@ def __init__(self, address, code=None, contract_name="unknown", balance=None, co
         :param concrete_storage: Interpret storage as concrete
         """
         self.nonce = 0
-        self.code = code
+        self.code = code if code is not None else Disassembly("")
```
norhh: we can write "code or Disassembly("")" here right?

JoranHonig: yes

PR #519, opened 2018-09-20
fix dependency issues

norhh:


norhh:
aagh,this should have been done on a branch of ConsenSys to run integration tests.

norhh:
The check which failed was some TLS error related to the api

PR #534, opened 2018-09-26
Solver per branch

norhh:
It seems to be quite costly(memory wise) to maintain solver per branch.If it is possible to make it more optimized or I did some sort of memory leak here then it should probably be great :smiley: .


norhh:
looks like tests were killed due to the lack of memory of circleci.

mythril/analysis/solver.py
```diff
@@ -2,15 +2,13 @@
 from mythril.exceptions import UnsatError
 import logging
 
+
 def get_model(constraints):
-    s = Solver()
-    s.set("timeout", 100000)
 
-    for constraint in constraints:
-        s.add(constraint)
```
JoranHonig: Now constraints aren't added anymore

norhh: The constraint class automatically does that, so we may not need it.

mythril/laser/ethereum/state.py
```diff
@@ -129,6 +129,37 @@ def as_dict(self):
                     calldata_type=self.calldata_type)
 
 
+class Constraints(list):
```
JoranHonig: Docstrings? :smiley:

mythril/laser/ethereum/state.py
```diff
@@ -183,14 +214,14 @@ class MachineState:
     """
     MachineState represents current machine state also referenced to as \mu
     """
-    def __init__(self, gas):
+    def __init__(self, gas, pc=0, stack=[], memory=[], constraints=None, depth=0):
```
JoranHonig: There are mutable defaults here. You could use: memory = None
and later
self.memory = memory or []

norhh: oops, this I forgot, this PR was older than that one :sweat_smile:

JoranHonig: bump

JoranHonig: @norhh This is still mutable

norhh:
oops this PR was for master, will add a new PR to develop

PR #536, opened 2018-09-28
Add SWC ID to the analysis results

### LINKED ISSUE #526: Include SWC ID in analysis results
## Description

Include the [SWC ID](https://github.com/SmartContractSecurity/SWC-registry/) for each issue reported in the text, markdown and json outputs. This will make it easier to integrate with the SWC registry for benchmarking.

E.g.:

```
==== Unchecked CALL return value ====
SWC ID: 104
Type: Informational
Contract: ReturnValue
Function name: callnotchecked()
PC address: 294
The return value of an external call is not checked. Note that execution continue even if the called contract throws.
--------------------
In file: returnvalue.sol:5

function callnotchecked()  {
    callee.call();
  }

--------------------
```

Tagging @s0b0lev as it is relevant for SWC integration.

norhh:
fixes #526


mythril/analysis/swc_data.py
```diff
@@ -0,0 +1,23 @@
...
+REENTRANCY = '107'
+DEFAULT_STATE_VARIABLE_VISIBILITY = '108'
+UNINITIALIZED_STORAGE_POINTER = '109'
+ASSERT_VIOLATION = '110'
+DEPRICATED_FUNCTIONS_USAGE = '111'
+DELEGATECALL_TO_UNTRUSTED_CONTRACT = '112'
+MULTIPLE_SENDS = '113'
+TX_ORDER_DEPENDENCE = '114'
+TX_ORIGIN_USAGE = '115'
+TIMESTAMP_DEPENDENCE = '116'
```
norhh: @b-mueller is SWC 116 just time stamp dependence?

PR #538, opened 2018-09-29
Invalid opcode exception

JoranHonig:
Implementation of the invalid opcode handler, this is now correctly modeled as a VmException


JoranHonig:
I still need to add the correct VmException handling

norhh:
wait!, I just missed a part.let me check.

PR #542, opened 2018-10-01
Remove all broad exception catches in laser

### LINKED ISSUE #537: Remove all broad exceptions catches in Laser
## Description
Remove all broad exception catches in Laser.
It is quite terrible when the broad exception catches `NotImplementedError` and does something with that.

norhh:
fixes #537


mythril/laser/ethereum/call.py
```diff
@@ -63,6 +63,7 @@ def get_callee_address(global_state:GlobalState, dynamic_loader: DynLoader, symb
         # attempt to read the contract address from instance storage
         try:
             callee_address = dynamic_loader.read_storage(environment.active_account.address, index)
+        # TODO: verify whether this happens or not
         except:
```
JoranHonig: There is still a very broad exception catch here

mythril/laser/ethereum/util.py
```diff
@@ -10,6 +10,10 @@
 TT255 = 2 ** 255
 
 
+class ConcreteIntException(AttributeError):
```
JoranHonig: Why did you add this specific exception here? From the name its not immediately clear what kind of situation it is meant for and I think in this case that a Type error is more adequate for the situation

mythril/laser/ethereum/instructions.py
```diff
@@ -367,16 +366,15 @@ def calldataload_(self, global_state):
...
-                    val += environment.calldata[i].to_bytes(1, byteorder='big')
+                val = b''.join([calldata.to_bytes(1, byteorder='big') for calldata in
+                                environment.calldata[offset:offset+32]])
 
                 logging.debug("Final value: " + str(int.from_bytes(val, byteorder='big')))
                 state.stack.append(BitVecVal(int.from_bytes(val, byteorder='big'), 256))
-            # FIXME: broad exception catch
-            except:
+
+            except (util.ConcreteIntException, AttributeError):
```
JoranHonig: Isn't, ConcreteIntException also an AttributeError?

mythril/laser/ethereum/util.py
norhh: There are 2 exceptions possible here, so I thought lets return this, or should i return TypeError when a ValueError is raised?

mythril/laser/ethereum/instructions.py
norhh: here AttributeError comes from a different location.

mythril/laser/ethereum/call.py
norhh: It seems there are atleast 5-6 exceptions possible to be raised at for this catch, so how about shifting this broad exception to read_storage function which raises some sort of ErrorReadingStorage exception when it catches any exception?

mythril/laser/ethereum/util.py
JoranHonig: I'd like one of two situations:
Either we use a TypeError, because it describes the problem that happens. Namely we assume an element has a certain type, while it is not.
Or we do use a specific exception, but then I think it should have a name that describes what the problem is. ```ConcreteIntException``` isn't immediately clear in that aspect (at least to me)

PR #545, opened 2018-10-07
fix mythril signature db throwing FileNotFoundError when using mythril as a library

tintinweb:
Hi,

mythril signature db does not check if the folder structure is set up (e.g.`~/.mythril/`) before calling `signaturedb.write()`. This PR checks if the folder exists and if not mkdirs it. Autocreating the folder is expected behavior as one can explicitly call `signature.write()` even if the signaturedb file does not yet exist.

only happens when using mythril as a library. otherwise mythril would initially setup the folder.

this was reported in https://github.com/tintinweb/ethereum-dasm/issues/10 

cheers,
tin

### DIFF
--- a/mythril/support/signatures.py
+++ b/mythril/support/signatures.py
@@ -111,6 +111,8 @@ class SignatureDb(object):
         :return: self
         """
         path = path or self.signatures_file
+        directory = os.path.split(path)[0]
+
         if sync and os.path.exists(path):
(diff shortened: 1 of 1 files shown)


tintinweb:
unittests 1, tintin 0 :p

fixed: directory can be empty if the signature file path is only a filename in the cwd.

good to merge from my side @norhh

PR #549, opened 2018-10-08
Implement revert opcode

### LINKED ISSUE #403: Implement transaction finalizing ops

JoranHonig:
closes #403

### DIFF
--- a/mythril/laser/ethereum/svm.py
+++ b/mythril/laser/ethereum/svm.py
@@ -152,7 +152,7 @@ class LaserEVM:
             transaction, return_global_state = e.global_state.transaction_stack.pop()
 
             if return_global_state is None:
-                if not isinstance(transaction, ContractCreationTransaction) or transaction.return_data:
+                if (not isinstance(transaction, ContractCreationTransaction) or transaction.return_data) and not e.revert:
                     e.global_state.world_state.node = global_state.node
                     self.open_states.append(e.global_state.world_state)
                 new_global_states = []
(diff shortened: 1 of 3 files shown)


mythril/laser/ethereum/svm.py
```diff
@@ -161,7 +161,8 @@ def execute_state(self, global_state):
                 self._execute_post_hook(op_code, [e.global_state])
 
                 new_global_states = self._end_message_call(return_global_state, global_state,
-                                                           revert_changes=False, return_data=transaction.return_data)
+                                                           revert_changes=False or e.revert,
```
norhh: it might be good to rename the exception ‘e’ right?

norhh:
How about we start adding static typechecking for functions?

mythril/laser/ethereum/svm.py
JoranHonig: I'm not sure. In these cases I think that err or e is often used
To what would you want to rename it?

JoranHonig:
> How about we start adding static typechecking for functions?That might reduce some potential crashes in the future right?

Sure, however I think no all type hint functionality is supported in 3.5 so we should be careful with that

mythril/laser/ethereum/svm.py
norhh: Yes, but while looking at the PR, I had to go up to check what e was as i had no clue what it might be from the lines mentioned in github's diff .So in those cases it might help.for variable name something in lines of it being a transaction end signal might help.

dmuhs:
@JoranHonig I've just finished adding some type hints to the laser module that are conform to Python 3.5. Still have to fix 23 signatures and then we should have full hinting at least for laser.

JoranHonig:
> @JoranHonig I've just finished adding some type hints to the laser module that are conform to Python 3.5. Still have to fix 23 signatures and then we should have full hinting at least for laser.

For reference #571

PR #553, opened 2018-10-09
Fix a couple of issues with the solc-args option

maxme:
Now your can run it against a truffle project with the appropriate options.

To test it, create a new truffle project with some dependencies to openzeppelin and run:
```
$ myth --solc-args "openzeppelin-solidity=/Users/max/work/ponycontract/node_modules/openzeppelin-solidity --allow-paths /Users/max/work/ponycontract/node_modules/openzeppelin-solidity" \
-x contracts/ponycontract.sol
```

Previously it was failing due to the `import_from_solidity_source` method that was ignoring the `solc-args` and then it was failing due to multiple `--allow-paths` args provided to `solc`.

Note: I'm not proud of the argument cleanup function, I tried to find some utilities in the python library but didn't find any.

### DIFF
--- a/mythril/ether/util.py
+++ b/mythril/ether/util.py
@@ -18,10 +18,17 @@ def safe_decode(hex_encoded_string):
 
 def get_solc_json(file, solc_binary="solc", solc_args=None):
 
-    cmd = [solc_binary, "--combined-json", "bin,bin-runtime,srcmap,srcmap-runtime", '--allow-paths', "."]
+    cmd = [solc_binary, "--combined-json", "bin,bin-runtime,srcmap,srcmap-runtime"]
 
     if solc_args:
-        cmd.extend(solc_args.split(" "))
+        cmd.extend(solc_args.split())
+
+    if not "--allow-paths" in cmd:
+        cmd.extend(["--allow-paths", "."])
+    else:
+        for i, arg in enumerate(cmd):
+            if arg == "--allow-paths":
+                cmd[i + 1] += ",."
 
     cmd.append(file)
 
--- a/mythril/support/signatures.py
+++ b/mythril/support/signatures.py
@@ -183,7 +183,7 @@ class SignatureDb(object):
         :param file_path: solidity source code file path
         :return: self
         """
-        self.signatures.update(SignatureDb.get_sigs_from_file(file_path, solc_binary, solc_args))
+        self.signatures.update(SignatureDb.get_sigs_from_file(file_path, solc_binary=solc_binary, solc_args=solc_args))
         return self
 
     @staticmethod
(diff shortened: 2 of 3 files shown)


mythril/ether/util.py
```diff
@@ -18,10 +18,17 @@ def safe_decode(hex_encoded_string):
 
 def get_solc_json(file, solc_binary="solc", solc_args=None):
 
-    cmd = [solc_binary, "--combined-json", "bin,bin-runtime,srcmap,srcmap-runtime", '--allow-paths', "."]
+    cmd = [solc_binary, "--combined-json", "bin,bin-runtime,srcmap,srcmap-runtime"]
 
     if solc_args:
-        cmd.extend(solc_args.split(" "))
+        cmd.extend([i for i in solc_args.split(" ") if i != ""])
```
norhh: I think you can remove the " " in the solc_args.split(" "), and make it `solc_args.split()` and then you don't need the check for `i != ""` .

mythril/support/signatures.py
```diff
@@ -206,13 +206,15 @@ def lookup_online(sighash, timeout=None, proxies=None):
...
+    def get_sigs_from_file(file_name, solc_binary="solc", solc_args=None):
         """
         :param file_name: accepts a filename
         :return: their signature mappings
         """
         sigs = {}
-        cmd = ["solc", "--hashes", file_name]
+        cmd = [solc_binary, "--hashes", file_name]
+        if solc_args:
+            cmd.extend([i for i in solc_args.split(" ") if i != ""])
```
norhh: same as above

mythril/support/signatures.py
```diff
@@ -177,13 +177,13 @@ def __getitem__(self, item):
...
 
-    def import_from_solidity_source(self, file_path):
+    def import_from_solidity_source(self, file_path, solc_binary="solc", solc_args=None):
         """
         Import Function Signatures from solidity source files
         :param file_path: solidity source code file path
         :return: self
         """
-        self.signatures.update(SignatureDb.get_sigs_from_file(file_path))
+        self.signatures.update(SignatureDb.get_sigs_from_file(file_path, solc_binary, solc_args))
```
norhh: can you also add the respective args names here, it might be useful for the future.

norhh:
:+1:  looks great overall, thanks a lot @maxme , can you look into those few comments.

mythril/ether/util.py
maxme: Good point, I haven't written Python code in a while :) - updated in 6fd8b76900bc2098e6007b126e5eb2995f43b6d1 (rebased)

mythril/support/signatures.py
maxme: Updated in 6fd8b76900bc2098e6007b126e5eb2995f43b6d1

PR #554, opened 2018-10-09
Fix mutable args

dmuhs:
This fixes a common Python gotcha. Default arguments are evaluated only once when the function is *defined*. Mutable default arguments thus are declared once in memory and referenced over and over again when used. A simple example:

```python
In [9]: def append_to_list(value, l=[]):
   ...:     l.append(value)
   ...:     return l
   ...: 

In [10]: append_to_list(42)
Out[10]: [42]

In [11]: append_to_list(1337)
Out[11]: [42, 1337]
```

On the second function call you might have expected `[1337]` to return, however what you get is the mutable default type's appended version. This might introduce some subtle bugs where e.g. more states are in a list (and evaluated) than there actually are. The solution is to simply define a new list at runtime if no argument is given.

### DIFF
--- a/mythril/laser/ethereum/state.py
+++ b/mythril/laser/ethereum/state.py
@@ -135,8 +135,10 @@ class MachineStack(list):
     """
     STACK_LIMIT = 1024
 
-    def __init__(self, default_list=[]):
+    def __init__(self, default_list=None):
         super(MachineStack, self).__init__(default_list)
+        if default_list is None:
+            default_list = []
 
     def append(self, element):
         """
--- a/mythril/laser/ethereum/state.py
+++ b/mythril/laser/ethereum/state.py
@@ -136,9 +136,9 @@ class MachineStack(list):
     STACK_LIMIT = 1024
 
     def __init__(self, default_list=None):
-        super(MachineStack, self).__init__(default_list)
         if default_list is None:
             default_list = []
+        super(MachineStack, self).__init__(default_list)
 
     def append(self, element):
         """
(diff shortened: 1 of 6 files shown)


mythril/laser/ethereum/state.py
```diff
@@ -135,7 +135,9 @@ class MachineStack(list):
     """
     STACK_LIMIT = 1024
 
-    def __init__(self, default_list=[]):
+    def __init__(self, default_list=None):
+        if default_list is None:
+            default_list = []
```
norhh: We don’t need that here right? As default_list doesn’t change here and the constructor is of one time usage for the class instance.correct me if I am wrong

dmuhs: `default_list=[]` in the function header is the same data structure globally - across all `MachineStack` objects. An example to show the dependencies across objects:
```python
In [3]: class Foo(object):
   ...:     def __init__(self, default_list=[]):
   ...:         self.my_list = default_list
   ...:     def append(self, v):
   ...:         self.my_list.append(v)
   ...:         
In [4]: f = Foo()

In [5]: f.append(42)

In [6]: f.my_list
Out[6]: [42]

In [7]: f2 = Foo()

In [8]: f2.append(1337)

In [9]: f2.my_list
Out[9]: [42, 1337]
```

As you can see, even though we have two separate instances, the state of `f` is dragged into `f2`, because the default arguments of `Foo.__init__` is evaluated only once - when the class is defined. Basically all instances of `Foo` work on the same list pointer. So it's not a problem limited to the constructor, but to everything that touches `default_list` directly or indirectly.

norhh: Reasons why I hate python seems to keep on growing

dmuhs: It's a pretty language once you've figured out its kinks.. :laughing:

PR #555, opened 2018-10-09
Remove unused variables and fix name shadowing

dmuhs:
While going through the code I've noticed some occurrences of inner and outer scope name shadowing of variable names and some instances where local variables weren't used at all. This PR fixes some of those cases. In others, like methods that still have to be implemented, I've left them.


mythril/interfaces/cli.py
```diff
@@ -17,8 +17,8 @@
 from mythril.version import VERSION
 
 
-def exit_with_error(format, message):
-    if format == 'text' or format == 'markdown':
+def exit_with_error(fmt, message):
```
JoranHonig: Why this change? Isn't format more readable in this case?

mythril/leveldb/state.py
```diff
@@ -43,9 +43,9 @@ class Account(rlp.Serializable):
         ('code_hash', hash32)
     ]
 
-    def __init__(self, nonce, balance, storage, code_hash, db, address):
+    def __init__(self, nonce, balance, storage, code_hash, db, addr):
```
JoranHonig: Same here, why address -> addr

mythril/interfaces/cli.py
dmuhs: In this case format is also the name of a python built-in function: https://docs.python.org/3/library/functions.html#format

mythril/leveldb/state.py
dmuhs: The local name address shadows an imported thingie that's also called address

mythril/interfaces/cli.py
JoranHonig: Can we do format_ vs fmt?

norhh: or how about `format_type`

mythril/laser/ethereum/instructions.py
```diff
@@ -710,11 +709,8 @@ def mstore_(self, global_state):
             # Attempt to concretize value
             _bytes = util.concrete_int_to_bytes(value)
 
-            i = 0
-
-            for b in _bytes:
-                state.memory[mstart + i] = _bytes[i]
-                i += 1
+            for i, b in enumerate(_bytes):
+                state.memory[mstart + i] = b
```
norhh: we can assign a slice of `state.memory` as _bytes right? 
something like `state.memory[mstart: mstart+len(_bytes)] = _bytes`

mythril/interfaces/cli.py
JoranHonig: Not saying its the optimal solution. But in my defense, using a trailing underscore is the pep8 proposed way to prevent shadowing: https://www.python.org/dev/peps/pep-0008/#descriptive-naming-styles

dmuhs: Wow, didn't know that! Well, we (usually) shouldn't argue with PEP8, so I'll change that back with the rest of the additions. Thanks for going out of your way to look that up :+1:

JoranHonig: I'm totally fine with format_type in this case :smile:

dmuhs: Too late. Both solutions are equally good anyway and just as readable. :laughing:

norhh:
hey, can you resolve those merge conflicts?

dmuhs:
@norhh Should all be fixed. Just some stray single-quote docstrings :+1:

dmuhs:
Merged the most recent changes of master into the branch. Surprisingly, the renames from the last refactoring were auto-merged. :+1:

Issue #527, opened 2018-09-21
Calldatasize constraints


JoranHonig:
When there is a constraint on some ```calldata[x] != 0``` then there should be a constraint that the ```calldatasize > x```

Take the example by Chofhe from discord( @nbanmp ):
```
contract Suicide {
  uint256 isAllowed = 0;

  function kill(address addr, uint256 isAllowed) {
    require(isAllowed == 42);
    require(msg.data.length == 4);
    selfdestruct(addr);
  }
}
```

Where mythril now reports a false positive

## Proposed solution
Add the following constraint on calldataload:
```
# given variables: index, caldataload[index], calldatasize
Or(calldataload[index] == 0, index < calldatasize)
```

### Good first issue
This issue is labeled as a good first issue. Hop by on our discord server if you'd like to contribute!

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 0.6 ETH (137.55 USD @ $229.25/ETH)  attached to it__ as part of the Mythril fund__.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ConsenSys/mythril/527/1400).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ConsenSys/mythril/527/1400).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $48,292.36 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [nbanmp](https://gitcoin.co/profile/nbanmp) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ConsenSys/mythril/527/1400?mutate_worker_action=approve&worker=nbanmp) | [reject worker](https://gitcoin.co/issue/ConsenSys/mythril/527/1400?mutate_worker_action=reject&worker=nbanmp))_.**

Complete #528, then add constraint.

[...]

PR #557, opened 2018-10-09
calldataload indexing (closes #528)

### LINKED ISSUE #528: Calldataload indexing
Currently calldataload creates a symbol called ```calldata[index]``` while what it should create is a concatination of ```calldata[index] ... calldata[index + 31]``` according to the yellowpaper.

This can cause false positives when mythril can't recognize that msg.data[37] is the second byte in the result from calldataload(36).

### Good first issue
This issue is labeled as a good first issue ( not super simple though). Hop by on our discord server if you'd like to contribute!

nbanmp:


mythril/laser/ethereum/transaction/concolic.py
```diff
@@ -10,13 +10,14 @@ def execute_message_call(laser_evm, callee_address, caller_address, origin_addre
     open_states = laser_evm.open_states[:]
     del laser_evm.open_states[:]
 
+    next_transaction_id = get_next_transaction_id()
```
JoranHonig: Each transaction should have a unique transaction id, so this should be moved up to the for loop

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,44 @@
...
+            if item.step != None \
+            or item.start > item.stop \
+            or item.start < 0 \
+            or item.stop < 0: raise IndexError("Invalid Calldata Slice")
+
+            dataparts = []
+            for i in range(item.start, item.stop):
+                dataparts.append(self[i])
+            return simplify(Concat(dataparts))
+        else:
```
JoranHonig: This else isn't necessary, so to reduce nesting you can do the following:

```python3
def __getitem__(self, item: int):
    if isinstance(item, slice):
        if item.step != None \
            or item.start > item.stop \
            or item.start < 0 \
            or item.stop < 0: raise IndexError("Invalid Calldata Slice")

        dataparts = []
        for i in range(item.start, item.stop):
            dataparts.append(self[i])
        return simplify(Concat(dataparts))


    try:
        return self._calldata[item]
    except KeyError:
        self._calldata[item] = BitVec(str(self.tx_id) + '_calldata[' + str(item) + ']', 8)
        return self._calldata[item]
```

mythril/laser/ethereum/instructions.py
```diff
@@ -353,37 +353,42 @@ def calldataload_(self, global_state):
...
-            state.stack.append(global_state.new_bitvec(
-                "calldata_" + str(environment.active_account.contract_name) + "[" + str(simplify(op0)) + "]", 256))
-            return [global_state]
-        except IndexError:
-            logging.debug("Calldata not set, using symbolic variable instead")
-            state.stack.append(global_state.new_bitvec(
-                "calldata_" + str(environment.active_account.contract_name) + "[" + str(simplify(op0)) + "]", 256))
-            return [global_state]
+        if environment.calldata_type == CalldataType.CONCRETE:
+            try:
```
JoranHonig: Can you rewrite the code so that there is only one check for a symbolic index.

mythril/laser/ethereum/instructions.py
```diff
@@ -353,37 +353,42 @@ def calldataload_(self, global_state):
...
+                offset = util.get_concrete_int(simplify(op0))
+                b = environment.calldata[offset]
+            except AttributeError:
+                logging.debug("CALLDATALOAD: Unsupported symbolic index")
+                state.stack.append(global_state.new_bitvec(
+                    "calldata_" + str(environment.active_account.contract_name) + "[" + str(simplify(op0)) + "]", 256))
+                return [global_state]
+            except IndexError:
+                logging.debug("Calldata not set, using symbolic variable instead")
+                state.stack.append(global_state.new_bitvec(
```
JoranHonig: This behavior was also wrong before. If an index is accessed outside of the calldata "bounds" then 0 should be returned

Can you change that as well?

mythril/laser/ethereum/instructions.py
```diff
@@ -353,37 +353,42 @@ def calldataload_(self, global_state):
...
-            logging.debug("CALLDATALOAD: Unsupported symbolic index")
-            state.stack.append(global_state.new_bitvec(
-                "calldata_" + str(environment.active_account.contract_name) + "[" + str(simplify(op0)) + "]", 256))
-            return [global_state]
-        except IndexError:
-            logging.debug("Calldata not set, using symbolic variable instead")
-            state.stack.append(global_state.new_bitvec(
-                "calldata_" + str(environment.active_account.contract_name) + "[" + str(simplify(op0)) + "]", 256))
-            return [global_state]
+        if environment.calldata_type == CalldataType.CONCRETE:
```
JoranHonig: Besides the code for symbolic index checking, can't most of the logic be done in the SymbolicCalldata object

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,44 @@
...
 from random import randint
 
 from mythril.laser.ethereum.evm_exceptions import StackOverflowException, StackUnderflowException
 
-
 class CalldataType(Enum):
     CONCRETE = 1
     SYMBOLIC = 2
 
+class SymbolicCalldata:
```
JoranHonig: Can we rename this to Calldata, and implement logic for both concrete + symbolic calldata.
That way we can do away with the calldatatype

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,44 @@
...
+    def __init__(self, tx_id: int, starting_calldata: bytes=None):
+        self.tx_id = tx_id
+        self._calldata = {}
+
+        if starting_calldata:
+            for i in range(len(starting_calldata)):
```
norhh: we can use python's enumerate here right?

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,44 @@
...
+    def __getitem__(self, item: int):
+        if isinstance(item, slice):
+            if item.step != None \
+            or item.start > item.stop \
+            or item.start < 0 \
+            or item.stop < 0: raise IndexError("Invalid Calldata Slice")
+
+            dataparts = []
+            for i in range(item.start, item.stop):
+                dataparts.append(self[i])
```
norhh: I think we can compress the above 3 lines to something like `dataparts = [self[i] for i in range(item.start, item.stop)]`

nbanmp:
Oops. An unwanted pull was in there somewhere. I'll fix that...

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,62 @@
...
+        self.state = state
+        for i in range(len(self.starting_calldata)):
+            self.state.mstate.constraints.append(self._calldata[BitVecVal(i, 256)] == BitVecVal(self.starting_calldata[i], 8))
+
+    def concretized(self, model):
+        concrete_calldata = model[self._calldata].as_list()
+        concrete_calldata.sort(key=lambda x: x[0].as_long() if type(x) == list else -1)
+        result = []
+        arr_index = 1
+        for i in range(0, concrete_calldata[len(concrete_calldata)-1][0].as_long()+1):
```
norhh: i think 0 is not essential here

mythril/analysis/modules/ether_send.py
```diff
@@ -111,10 +111,10 @@ def execute(statespace):
                 try:
                     model = solver.get_model(node.constraints)
 
-                    for d in model.decls():
-                        logging.debug("[ETHER_SEND] main model: %s = 0x%x" % (d.name(), model[d].as_long()))
+                    pretty_model = solver.pretty_print_model(model)
+                    logging.debug(pretty_model)
```
norhh: I think it would be good to also have the model name here to debug.

nbanmp:
I am unsure why `tests/native_test.py::NativeTests::runTest` is failing. That test case is working for me locally. With regards to the `tests/svm_test.py::SVMTestCase::test_laser_result` test case, which is timing out, my change to calldata does make mythril noticeably slower, and if the timeout duration was increased to say 20 minutes, that test case would likely pass.

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,62 @@
...
 from random import randint
 
 from mythril.laser.ethereum.evm_exceptions import StackOverflowException, StackUnderflowException
 
-
 class CalldataType(Enum):
     CONCRETE = 1
     SYMBOLIC = 2
 
+class Calldata:
```
norhh: shouldn't __len__() also be implemented for this class as calldatasize opcode uses it?

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,62 @@
...
+    def __init__(self, tx_id, starting_calldata: bytes=None):
+        self.tx_id = tx_id
+        self._calldata = Array('{}_calldata'.format(self.tx_id), BitVecSort(256), BitVecSort(8))
+        self.starting_calldata = starting_calldata if starting_calldata else []
+
+    def set_global_state(self, state):
+        self.state = state
+        for i in range(len(self.starting_calldata)):
+            self.state.mstate.constraints.append(self._calldata[BitVecVal(i, 256)] == BitVecVal(self.starting_calldata[i], 8))
```
norhh: I think calldata is typically list of symbols and integers, so this might fail in some circumstances when we are trying to convert a BitVecVal to a symbol.

norhh:
> I am unsure why `tests/native_test.py::NativeTests::runTest` is failing. That test case is working for me locally. With regards to the `tests/svm_test.py::SVMTestCase::test_laser_result` test case, which is timing out, my change to calldata does make mythril noticeably slower, and if the timeout duration was increased to say 20 minutes, my guess is that test case would likely pass.

Thats weird, in my PC sha256 tests were failing. For some reason I have a feeling that the TypeError being raised there might be catching the exception which was expected, but i am not totally sure.

I am not sure of the behaviour of calldata when bytes() is called, can you check into that?

nbanmp:
Thanks. I figured out the issue with natives. The next issues to work out involve times when data is copied from memory to calldata, e.g. for an internal call, and probably finding some hack to work around the memory byte indexing issue. (https://github.com/ConsenSys/mythril/issues/563)

nbanmp:
I'm seeing some confusing behavior with the testcases that are failing after I pushed the modification to `evm_test.py`. It seems to be working locally for me, but it obviously has caused all of its test cases to error.

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,88 @@
...
+
+        self.starting_calldata = starting_calldata if starting_calldata else []
+
+    def set_global_state(self, state):
+        self.state = state
+        constraints = []
+
+
+        if self.concrete:
+            for i in range(len(self.starting_calldata)):
```
norhh: Here you can just iterate through the call data right?

nbanmp:
This PR is ready for review now. I've double checked the `calldatasize` constraints, those are now working as expected (#527).

mythril/laser/ethereum/state.py
norhh: hey @nbanmp, did you miss this comment, as i didn't see you responding to this :sweat_smile: .

norhh:
I was testing it various contracts and found that this seems to be breaking with dependence_on_predictable_vars module for some contracts like this `0x4fe3c4e0c8fcf9ef23b1391cad0205b3e338e013` due to some type issues, can you check it out.some initial investigation says that the model's solver is giving out lambda objects due to Implies being used now which isn't being accounted there.

mythril/laser/ethereum/state.py
nbanmp: :sweat_smile: I misread something and thought that is was a non-issue anymore... Fixing.

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,80 @@
...
+
+    def __getitem__(self, item):
+        if isinstance(item, slice):
+            MAX_SLICE = 1024
+            try:
+                current_index = item.start if type(item.start) in [BitVecRef, BitVecNumRef] else BitVecVal(item.start, 256)
+                dataparts = []
+                i = 0
+                while simplify(current_index != item.stop):
+                    dataparts.append(self[current_index])
```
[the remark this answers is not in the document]
nbanmp: Sure, I'll add something like that if the calldatasize is concrete. (I'll grab the calldatasize directly instead of finding the length)

I can't remember why I added the MAX_SLICE check... I think that was from when I was worried about things taking a long time, but I don't think that it is needed either, and it's obviously not being reached, so I'll remove that too.

nbanmp:
@norhh I figured out the problem with that when I was working on this and totally forgot about it, thanks for the reminder.

The issue is with the `FuncInterp` objects in the model that are in the model when an Array is concretized. I just need to update a whichever modules try to loop through the model them out without using the `pretty_print_model` from `solver.py`.

mythril/laser/ethereum/call.py
```diff
@@ -122,17 +122,32 @@ def get_call_data(global_state, memory_start, memory_size, pad=True):
...
     """
     state = global_state.mstate
+    transaction_id = '{}_internalcall'.format(global_state.current_transaction.id)
     try:
         # TODO: This only allows for either fully concrete or fully symbolic calldata.
         # Improve management of memory and callata to support a mix between both types.
-        call_data = state.memory[util.get_concrete_int(memory_start):util.get_concrete_int(memory_start + memory_size)]
-        if len(call_data) < 32 and pad:
-            call_data += [0] * (32 - len(call_data))
+        calldata_from_mem = state.memory[util.get_concrete_int(memory_start):util.get_concrete_int(memory_start + memory_size)]
```
JoranHonig: How about something like this?
```
starting_calldata = [state.memory[i] for i in range(memory_start, memory_start + memory_size)]:
```

mythril/laser/ethereum/call.py
```diff
@@ -122,17 +122,32 @@ def get_call_data(global_state, memory_start, memory_size, pad=True):
...
+        i = 0
+        starting_calldata = []
+        while i < len(calldata_from_mem):
+            elem = calldata_from_mem[i]
```
JoranHonig: calldata_from_mem can also be a BitVec here, right?

mythril/laser/ethereum/call.py
JoranHonig: Of with concrete variables ^^

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+from mythril.laser.ethereum.util import get_concrete_int
 
 from mythril.laser.ethereum.evm_exceptions import StackOverflowException, StackUnderflowException
 
-
 class CalldataType(Enum):
     CONCRETE = 1
     SYMBOLIC = 2
 
+class Calldata:
```
JoranHonig: Can you add docstrings?

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+        if starting_calldata:
+            self._calldata = []
+            self.calldatasize = BitVecVal(len(starting_calldata), 256)
+            self.concrete = True
+        else:
+            self._calldata = Array('{}_calldata'.format(self.tx_id), BitVecSort(256), BitVecSort(8))
+            self.calldatasize = BitVec('{}_calldatasize'.format(self.tx_id), 256)
+            self.concrete = False
+
+        self.starting_calldata = starting_calldata if starting_calldata else []
```
JoranHonig:
```suggestion
        self.starting_calldata = starting_calldata or []
```

mythril/laser/ethereum/state.py
JoranHonig: Why was this resolved?

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+
+    def set_global_state(self, state):
+        self.state = state
+        constraints = []
+
+        if self.concrete:
+            for cd_byte in self.starting_calldata:
```
JoranHonig: Can you rename it to calldata_byte, imo a bit more readable

JoranHonig: What do you think about:
```python
self._calldata = [BitVecVal(calldata_byte, 8) if instanceof(calldata_byte, int) else calldata_byte for calldata_byte in self.starting_calldata]
```

JoranHonig: Also, why not do this part in the constructor? Seems to me that this is part of the initialization logic

mythril/laser/ethereum/state.py
norhh: He is storing calldatasize as a variable and is accessing it in that opcode, but now i am thinking it might be cleaner to have \_\_len\_\_() over having to use the class variables, what do you think?

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+            self.calldatasize = BitVecVal(len(starting_calldata), 256)
+            self.concrete = True
+        else:
+            self._calldata = Array('{}_calldata'.format(self.tx_id), BitVecSort(256), BitVecSort(8))
+            self.calldatasize = BitVec('{}_calldatasize'.format(self.tx_id), 256)
+            self.concrete = False
+
+        self.starting_calldata = starting_calldata if starting_calldata else []
+
+    def set_global_state(self, state):
```
JoranHonig: What you want to do is add constraints to the global_state right?
However, by building it like this the calldata becomes aware/dependent on GlobalState. 
Can you maybe build it where Calldata provides a function:
```python3
@property
def constraints(self):
    constraints = []
    ...
    if ...:
        x = BitVec('x', 256)
        constraints.append(ForAll(x, Implies(self[x] != 0, UGT(self.calldatasize, x))))
    return constraints
```

And the caller (who already is dependent on GlobalState) does:
globalstate.mstate.constraints.extend(calldata.constraints)

That way the calldata code is simpler and not dependent anymore

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+                else:
+                    self._calldata.append(cd_byte)
+            constraints.append(self.calldatasize == len(self.starting_calldata))
+        else:
+            x = BitVec('x', 256)
+            constraints.append(ForAll(x, Implies(self[x] != 0, UGT(self.calldatasize, x))))
+
+        self.state.mstate.constraints.extend(constraints)
+
+    def concretized(self, model):
```
JoranHonig: This is good, but it would also be possible with yield (requires less boilerplate)

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+        result = []
+        for i in range(get_concrete_int(model.eval(self.calldatasize, model_completion=True))):
+            result.append(get_concrete_int(model.eval(self[i], model_completion=True)))
+
+        return result
+
+    def get_word_at(self, index: int):
```
JoranHonig: Thanks ^^

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+                dataparts = []
+                while simplify(current_index != stop):
+                    dataparts.append(self[current_index])
+                    current_index = simplify(current_index + 1)
+            except Z3Exception:
+                raise IndexError("Invalid Calldata Slice")
+
+            return simplify(Concat(dataparts))
+
+        if self.concrete:
```
JoranHonig: Given concrete calldata initialized with [1, 3, 5]
What happens if I do:
calldata[10] ?
It seems like this will return an index error, while it should return 0

mythril/laser/ethereum/state.py
nbanmp: I haven't been paying attention to pep8 when coding this... I could swap out my code with a line like that though, but it's much longer than the 79 char limit in pep8.

It's not in the constructor because it is dependent on global state. I like the constraints idea below better.

mythril/laser/ethereum/state.py
nbanmp: Ah, that's what I was looking for when coding this. No clue how I missed this much better option.

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+    def __getitem__(self, item):
+        if isinstance(item, slice):
+            stop = item.stop
+            if self.concrete and type(stop) == int:
+                calldatasize = get_concrete_int(self.calldatasize)
+                if stop > calldatasize:
+                    stop = calldatasize
+
+            try:
+                current_index = item.start if type(item.start) in [BitVecRef, BitVecNumRef] else BitVecVal(item.start, 256)
```
JoranHonig:
```suggestion
                current_index = item.start if instanceof(item.start, BitVecRef) else BitVecVal(item.start, 256)
```

mythril/laser/ethereum/state.py
norhh: He added a variable calldatasize which he used in that opcode, now it makes me think that using \_\_len\_\_() might be cleaner as we don't need to remember that such a variable exists in the class, what do you think?

mythril/laser/ethereum/state.py
```diff
@@ -1,16 +1,82 @@
...
+    def get_word_at(self, index: int):
+        return self[index:index+32]
+
+    def __getitem__(self, item):
+        if isinstance(item, slice):
+            stop = item.stop
+            if self.concrete and type(stop) == int:
+                calldatasize = get_concrete_int(self.calldatasize)
+                if stop > calldatasize:
+                    stop = calldatasize
```
JoranHonig: Why dis? :upside_down_face:

mythril/laser/ethereum/transaction/concolic.py
```diff
@@ -11,12 +11,13 @@ def execute_message_call(laser_evm, callee_address, caller_address, origin_addre
...
     for open_world_state in open_states:
+        next_transaction_id = get_next_transaction_id()
         transaction = MessageCallTransaction(
-            identifier=get_next_transaction_id(),
+            identifier=next_transaction_id,
             world_state=open_world_state,
             callee_account=open_world_state[callee_address],
             caller=caller_address,
-            call_data=data,
+            call_data=Calldata(next_transaction_id, data),
```
JoranHonig: Shouldn't there be some set_global_state() happening somewhere?

mythril/laser/ethereum/transaction/symbolic.py
```diff
@@ -23,7 +23,7 @@ def execute_message_call(laser_evm, callee_address):
             callee_account=open_world_state[callee_address],
             caller=BitVec("caller{}".format(next_transaction_id), 256),
             identifier=next_transaction_id,
-            call_data=[],
+            call_data=Calldata(next_transaction_id),
```
JoranHonig: Shouldn't there be some set_global_state() happening somewhere?

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -44,7 +44,7 @@ def __init__(self,
         self.world_state = world_state
         self.callee_account = callee_account
         self.caller = caller
-        self.call_data = call_data
+        self.call_data = Calldata(self.id) if call_data is None else Calldata(self.id, call_data) if type(call_data) == list else call_data
```
JoranHonig: I feel like this might be getting a little bit too long of a one liner, wdyt?
Why should a list and Calldata object be admissible? Why not only Calldata?

tests/report_test.py
```diff
@@ -38,7 +38,7 @@ def _generate_report(input_file):
 def reports():
     """ Fixture that analyses all reports"""
     pool = Pool(cpu_count())
-    input_files = sorted([f for f in TESTDATA_INPUTS.iterdir()])
+    input_files = sorted([f for f in TESTDATA_INPUTS.iterdir() if f.name != 'environments.sol.o'])
```
JoranHonig: Why does it need to be disabled here?

JoranHonig: The jsonification?

mythril/laser/ethereum/call.py
nbanmp: I'm pretty sure it's always a list...
```
 calldata_from_mem = state.memory[util.get_concrete_int(memory_start):util.get_concrete_int(memory_start + memory_size)] 
```

JoranHonig:
Can the environment.calldatasize be removed?

mythril/laser/ethereum/call.py
nbanmp: This will not work, because memory isn't always byte indexed, so I need some hack to catch 256 bit BitVecs that are in memory.

mythril/laser/ethereum/instructions.py
```diff
@@ -448,11 +413,15 @@ def calldatacopy_(self, global_state):
                 return [global_state]
 
             try:
-                i_data = environment.calldata[dstart]
+                i_data = dstart
 
+                new_memory = []
                 for i in range(mstart, mstart + size):
```
JoranHonig:
```suggestion
                for i in range(size):
```

mythril/laser/ethereum/instructions.py
```diff
@@ -448,11 +413,15 @@ def calldatacopy_(self, global_state):
...
-                    state.memory[i] = environment.calldata[i_data]
-                    i_data += 1
+                    new_memory.append(environment.calldata[i_data])
+                    i_data = simplify(i_data + 1)
```
JoranHonig: this will break if i_data is an int right?

JoranHonig: Or can that never happen?

mythril/laser/ethereum/state.py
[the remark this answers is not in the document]
norhh: this will still result in the same problem right? as the the index might not be an int.

mythril/laser/ethereum/state.py
JoranHonig: Can you add unit tests for Calldata?

mythril/laser/ethereum/state.py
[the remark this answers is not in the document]
norhh: this will still result in the same problem right? as the index may not be an int(say some BitVecNum or something of that sort which is in a way an int) ?

mythril/laser/ethereum/call.py
nbanmp: This will not work, because memory isn't always byte indexed, so I need some hack to catch 256 bit BitVecs that are in memory.

mythril/laser/ethereum/state.py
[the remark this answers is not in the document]
norhh: this might still result in the same problem right? as the index can be some sort of BitVecNum or something which is `int` in a way?

mythril/laser/ethereum/transaction/concolic.py
nbanmp: It's happening. In the `initial_global_state` function of MessageCallTransaction.

mythril/laser/ethereum/transaction/symbolic.py
nbanmp: It's happening. In the `initial_global_state` function of MessageCallTransaction.

mythril/laser/ethereum/transaction/transaction_models.py
nbanmp: I'll shorten it to:
```
self.call_data = Calldata(self.id) or Calldata(self.id, call_data) if type(call_data) == list else call_data
```

If calldata is passed in as a list, I assume the concrete calldata bytes are being passed in.

tests/report_test.py
nbanmp: Yup.

mythril/laser/ethereum/instructions.py
nbanmp: Great catch. I think it might be able to happen.

mythril/laser/ethereum/transaction/symbolic.py
nbanmp: It's happening. In the initial_global_state function of MessageCallTransaction.

mythril/laser/ethereum/transaction/transaction_models.py
nbanmp: I assume that a list passed in is the concrete calldata.

I think it's also long, but a bunch of the one liners there are also really long. It's only a few characters longer than the next longest.
If someone is going to fix pep8, I think they can probably deal with it in conjunction with all of the other lines following it, and if not, the line fits in pretty well with all the other ternary lines following it.

tests/report_test.py
nbanmp: Yup

PS: I made this comment like 3 times... github is dying on me... :(

mythril/laser/ethereum/instructions.py
nbanmp: Good catch. I think it can happen.

mythril/laser/ethereum/state.py
nbanmp: I think it would be nice nice to have an array rather than a generator returned.

mythril/laser/ethereum/state.py
nbanmp: Fixed.

mythril/laser/ethereum/state.py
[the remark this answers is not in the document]
nbanmp: You're right, item.stop could be a BitVecVal. Good catch.

mythril/laser/ethereum/state.py
nbanmp: That does sound cleaner. @norhh I'll implement it that way.

nbanmp: @norhh Just kidding. I believe that in Python, `__len__` must return an integer, and I need to return a BitVec sometimes.

mythril/laser/ethereum/state.py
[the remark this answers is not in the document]
nbanmp: Alright, it should be good now.

mythril/laser/ethereum/state.py
nbanmp: Commit https://github.com/ConsenSys/mythril-classic/pull/557/commits/34daeebe693a7ff5f328a69fd754429f916d6cd5

mythril/laser/ethereum/call.py
JoranHonig: nvm, I had a typeconfusion bug in my brain

mythril/laser/ethereum/call.py
```diff
@@ -122,17 +122,32 @@ def get_call_data(global_state, memory_start, memory_size, pad=True):
...
+            if isinstance(elem, int):
+                starting_calldata.append(elem)
```
JoranHonig: what if elem is bigger then (2^8) - 1

mythril/laser/ethereum/call.py
JoranHonig: ok, for documentation: with https://github.com/ConsenSys/mythril-classic/issues/563 the solution would work

mythril/laser/ethereum/transaction/transaction_models.py
JoranHonig: I kind of disagree, although the following lines could be shortened, the logic is quite simple
But this line has a nested if statement, and isn't simple anymore.
I'd either rewrite it as a normal if statement or limit the input to __init__ to either None or a list

mythril/laser/ethereum/call.py
nbanmp: That shouldn't be possible with a correct memory implemention, but the current memory system is not quite correct, so I don't know if it is possible or not.

If the int is bigger than 2^8-1, I don't know how we should handle it. If, for example there are prepended 0 bytes, it is impossible to know the correct size.

tests/laser/state/calldata_test.py
```diff
@@ -0,0 +1,80 @@
...
+    value = calldata[100]
+
+    solver.add(calldata.constraints)
+    solver.check()
+    model = solver.model()
+
+    value = model.eval(value)
+
+    # Assert
+    assert value == 0
```
JoranHonig: why is this 0 in the symbolic calldata situation?

tests/laser/state/calldata_test.py
```diff
@@ -0,0 +1,80 @@
...
+    solver.add(calldata.constraints + [constraint])
+    solver.check()
+    model = solver.model()
+
+    value = model.eval(value)
+    calldatasize = model.eval(calldata.calldatasize)
+
+    # Assert
+    assert value == 50
+    assert simplify(calldatasize >= 100)
```
JoranHonig: calldatasize is an int here right? Why simplify?

tests/laser/state/calldata_test.py
```diff
@@ -0,0 +1,80 @@
+import pytest
```
JoranHonig: One more interesting situation might be the following:

calldatasize == 50
calldata[55] == 0

Then solver.check should return unsat

tests/laser/state/calldata_test.py
nbanmp: Because I forgot to delete a line. :(

tests/laser/state/calldata_test.py
nbanmp: That's what I thought too, but the comparison wasn't returning a bool unless I simplified.

tests/report_test.py
JoranHonig: Okay, and its the debug section that causes this right?
Because if the jsonification fails/takes too long on the report.as_json() for any of the test contracts then we can't just ignore this

tests/laser/state/calldata_test.py
nbanmp: I'll add that.

mythril/laser/ethereum/state.py
JoranHonig: I dont think this statement should be here:
if you have
calldata[0: 5] and calldata = [1]
then it should return
[1, 0, 0, 0, 0] not [1]

tests/report_test.py
nbanmp: Yup.

It also significantly increases runtime of the tests without this calldata improvement, so it's really an issue that should be resolved somehow. Perhaps open a new issue?

mythril/laser/ethereum/state.py
nbanmp: @norhh you trickster. :P

mythril/laser/ethereum/transaction/transaction_models.py
nbanmp:
```
self.call_data = Calldata(self.id, call_data) if not isinstance(call_data, Calldata) else call_data
```
This seems nicer, I'll do this.

nbanmp:
> Can the environment.calldatasize be removed?

@JoranHonig Huh? `Environment.calldatasize` was never a thing?

JoranHonig:
> > Can the environment.calldatasize be removed?
> 
> @JoranHonig Huh? `Environment.calldatasize` was never a thing?

Oh nvm :sweat_smile:

mythril/laser/ethereum/state.py
```diff
@@ -64,20 +64,10 @@ def get_word_at(self, index: int):
...
-            if self.concrete and isinstance(stop, int):
-                calldatasize = get_concrete_int(self.calldatasize)
-                if stop > calldatasize:
-                    stop = calldatasize
-
             try:
                 current_index = item.start if isinstance(item.start, BitVecRef) else BitVecVal(item.start, 256)
                 dataparts = []
-                while simplify(current_index != stop):
+                while simplify(current_index != item.stop):
```
JoranHonig: If item.stop is symbolic, then does this loop run forever?

mythril/laser/ethereum/call.py
JoranHonig: what if you do BitVec(elem, 256) ?

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -102,7 +103,7 @@ def __init__(self,
         self.origin = BitVec("origin{}".format(identifier), 256) if origin is None else origin
         self.call_data_type = BitVec("call_data_type{}".format(identifier), 256) if call_data_type is None else call_data_type
 
-        self.call_data = call_data
+        self.call_data = Calldata(self.id) if call_data is None else Calldata(self.id, call_data) if type(call_data) == list else call_data
```
JoranHonig: Can you do the same fix here that you did at ln 47

tests/laser/state/calldata_test.py
```diff
@@ -6,27 +6,29 @@
...
 @pytest.mark.parametrize("starting_calldata", uninitialized_test_data)
-def test_calldata_uninitialized_index(starting_calldata):
+def test_concrete_calldata_uninitialized_index(starting_calldata):
     # Arrange
     calldata = Calldata(0, starting_calldata)
     solver = Solver()
 
     # Act
     value = calldata[100]
+    value2 = calldata.get_word_at(200)
```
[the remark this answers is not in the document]
nbanmp: Just to check that calldata[200:232] also returns 0 and doesn't error because of an out of range slice.

mythril/laser/ethereum/state.py
nbanmp: I believe the behavior here is correct.

Apparently the `==` and `!=` operators have different behavior with regards to exceptions, so it's kind of weird. I'm using the `!=` operator, so we're all good.
```
x = BitVec('x', 256)
y = x + 32
z = BitVec('z', 256)

if not simplify(x + 32 != y): print("x + 32 != y") # prints x + 32 != y
if not simplify(x + 32 != z): print("x + 32 != z") # raises Z3Exception("Symbolic expressions cannot be cast to concrete Boolean values.") 
```
The Z3Exception is caught and and IndexError is raised.

JoranHonig: What if y = x + Bitvec('p', 256)

JoranHonig: It seems that that also throws a z3 exception

tests/laser/state/calldata_test.py
JoranHonig: It seems that eval returns ExprRefs

mythril/laser/ethereum/call.py
JoranHonig: for documentation #563

tests/report_test.py
JoranHonig: I created one #614

nbanmp:
Closing to merge into develop instead.

PR #558, opened 2018-10-10
Add pip downloads

JoranHonig:


b-mueller:
The badges bar is getting a bit crowded, should we remove the Waffle badge?

JoranHonig:
Yeah, or maybe a smaller version of the waffle badge?

b-mueller:
> Yeah, or maybe a smaller version of the waffle badge?

Yep, it should at least all fit into one line.

PR #560, opened 2018-10-10
Added coloredlogs

nbanmp:
Added the `coloredlogs` dependency to get colored logging.


norhh:
Can you add the dependencies in the setup.py too, that may be the reason for failures of tests in circleci

PR #571, opened 2018-10-14
Add type hints to LASER

dmuhs:
Add type hints to LASER engine where applicable. There are 2 or 3 headers where a type cannot really be inferred or hinting is not possible due to circular imports that'd be required. Resolving that would be a major change, though, so for now I have added some TODOs in the code where that problem occurs.


mythril/laser/ethereum/natives.py
```diff
@@ -53,30 +54,29 @@ def ecrecover(data):
...
+def sha256(data: Union[bytes, str]) -> bytes:
     try:
         data = bytes(data)
     except TypeError:
         raise NativeContractException
     return hashlib.sha256(data).digest()
 
 
-def ripemd160(data):
+def ripemd160(data: Union[bytes, str]) -> List:
```
norhh: for data variable for the functions namely, this one, sha256, ecrecover should have the same type right? As all are being called from the same function.

dmuhs: Interesting finding. Regarding the function signatures, `ripemd160` returns a zero-padded list, while `sha256` returns the raw bytestring digest. You can also check that out in ipython or similar:
```
In [11]: ripemd160("foo".encode())[:16]
Out[11]: [0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 0, 66, 207, 162, 17]

In [12]: sha256("foo".encode())
Out[12]: b',&\xb4kh\xff\xc6\x8f\xf9\x9bE<\x1d0A4\x13B-pd\x83\xbf\xa0\xf9\x8a^\x88bf\xe7\xae'
```
The output of both could be unified as `typing.Iterable`, but I tried to stay as close to the implementation as possible when coming up with the hints.

norhh: It might be good to make them return something same as either of the 4 functions are called each time, let's add that as a Todo in that case.

And also the data arg for identity should also have the same type as that of the other 3 functions mentioned above.

dmuhs: I have added the missing hint for the identity function. It also now throws a `NativeContractException` if the input cannot be converted to bytes - just like the other two functions. I also made sure `ripemd160` also always returns `bytes` after the padding has been added. So now all functions exhibit the same behaviour, take the same inputs and yield the same outputs. :+1:

mythril/laser/ethereum/state.py
```diff
@@ -41,17 +42,18 @@ def __getitem__(self, item):
...
+    def keys(self) -> KeysView:
         return self._storage.keys()
 
+
 class Account:
     """
     Account class representing ethereum accounts
     """
-    def __init__(self, address, code=None, contract_name="unknown", balance=None, concrete_storage=False,
+    def __init__(self, address: str, code=None, contract_name="unknown", balance=None, concrete_storage=False,
```
norhh: This returns a None i think.

mythril/laser/ethereum/state.py
```diff
@@ -162,7 +163,7 @@ def pop(self, index=-1):
         except IndexError:
             raise StackUnderflowException("Trying to pop from an empty stack")
 
-    def __getitem__(self, item):
+    def __getitem__(self, item: int) -> Any:
```
norhh: item can also be of type slice.

mythril/laser/ethereum/util.py
```diff
@@ -10,46 +12,38 @@
 TT255 = 2 ** 255
 
 
-def sha3(seed):
+def sha3(seed: str) -> bytes:
     return _sha3.keccak_256(bytes(seed)).digest()
```
norhh: i think bytes doesn't take straight string as input. as it will giveout some error with that.It seems to take some Iterable[Int] as an input.

norhh:
Nevermind, guessing types now might be painful without integrating with mypy, I am adding the mypy tests on circleci to a branch of mythril repo, when thats done, lets add the types to that branch and then when everything perfect with the results then lets setup a pr to the master.

norhh:
hey, can you setup a PR for mypy_branch or just push it to mypy_branch and then let's check how this fares :slightly_smiling_face:

dmuhs:
Good idea. Is the mypy branch based on the current master? There are still some open merge conflicts from the last module refactoring. I hope to fix these by this evening (in about 4h)

norhh:
it's a bit behind the current master, but i think there will still be the conflicts.

dmuhs:
@norhh Pushed the changes (and the most recent `master` commits) to the mypy branch :+1:

dmuhs:
Closing and opening another PR to integrate changes to current `develop` branch

Issue #539, opened 2018-09-29
Random search strategy


JoranHonig:
## Description
Laser currently implements two basic search strategies, namely bfs and dfs.
However from literature we know that other seach strategies can be more effective. 

### Requirements

Implentation of three search strategies.
1. Each iteration pick a random element from the worklist
2. Each iteration pick a random element from the worklist but elements that are deeper in the statespace should be chosen with less likeliness. 
3. Each iteration consider a statespace described by the traces that lead to the open states. Start from the beginning node and at each node randomly pick one branch to follow until you reach an open state

norhh:
@JoranHonig can you elaborate on what to do when an open state is reached?

PR #579, opened 2018-10-15
Add the two probabilistic strategies

norhh:
currently implemented the first 2 strategies mentioned in #539 .


mythril/laser/ethereum/strategy/basic.py
```diff
@@ -1,54 +1,80 @@
...
+    from bisect import bisect
+
+    def choices(population, weights=None):
+        if weights is None:
+            return [population[int(random() * len(population))]]
+        cum_weights = accumulate(weights)
+        return [population[bisect(cum_weights, random()*cum_weights[-1], 0, len(population)-1)]]
+
+
+class BasicStrategy(ABC):
```
JoranHonig: I'd name this SearchStrategy.
Also can you put this in __init__ ? I think that is a better place for this class

mythril/laser/ethereum/strategy/basic.py
```diff
@@ -1,54 +1,80 @@
...
 
+try:
+    from random import choices
+except ImportError:
+
+    from itertools import accumulate
+    from random import random
+    from bisect import bisect
+
+    def choices(population, weights=None):
```
JoranHonig: pydoc?

mythril/laser/ethereum/strategy/basic.py
norhh: Let’s add the ‘basic’ term as a prefix as not all search strategies can be emulated with this class.how about calling it something like ‘BasicSearchStrategy’.

mythril/laser/ethereum/strategy/basic.py
norhh: Sorry, I didn't get what exactly to be done, if it is comments about the function then I have just added it now.

norhh:
I think it might be good to construct a graph now for the 3rd strategy as it might be possible to try out a lot more strategies easily if we have a graph for the worklist.I will put that PR after this gets reviewed as it might be good to have smaller PRs for review.

mythril/analysis/symbolic.py
```diff
@@ -21,6 +22,10 @@ def __init__(self, contract, address, strategy, dynloader=None, max_depth=22,
             s_strategy = DepthFirstSearchStrategy
```
JoranHonig: What do you think about the following:
Modules in laser.strategy all define some metadata and are dynamically loaded vs doing it hardcoded here

JoranHonig: This should be a separate pr ofc

norhh: :+1:

Issue #533, opened 2018-09-26
SMT abstraction


JoranHonig:
## Description
Currently we are directly using z3 objects in laser, and are often directly using the z3 api. This has a few downsides. 

- Firstly this makes most of the code dependent on z3, which makes it difficult to plug in another SMT solver which might work better on the specific queries we have in mythril.
- Secondly, at some point we want to start tracking taint in "run time" this would be easy to do if we can add a record to a BitVec object, which is impossible using z3 objects directly.
How this could look:
```python3
> x = BitVector("x")
> x.has_taint()
false
> x.taint()
> x.has_taint()
true
```

#### Abstraction layer
To solve this problem we should build abstraction around SMT solving logic, which would provide an interface along the lines of https://github.com/angr/claripy.

#### Taint analysis
For the dynamic taint analysis we'd need to be able to taint a bitvector multiple times, and to distinguish the different taints. This is required because different analysis modules will run at the same time, and the integer overflow detection should not look for tainted variables from another module.

norhh:
@hzzhang , Thats quite interesting, will look into that

JoranHonig:
@hzzhang That still leaves taint analysis unimplemented

norhh:
we can implement an interface between pysmt and mythril where we can implement stuff like taint .

hzzhang:
@JoranHonig  @norhh Agree, another interface would be needed.

hzzhang:
@JoranHonig If no one has already been working on this, I'd like to take it.

JoranHonig:
@hzzhang Go ahead and pick this up :+1:

hzzhang:
![](https://github.trello.services/images/mini-trello-icon.png) [Add SMT abstract layer for Mythril OSS](https://trello.com/c/do9uht3s/55-add-smt-abstract-layer-for-mythril-oss)

JoranHonig:
I'm going to take this over because @hzzang doesn't have time to work on this right now.

JoranHonig:
I've looked into the following three options:

1. Roll our own abstraction layer
Possible, and it would allow us to build a solution that exactly fits our needs. But would take more time than the other two options

2. Extend pySMT
This is the approach that hzzang took. In the pr we see that a symbol manager can be used to store properties for each symbol. However, this doesn't allow us to taint or annotate expressions, therefore it doesn't fully fit our needs yet.

3. Use claripy
- Has nice annotation system that already works. 
[...]

PR #581, opened 2018-10-16
Add SMT abstraction

hzzhang:
This patch series is to address issue #533. It replaces the use of Z3 in Mythril by [pySMT](https://pysmt.readthedocs.io/en/latest/) that provides a unified APIs for SMT-lib compatible constraint solvers (including Z3, CVC4, Yices, etc.).

In addition, a modified [pySMT](https://github.com/hzzhang/pysmt/tree/hzzhang/props) is used to support attaching arbitrary properties to pySMT FNode (which represents expressions, formulas and constraints akin to Z3 ExprRef), so that we can, for example, add the taint information to symbols in the future. You can refer to `tests/laser/test_pysmt.py` for a simple example.

All tests from `all_tests.sh` are passed locally when the Z3 solver is used. Other solvers have not been tested, because they may require additional adjustment, especially for configurations of constraint solving, e.g., the timeout may not be supported by every solver and may be configured in a solver-specific way which has not been unified by pySMT).

This patch series is organized as below.
* Patch 01-02 fix/cleanup the existing code and are not quite relevant to this patch series.
* Patch 03-11 replaces most uses of Z3 APIs by a set of wrapper functions, so that when switching to pySMT APIs, we only need to reimplement wrappers to use pySMT APIs.
* Patch 12-17 reimplement above wrappers on pySMT APIs and fix several issues after the API switch.

### DIFF
--- a/mythril/analysis/callgraph.py
+++ b/mythril/analysis/callgraph.py
@@ -2,7 +2,7 @@ import re
 
 from jinja2 import Environment, PackageLoader, select_autoescape
 from mythril.laser.ethereum.svm import NodeFlags
-import z3
+from mythril.laser.ethereum.smt_wrapper import simplify, SimplificationError
 
 default_opts = {
     'autoResize': True,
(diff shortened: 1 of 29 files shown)


hzzhang:
The order of commits in this PR seems inconsistent to what are in my repo. Close and will resend later.

PR #587, opened 2018-10-17
Remove %s in formatting

norhh:
previously %s was used to display variable for string formatting which won't work.


norhh:
Looks like the integration tests have failed due to some reason with the api

PR #592, opened 2018-10-17
Add SMT abstraction

hzzhang:
This patch series is to address #533, and has passed all tests from `all_tests.sh` in my local environment.

It replaces all usages of Z3 APIs to [pySMT](https://github.com/pysmt/pysmt) [APIs](https://pysmt.readthedocs.io/en/latest/). The latter provides a unified interface to interact with SMTLib compatible constraint solvers (including Z3, CVC4, etc.). Note, this patch series only uses Z3 via pySMT and is only tested with Z3. The support for other solvers may need additional work, such as adjusting the solver-specific parameters of constraint solving.

In addition, this patch series also introduces a symbol manager to allow attaching arbitrary properties to a symbol (e.g., a bit vector symbol), which can address the second requirement in #533. Note, this patch series has not switched the taint analysis to the symbol manager yet, which is left for the future work. A simple example can be found in `tests/laser/test_symbol_manager.py`.

In order to ease the API switch and to avoid merge conflict with other pending PRs, this patch series introduces a set of wrapper functions which have the same function prototypes as those Z3 APIs ever used in Mythril. Other codes could keep some Z3-like usages, except they have to replace `from z3 import xxx` to `from mythril.laser.ethereum.smt_wrapper import xxx`.  For example, you can continue to use `bv = BitVecVal(0x100, 256)` to create a 256-bit bit vector of value 0x100.

However, there are still some differences between pySMT APIs and Z3 APIs, which disable some Z3-like usages.
1. **Equations** and **inequations** must be explicitly expressed by `Eq` and `Neq` from `mythril.laser.ethereum.smt_wrapper`, or `Equals` and `NotEquals` from pySMT, rather than overrides of `==` and `!=`, because pySMT does not override those operators.

2. **Signed bit vector division** must be explicitly expressed by `SDiv` from `mythril.laser.ethereum.smt_wrapper`, or `BVSDiv` from pySMT, rather than the override of `/`, because pySMT overrides `/` as unsigned bit vector division.

3. **Signed bit vector comparison** must be explicitly expressed by `SLT` and `SGT` from `mythril.laser.ethereum.smt_wrapper`, rather than the override of `<` and `>`, because pySMT overrides those operators as signed comparison.

4. **`expr.as_long()`** is not available any more, and should be replaced by `get_concrete_value(expr)` from `mythril.laser.ethereum.smt_wrapper`, or `expr.constant_value()` from pySMT.

5. **All Z3 defined types**, such as `BoolRef`, `BitVecRef` and `BitVecNumRef`, are not available any more, and should be replaced by `is_bool`, `is_bv` and `is_bv_value` from `mythril.laser.ethereum.smt_wrapper`, or corresponding type inspect functions in pySMT.

The patch series is organized as below.
1. Patch 1&2 make minor fixes and cleanup which are not quite relevant to this patch series.
2. Patch 3-11 replace most Z3 API usages by a set of wrapper functions.
3. Patch 12-16 reimplement above wrapper functions on pySMT APIs, replace the remaining Z3 APIs and fixes some regressions after the API switch.
4. Patch 17 introduces the symbol manager.


b-mueller:
@hzzhang @JoranHonig it looks like develop has diverged significantly from this PR. I suggest we close it for now and once @hzzhang becomes available again, start applying his changes in smaller steps?

PR #594, opened 2018-10-17
Add query-signature argument, add ethereum-input-decoder to requirements

mmhh1910:
This PR adds the ethereum-input-decoder python module to the requirements installed with pip and adds an argument -q/--query-signature to enable function signature lookups through www.4byte.directory.

"Fixes" https://github.com/ConsenSys/mythril/issues/590.

### DIFF
--- a/mythril/disassembler/disassembly.py
+++ b/mythril/disassembler/disassembly.py
@@ -5,7 +5,7 @@ import logging
 
 class Disassembly(object):
 
-    def __init__(self, code, enable_online_lookup=True):
+    def __init__(self, code, enable_online_lookup=False):
         self.instruction_list = asm.disassemble(util.safe_decode(code))
         self.func_hashes = []
         self.func_to_addr = {}
--- a/mythril/ether/ethcontract.py
+++ b/mythril/ether/ethcontract.py
@@ -6,7 +6,7 @@ import re
 
 class ETHContract(persistent.Persistent):
 
-    def __init__(self, code, creation_code="", name="Unknown", enable_online_lookup=True):
+    def __init__(self, code, creation_code="", name="Unknown", enable_online_lookup=False):
         
         # Workaround: We currently do not support compile-time linking.
         # Dynamic contract addresses of the format __[contract-name]_____________ are replaced with a generic address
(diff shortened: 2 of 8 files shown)


norhh:
We didn't want to increase extra dependencies so can you remove it from requirements and add a warning saying install ethereum_input_decoder if it doesn't exist when flag is turned on?

mmhh1910:
> We didn't want to increase extra dependencies so can you remove it from requirements and add a warning saying install ethereum_input_decoder if it doesn't exist when flag is turned on?

If the flag is turned on and the ethereum-input-decoder module could not be loaded, there will be an error message already saying "The --query-signature function requires the python package ethereum-input-decoder" and program termination.

Regarding the extra dependency on ethereum-input-decoder: I think the online lookup is a very useful feature for the user, so I wonder if the additional dependency is that bad. It's a really simple module with no further new dependencies. 
Not having it in the requirements file will make the feature practically unusable for all users of the docker image.

norhh:
I think the dependency should also be added in setup.py

mmhh1910:
> I think the dependency should also be added in setup.py

Right! Added and changed the strict version requirement ==0.2.2 to >=0.2.2 to be in line with the overall requirements setup.

norhh:
Add  'enable_online_lookup=False'  for ETHContract class in report_test.py.That might fix it.

mmhh1910:
Did that and now it passed for session/env Python 3.5, but fails for 3.6.

Not sure how this CI system works... Is it possible that the function name was saved to `~/mythril/signatures.json` in a previous run and is now picked up from there on subsequent runs?

norhh:
I am not sure as these are run on bytecode files right?, can remove the ETHContract's default of enable_online_lookup to False and then rerun it?

mmhh1910:
I assume this was due to `self.sigs = signatures.SignatureDb()` in `__init__` of `class Mythril(object)`.

Not sure where to go from here. Disassembly, ETHContract and SignatureDb all have a enable_online_lookup parameter with a default value, which is never overwritten in constructor calls.

Should we add an enable_online_lookup parameter to the "top" Mythril class and make sure that all constructors for Disassembly, ETHContract and SignatureDb are called with the value passed into the Mythril constructor?

norhh:
Yeah thats better.

mmhh1910:
- Default is now off (no online lookups) no matter how it/what is used (class Disassembly, ETHContract, SignatureDb or Mythril or through the cli interface)
- Specifying --query-signature as a parameter for the cli interface without module ethereum-input-decoder installed, results in an error message
- Setting enable_online_lookup to True in the constructor of on the four classes without module ethereum-input-decoder installed will simply result in no lookups.

PR #603, opened 2018-10-22
Refactor asm

JoranHonig:
The module is moved to disassembler, it seems like this is a more logical place for it to be.

Additionally some refactoring and code simplification

Todo:
- [x] Unit Tests


dmuhs:
Looks good! Could you also add type hints for the function/method returns?

dmuhs:
Check the last review comment in `mythril/disassembler/asm.py` ;)

PR #606, opened 2018-10-23
Add solc-args to signatures

### LINKED ISSUE #605: Fix solc-args when using --allow-paths
## Description
using --solc-args isn't working for any contracts when imports are happening.

## Bug
Run `myth --solc-args '<some arg related to allowed paths>' <filename with an import>`
results in an error.

norhh:
fixes #605


norhh:
nvm this seems to be already fixed in master recently :laughing: .

Issue #523, opened 2018-09-21
Output transaction sequence for reaching vulnerable state


b-mueller:
## Description

Add an option that outputs the transactions for reaching each vulnerable state.

## Background

We already have all the data we need from the solver. It should be trivial to reformat this output for various purposes:

- As a Truffle test;
- As an executable exploit;
- As a script that deploys the contract to Ganache and attempts to exploit the issue.

## Scope

Compose the call data for each message call in the transaction sequence. For example, running an assert test on the following example should produce three transactions:

```
contract Token{

    mapping(address => uint) public balances;
    function airdrop() public{
        balances[msg.sender] = 1000;
    }

    function consume() public{ 
        require(balances[msg.sender]>0);
        balances[msg.sender] -= 1;
    }

    function backdoor() public{
        balances[msg.sender] += 1;
    }
 
   function test_invariants() {
      assert(balances[msg.sender] <= 1000);
  }
}
```

nbanmp:
We don't necessarily have all the data we need from the solver. To generate a test case, we'd need to add a few more constraints. Most importantly: `caller` and `value`. One cannot send ether from any wallet, nor can they send an unlimited amount of ether.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 2.2 ETH (506.31 USD @ $230.14/ETH)  attached to it__ as part of the Mythril fund__.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ConsenSys/mythril/523/1399).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ConsenSys/mythril/523/1399).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $49,675.65 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


[...]

PR #619, opened 2018-10-25
Concrete transaction sequence generation

nbanmp:
PR for #523.


b-mueller:
Generates the following output for me:

```
$ $MYTH -msuicide -v2 -x simple_suicide.sol

Transaction Sequence: {
  '2': {'calldata': '0x1651065990', 'caller': '0xCA11EDEADBEEFCA11EDEADBEEFCA11ED37E636E6', 'call_value': '0x0'}
}

$ MYTH -msuicide -v2 -x suicide_multitx_feasible.sol

Transaction Sequence: 
{
  '2': {'call_value': '0x0', 'caller': '0xCA11EDEADBEEFCA11EDEADBEEFCA11ED37E636E6', 'calldata': '0x16468245233'}, 
  '4': {'call_value': '0x0', 'caller': '0xCA11EDEADBEEFCA11EDEADBEEFCA11ED37E636E6', 'calldata': '0x16468245233'}, 
  '8': {'call_value': '0x0', 'caller': '0xCA11EDEADBEEFCA11EDEADBEEFCA11ED37E636E6', 'calldata': '0x16468245233'}
}
```

Seems that calldata is repeated in the second output, is this caused by the bug fixed in #620?

Examples used are from [SWC}(https://github.com/SmartContractSecurity/SWC-registry/tree/master/test_cases/unprotected_critical_functions).

b-mueller:
Still a slight issue there:

```
$ $MYTH -x -msuicide -v2 simple_suicide.sol
root[91253] INFO Finished symbolic execution
root[91253] INFO 11 nodes, 10 edges, 65 total states
root[91253] INFO Achieved 93.47826086956522 coverage for code: 608060405260043610603f576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff168063a56a3b5a146044575b600080fd5b348015604f57600080fd5b5060566058565b005b3373ffffffffffffffffffffffffffffffffffffffff16ff00a165627a7a72305820ef91c543c51cdd0b7e9142c825b0ee060caa08df8acc375c1e444d8f1ba2f84b0029
root[91253] INFO Starting analysis
root[91253] INFO Executing <module 'suicide' from '/Users/bernhardmueller/Projects/mythril-classic/mythril/analysis/modules/suicide.py'>
root[91253] DEBUG Executing module: UNCHECKED_SUICIDE
root[91253] DEBUG [UNCHECKED_SUICIDE] suicide in function sudicideAnyone()
root[91253] DEBUG Transaction Sequence: {'2': {'call_value': '0x0', 'calldata': '0xa56a3b5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a5a', 'caller': '0xCA11EDEADBEEFCA11EDEADBEEFCA11ED37E636E6'}}
```

The last byte of the func hash seems to be repeated?

nbanmp:
@b-mueller That's just a side effect of the way z3 handles Arrays. When concretized, they specify certain values, then have a default value for everything that is unspecified. In this case, z3 has decided, quite arbitrarily, that the default value should be `5a`, and that the calldata length should be `260` If you were to a transaction with that generated test case, it would work.

b-mueller:
>  If you were to a transaction with that generated test case, it would work.

True, but it does look weird.

Is there any easy way to work around this besides actively managing calldata length? E.g. setting the default value to zero?

b-mueller:
Another option would be to keep track of the largest calldata index accessed during execution and adding that as a constraint (< max_index_accessed + 32) when concretizing. Also a hack though and not precise enough to realize if only 4 bytes are needed, like in a func call with no arguments.

I'd say let's merge this as-is now, and add a solution later on (even though it would be nice to have it before DevCon :))

nbanmp:
Looks like it was fairly straightforward to make it work with z3's `Optimize`.

Just kidding... I again pushed too soon... It's no surprise that z3 has to think way to hard to make that optimization work... reverting. :(

b-mueller:
Ok, I'll merge it without the optimization.

PR #621, opened 2018-10-26
Update issue descriptions

b-mueller:
Update descriptions slightly so they make more sense when displayed by the current versions of truffle+analyze and sabre.


b-mueller:
Seems like this is not passing tests because of `signatures.json` mismatch. Is there a default `signatures.json` I can use to generate the expected outputs?

JoranHonig:
You can replace your own signatures.json with:
tests/mythril_dir/signatures.json.example, that will give the correct results

PR #632, opened 2018-10-28
Feature/type hints develop

dmuhs:
Add type hints to laser engine, related to #571


mythril/laser/ethereum/state.py
```diff
@@ -283,7 +284,7 @@ def append(self, element):
             )
         super(MachineStack, self).append(element)
 
-    def pop(self, index=-1):
+    def pop(self, index=-1) -> BitVecNumRef:
```
norhh: I think the stack contains BitVec expressions. Also I think BitVecNum is inherited from BitVec.

mythril/laser/ethereum/natives.py
```diff
@@ -15,24 +16,26 @@ class NativeContractException(Exception):
...
     if i >= len(data):
         return 0
     o = data[i : min(i + 32, len(data))]
     o.extend(bytearray(32 - len(o)))
     return bytearray_to_int(o)
 
 
-def ecrecover(data):
-
+def ecrecover(data: str) -> List:
```
norhh: This should have the same input as the ones below, like sha256, ripemd etc

mythril/laser/ethereum/svm.py
```diff
@@ -36,7 +38,7 @@ class LaserEVM:
 
     def __init__(
         self,
-        accounts,
+        accounts: List[Account],
```
norhh: I think it is a dict rather then a list

mythril/laser/ethereum/svm.py
```diff
@@ -75,10 +77,16 @@ def __init__(
         )
 
     @property
-    def accounts(self):
+    def accounts(self) -> List[Account]:
```
norhh: Same here

mythril/laser/ethereum/state.py
```diff
@@ -295,7 +296,7 @@ def pop(self, index=-1):
         except IndexError:
             raise StackUnderflowException("Trying to pop from an empty stack")
 
-    def __getitem__(self, item):
+    def __getitem__(self, item: int) -> Any:
```
norhh: it can also be a slice right?

mythril/laser/ethereum/state.py
```diff
@@ -217,12 +218,12 @@ class Environment:
...
         self,
-        active_account,
-        sender,
-        calldata,
-        gasprice,
-        callvalue,
-        origin,
+        active_account: Account,
+        sender: BitVecRef,
+        calldata: List,
```
norhh: I think this will be of type CallData which is present in state.py

mythril/laser/ethereum/natives.py
```diff
@@ -53,30 +56,35 @@ def ecrecover(data):
...
     except TypeError:
         raise NativeContractException
-    return 12 * [0] + [i for i in hashlib.new("ripemd160", data).digest()]
+    digest = hashlib.new("ripemd160", data).digest()
+    padded = 12 * [0] + list(digest)
+    return bytes(padded)
 
 
-def identity(data):
+def identity(data: Union[bytes, str]) -> bytes:
```
norhh: I think the input type here will be same as that of starting_calldata which is a list

mythril/laser/ethereum/natives.py
```diff
@@ -53,30 +56,35 @@ def ecrecover(data):
...
+    try:
+        data = bytes(data)
+    except TypeError:
+        raise NativeContractException
     return copy.copy(data)
 
 
-def native_contracts(address, data):
+def native_contracts(address: int, data: List):
```
norhh: data is of type CallData, due to the PR in develop

mythril/laser/ethereum/call.py
```diff
@@ -138,12 +141,18 @@ def get_callee_account(global_state, callee_address, dynamic_loader):
     return callee_account
 
 
-def get_call_data(global_state, memory_start, memory_size, pad=True):
+def get_call_data(
+    global_state: GlobalState,
+    memory_start: Union[int, BitVecNumRef, BoolRef],
```
norhh: the memory_start and memory_size can both be symbolic right?

dmuhs: Correct. I just noticed that with the new `Calldata` object from the develop branch the `pad` parameter is obsolete. Should I throw that out?

mythril/laser/ethereum/natives.py
dmuhs: I think this would be in conflict with the other signatures for the native crypto functions, because they're all called in a row [here](https://github.com/ConsenSys/mythril-classic/blob/feature/type-hints-develop/mythril/laser/ethereum/natives.py#L90) with the same input for each

mythril/laser/ethereum/state.py
dmuhs: Just found the inheritance diagram for Z3 types: https://z3prover.github.io/api/html/classz3py_1_1_expr_ref__inherit__graph.png

So I'll replace the specific hints with `ExprRef` for symbolic types. That should also cover BoolRefs

mythril/laser/ethereum/state.py
dmuhs: Yep. Added the `slice` type to this and the other `__getitem__` signatures

mythril/laser/ethereum/call.py
JoranHonig: Remove it :+1:

mythril/laser/ethereum/natives.py
```diff
@@ -53,30 +57,35 @@ def ecrecover(data):
...
-    return 12 * [0] + [i for i in hashlib.new("ripemd160", data).digest()]
+    digest = hashlib.new("ripemd160", data).digest()
+    padded = 12 * [0] + list(digest)
+    return bytes(padded)
 
 
-def identity(data):
+def identity(data: Union[bytes, str]) -> bytes:
+    try:
+        data = bytes(data)
```
JoranHonig: Doesn't this already create some kind of copy of data?

JoranHonig: In any case, it seems like bytes is an immutable type, so copying wouldn't make sense anyway right?

JoranHonig: But right now the copy is unnecessary right?

norhh: yep

mythril/laser/ethereum/natives.py
```diff
@@ -3,36 +3,40 @@
...
     if i >= len(data):
         return 0
     o = data[i : min(i + 32, len(data))]
     o.extend(bytearray(32 - len(o)))
     return bytearray_to_int(o)
 
 
-def ecrecover(data):
-
+def ecrecover(data: Union[bytes, str]) -> bytes:
```
norhh: can you add a list too into the `data`'s type?

norhh: bunch of ascii numbers from what i saw.

dmuhs: Done :+1:

JoranHonig: I'm not seeing a change here, is that correct?

mythril/laser/ethereum/natives.py
```diff
@@ -75,11 +76,16 @@ def ripemd160(data: Union[bytes, str]) -> bytes:
...
-        data = bytes(data)
-    except TypeError:
-        raise NativeContractException
-    return copy.copy(data)
+    # Group up into an array of 32 byte words instead
+    # of an array of bytes. If saved to memory, 32 byte
+    # words are currently needed, but a correct memory
+    # implementation would be byte indexed for the most
+    # part.
+    return data
```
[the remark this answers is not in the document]
JoranHonig: https://github.com/ConsenSys/mythril-classic/pull/664 fixes the memory addressing thing. In which case the return type should be a bytes again right?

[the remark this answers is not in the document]
dmuhs: @JoranHonig So we can simply leave bytes here and stick to `return data` for now until #664 is merged?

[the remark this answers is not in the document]
JoranHonig: This pr is about type hints so I'm fine with merging this

Issue #630, opened 2018-10-27
NameError: name 'state' is not defined


b-mueller:
## Description

Mythril crashes with a `NameError` on the example file `etherstore.sol`.

## How to Reproduce

```
$ $MYTH -x solidity_examples/etherstore.sol 
Traceback (most recent call last):
  File "/Users/bernhardmueller/Projects/mythril-classic/myth", line 9, in <module>
    mythril.interfaces.cli.main()
  File "/Users/bernhardmueller/Projects/mythril-classic/mythril/interfaces/cli.py", line 234, in main
    max_transaction_count=args.max_transaction_count)
  File "/Users/bernhardmueller/Projects/mythril-classic/mythril/mythril.py", line 375, in fire_lasers
    issues = fire_lasers(sym, modules)
  File "/Users/bernhardmueller/Projects/mythril-classic/mythril/analysis/security.py", line 20, in fire_lasers
    issues += module.execute(statespace)
  File "/Users/bernhardmueller/Projects/mythril-classic/mythril/analysis/modules/dependence_on_predictable_vars.py", line 69, in execute
    bytecode=state.environment.code.bytecode,
NameError: name 'state' is not defined
```

## Environment

- Mythril version: output from  Master
- Solidity compiler and version:  0.4.24+commit.e67f0147.Darwin.appleclang
- Python version: 3.6.3
- OS and Version: Mac OS Mojave

PR #635, opened 2018-10-28
Merge master

JoranHonig:


mythril/analysis/modules/integer.py
```diff
@@ -198,15 +199,16 @@ def _check_integer_underflow(statespace, state, node):
...
                     address=instruction["address"],
                     swc_id=INTEGER_OVERFLOW_AND_UNDERFLOW,
+                    bytecode=state.environment.code.bytecode,
                     title="Integer Underflow",
                     _type="Warning",
                 )
 
                 issue.description = (
-                    "The subtraction can result in an integer underflow.\n"
+                    "The substraction can result in an integer underflow.\n"
```
norhh: this should be renamed to subtraction :+1: , mostly my mistake .

mythril/analysis/modules/dependence_on_predictable_vars.py
```diff
@@ -63,9 +63,10 @@ def execute(statespace):
                 )
                 issue = Issue(
                     contract=call.node.contract_name,
-                    function=call.node.function_name,
+                    function_name=call.node.function_name,
                     address=address,
                     swc_id=swc_type,
+                    bytecode=state.environment.code.bytecode,
```
norhh: This should be call.state right? might be related to #630

norhh: In this case it should be call.state right? may be related to #630

PR #639, opened 2018-10-29
Adds the third strategy and a graph module

### LINKED ISSUE #539: Random search strategy

norhh:
fixes #539


mythril/laser/ethereum/graph.py
```diff
@@ -0,0 +1,31 @@
+from typing import List, Dict
```
JoranHonig: Is there a reason why this should be in the laser.ethereum module vs laser.ethereum.strategy?

norhh: One reason was this isn't a strategy, but this collides with the control flow graph so it might be better to move it to strategy

mythril/laser/ethereum/strategy/__init__.py
```diff
@@ -1,12 +1,39 @@
 from abc import ABC, abstractmethod
 
+from mythril.laser.ethereum.graph import SimpleGraph, Graph
+
 
 class BasicSearchStrategy(ABC):
-    __slots__ = "work_list", "max_depth"
+    __slots__ = "graph", "max_depth"
+
+    def __init__(self, graph: SimpleGraph, max_depth: int):
```
JoranHonig:
```suggestion
    def __init__(self, graph: Graph, max_depth: int):
```

mythril/laser/ethereum/strategy/__init__.py
```diff
@@ -1,12 +1,39 @@
 from abc import ABC, abstractmethod
 
+from mythril.laser.ethereum.graph import SimpleGraph, Graph
+
 
 class BasicSearchStrategy(ABC):
```
JoranHonig: Its getting to be a BaseSearchStrategy, vs BasicSearchStrategy. I think we should rename the class

JoranHonig: WDYT?

mythril/laser/ethereum/graph.py
```diff
@@ -0,0 +1,31 @@
...
+
+    def add_edges(self, from_vertex: GlobalState, to_vertices: List):
+
+        self.adjacency_list[from_vertex] = []
+        for to_vertex in to_vertices:
+            self.work_list.append(to_vertex)
+            self.adjacency_list[from_vertex].append(to_vertex)
+
+
+class SimpleGraph:
```
JoranHonig: This is not really a graph

mythril/laser/ethereum/strategy/advanced.py
```diff
@@ -0,0 +1,34 @@
...
+    def __init__(self, graph: Graph, max_depth: int):
+        super(RandomBranchDepthFirstSearch, self).__init__(graph, max_depth)
+
+    def get_strategic_global_state(self) -> GlobalState:
+
+        if not self.graph.work_list:
+            raise IndexError
+
+        try:
+            edge_list = copy(self.graph.adjacency_list[self.graph.work_list[-1]])
```
JoranHonig: I'd make a function in the graph object for this

mythril/laser/ethereum/graph.py
```diff
@@ -0,0 +1,31 @@
...
+class Graph:
+    def __init__(self):
+        self.adjacency_list = dict()
+        self.work_list = []
+
+    def add_vertex(self, global_state: GlobalState):
+        self.work_list.append(global_state)
+        self.adjacency_list[global_state] = []
+
+    def add_edges(self, from_vertex: GlobalState, to_vertices: List):
```
JoranHonig: This is never used right?

JoranHonig:
Okay, while reviewing I have this one comment that is hard to pinpoint on any one piece of code. But to me it seems like everything is getting very dependent on the graph module while most of the code doesn't (need to) know or use any graph related functionality

For example:
 The svm module shouldn't know or care whether its strategy has a graph or what kind of graph, or whatever a graph is. As in doing so we make the svm code dependent on this graph interface,

Wouldn't it be possible to get the same information using the statespace maintained in the svm?

norhh:
` The svm module shouldn't know or care whether its strategy has a graph or what kind of graph, or whatever a graph is. As in doing so we make the svm code dependent on this graph interface,`
It doesn't know it right? we are just using self.graph.add_vertex() everywhere assuming every graph class has this, I think it might be good to add an abstract class which defines BaseGraph which everything needs to inherit.What do you think?

mythril/laser/ethereum/graph.py
norhh: :laughing: can you give me any suggestions for a nice name here

mythril/laser/ethereum/strategy/__init__.py
norhh: But this isn't the Baseclass for all the search strategies right?but we can make it base class for AdvancedSearchStrategy WDYT?

mythril/laser/ethereum/graph.py
norhh: moved it to strategy

tests/mythril_dir/config.ini
```diff
@@ -0,0 +1,4 @@
+[defaults]
```
JoranHonig: What is this doing here?

norhh: oops, some git add problem.

JoranHonig:
@norhh You are right that laser isn't really aware of the specific implementation of Graph. However it is still aware of what a BaseGraph is, and my question is "what justifies this added dependency"

Because I think a graph isn't really a generic thing used by all search strategies. 

For example, if I look at this line (the breath first search):
```
    def get_strategic_global_state(self) -> GlobalState:
        return self.graph.work_list.pop(0)
```
Then it is not immediately clear why or what the graph is doing here, it isn't even a graph.
This change didn't add any functionality to that module, but did make it more complex.

norhh:
@JoranHonig , Isn't it  hard to maintain multiple structures for multiple strategies without making to adhere them to some standards(in this case the BaseGraph)? I may have missed a better way of doing it, so can you provide me some suggestion in that aspect.

JoranHonig:
`Isn't it hard to maintain multiple structures for multiple strategies without making to adhere them to some standards` 
Exactly, and what I'm saying is that BaseGraph doesn't seem to me to be a generic thing that all strategies do or use

b-mueller:
What's up with this PR?

norhh:
:sweat: I am out of ideas of changing structure for BaseGraph without any complications in svm and also in other strategies without making it worse than the current situation.(Also i left it for some time to work on bugs and that painful solver per branch one) @dmuhs , @nbanmp , @JoranHonig any ideas here to make it better?

norhh:
closing it now, we can reuse this later

PR #642, opened 2018-10-29
Output transaction sequence in verbose output

nbanmp:


nbanmp:
Although it might not be directly related, this should be noted:

With the caller constraint added by transaction sequence generation, `token-with-backdoor-test.sol` does not work. Sort of.
```
import "./token-with-backdoor.sol";

contract TokenTest is Token {

    // Verify that the user's balance can never exceed 1000
    function verify_balance() {
        assert(balances[msg.sender] <= 1000);
        assert(false); // <----------------------
    }
}
```
It detects both asserts when that assert(false) line is there, but does not detect `assert(balances[msg.sender] <= 1000)` without the `assert(false)` line.

I believe that `./myth -mexceptions --max-transaction-count 2 -x ./token-with-backdoor-test.sol` (original version, not with `assert(false)`) does not work without the constraints for the concrete transaction added (so without this pr) when it probably should be working.

nbanmp:
Oh. It looks like the issue is that since it is writing to a storage variable indexed by `msg.sender`, sstore solves for a value for the sender and writes to that location in memory, which won't line up with my arbitrary constraint added later.

This makes me think that I should not be constraining a transaction's callvalue and caller immediately before concretizing it. If these are constrained, then it should be done during transaction initialization and should not be only for concretization. As such, I will be removing these constraints from the concretization logic. We can add them to transaction initialization later, perhaps passed in as command line arguments.

nbanmp:
See https://github.com/ConsenSys/mythril-classic/issues/643 to make the transaction concretion nicer.

PR #644, opened 2018-10-30
Minor cleanup and fixes

hzzhang:
For the cleanup
- remove the unused solver definition in sstore_()
- remove the unnecessary deepcopy of constraints in integer module

For the fixes,
- add the missing import of `copy` in transaction_order_dependence.py
- improve a condition checking in sstore_() and thus eliminate some unnecessary subsequent states

### DIFF
--- a/mythril/analysis/modules/transaction_order_dependence.py
+++ b/mythril/analysis/modules/transaction_order_dependence.py
@@ -1,5 +1,6 @@
 import logging
 import re
+import copy
 
 from mythril.analysis import solver
 from mythril.analysis.ops import *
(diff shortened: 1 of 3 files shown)


mythril/analysis/modules/transaction_order_dependence.py
```diff
@@ -1,5 +1,6 @@
 import logging
 import re
+import copy
```
JoranHonig: Is this necessary?

hzzhang: Yes, copy.deepcopy() is used in _can_change() in this file.

PR #646, opened 2018-10-30
Add --no-onchain-storage-access flag

### LINKED ISSUE #586: Remove the -l flag dependency for the onchain contract lookup
## Description
currently -l flag is being used for getting onchain data of mythril, which might lead to confusion.so remove that dependency.

norhh:
added a --no-onchain-storage-access 
when turned on: onchain storage lookup is turned off
else: onchain storage lookup happens

also the -l flag is now used for contract loading only.
fixes #586


mythril/laser/ethereum/call.py
```diff
@@ -122,9 +122,9 @@ def get_callee_account(global_state, callee_address, dynamic_loader):
 
     try:
         code = dynamic_loader.dynld(environment.active_account.address, callee_address)
-    except Exception:
-        logging.debug("Unable to execute dynamic loader.")
-        raise ValueError()
+    except ValueError as error:
+        logging.debug("Unable to execute dynamic loader because: {}".format(str(error)))
+        raise ValueError(str(error))
```
JoranHonig: Can you do `raise` here, that should allow us to keep the stack trace

mythril/support/loader.py
```diff
@@ -36,6 +43,9 @@ def read_storage(self, contract_address, index):
 
     def dynld(self, contract_address, dependency_address):
 
+        if not self.contract_loading:
+            raise ValueError("Cannot load contract when contract_loading flag is false")
```
JoranHonig: Why a value error here, and an Exception a while back?
In this case a specific exception might make sense, wdyt?

norhh: This exception is being caught in lot of other areas in laser, so we might need to import this exception from mythril to laser, should we do that?

b-mueller:
Sorry for messing up the formatting, you were quicker than me fixing it again :)

PR #648, opened 2018-10-31
analysis/integer: fix the overflow check of bit vector multiplication

hzzhang:
Differently than the bit vector addition, an overflowed bit vector
multiplication does not necessarily result in a value less than any
of its operands (e.g., the multiplication of two 8-bit vectors 0x7 and
4 produces 0xC). Fix this issue by using Z3 BVMulNoOverflow in the
overflow check of bit vector multiplication.


JoranHonig:
@hzzhang Awesome, I didn't know they updated the solver-z3 package. 
Can you update the dependencies, as the current dependency: `z3-solver>=4.5` does not ensure that BVMulNoOverflow is supported.

JoranHonig:
Oh, also could you do the same for addition with BVAddNoOverflow? :smile:

hzzhang:
Sure, I'll bump the minimal z3 version to 4.6 and add BVAddNoOverflow as well.

JoranHonig:
Can you make it 4.8, it seems they didn't release in between 4.5 and 4.8 so it should be equivalent

hzzhang:
Ok, let me change to 4.8.

PR #654, opened 2018-10-31
Related calldata and concretion improvements/fixes

nbanmp:
Includes:
Removing `ForAll` - Faster!
Minimizing `calldatasize` - No crazy long transactions.
Bugfixes

### DIFF
--- a/mythril/analysis/solver.py
+++ b/mythril/analysis/solver.py
@@ -1,4 +1,4 @@
-from z3 import Solver, simplify, sat, unknown, FuncInterp, UGE
+from z3 import Solver, simplify, sat, unknown, FuncInterp, UGE, Optimize
 from mythril.exceptions import UnsatError
 from mythril.laser.ethereum.transaction.transaction_models import (
     ContractCreationTransaction,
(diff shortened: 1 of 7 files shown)


mythril/laser/ethereum/state.py
```diff
@@ -56,28 +57,29 @@ def __init__(self, tx_id, starting_calldata=None):
             self.calldatasize = BitVec("{}_calldatasize".format(self.tx_id), 256)
             self.concrete = False
 
-        self.starting_calldata = starting_calldata or []
```
norhh: I think this should be reflected in `native_contracts.py`

nbanmp: All done, the modifications are reflected in `natives.py` now.

mythril/analysis/solver.py
```diff
@@ -62,6 +67,7 @@ def get_transaction_sequence(global_state, constraints):
     txs = {}
```
JoranHonig: Its not part of this pr, but I was wondering why the model used in the vulnerability check module isn't reused here

mythril/laser/ethereum/instructions.py
```diff
@@ -504,6 +505,8 @@ def calldatacopy_(self, global_state):
                         i_data + 1 if isinstance(i_data, int) else simplify(i_data + 1)
                     )
 
+                    environment.calldata.update_constraints(state)
```
JoranHonig: Why this? Didn't this already happen in the transactions initial state?

mythril/laser/ethereum/state.py
```diff
@@ -111,6 +113,9 @@ def __getitem__(self, item):
             except IndexError:
                 return BitVecVal(0, 8)
         else:
+            constraints = [Implies(item != 0, UGT(self.calldatasize, item))]
```
JoranHonig: Shouldn't it be:
```suggestion
            constraints = [Implies(self._calldata[item] != 0, UGT(self.calldatasize, item))]
```

PR #659, opened 2018-11-01
analysis/integer: fix the overflow check of bit vector multiplication

hzzhang:
Differently than the bit vector addition, an overflowed bit vector
multiplication does not necessarily result in a value less than any
of its operands (e.g., the multiplication of two 8-bit vectors 0x7 and
4 produces 0xC).

The previous attempt to fix this issue used Z3 API BVMulNoOverflow
which is only available in newer versions of Z3 and unfortunately
brought regressions. The fix in this commit implements a common
algorithm to check the overflow of symbolic bit vector multiplication
and does not reply on any new Z3 APIs.

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -8,6 +8,7 @@ from mythril.laser.ethereum.taint_analysis import TaintRunner
 import re
 import copy
 import logging
+import math
 
 """
 MODULE DESCRIPTION:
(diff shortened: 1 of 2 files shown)


JoranHonig:
@hzzhang Looks good, however it seems that we found the problem with z3 so we can use your old fix. Will get back to you on that

hzzhang:
That's a good news.

JoranHonig:
For reference #695

JoranHonig:
Closing this, because of #702

PR #660, opened 2018-11-01
Add gas simulation to LASER

### LINKED ISSUE #652: Mythril on a specific solc version isn't working
## Description
Currently mythril doesn't pass solc binary parameters to other functions, which makes execution of some contracts with fixed solc version fail.

## Screenshot
bug_report

dmuhs:
This PR will
- Refactor the transaction models into a more concise base type
- Add the respective gas costs to each instruction
- Dynamically calculate the costs for specific instructions
- Add up the gas costs in the machine state
- Discard any global state that has surpassed the gas limit

Right now the gas limit is set at a fix value of 8000000 based on current block gas limit statistics. If the concrete gas value cannot be calculated due to a symbolic value lying on the stack, the minimum usage for that instruction is assumed so that the resulting gas estimate is a lower bound of the gas that actually has to be invested in a transaction triggering the call graph path.

Closes #652


norhh:
Cool, Can you also make a few changes in the `/tests/laser/evm_testsuite/evm_test.py` , so that it also checks for the remainder of the gas after a transaction. That would be great.

norhh:
Instead of removing the state can you raise an OutOfGas Exception which inherits VM exception, So this can be used for gas tests in the VM tests.

dmuhs:
Checking the gas remainder is a tricky thing right now for two reasons:
- Some values are symbolic we aren't able to completely evaluate them. In that case I just add in the instruction's base gas cost. So what we end up is a lower bound.
- It seems like there's currently no way to get the global state from the Laser EVM object. Is that correct? Without it it doesn't seem possible to access the gas limit/usage of the machine state and the tx.

I played around a little with the fixtures and we could duplicate a valid one, set the gas value much lower and check that there's no open states left after the message call. Then we'd know for sure that (assuming the other case passes and it's a valid instruction) that an OOG error must have been triggered.

norhh:
we can just leave it to z3 i think regarding the symbolic gas additions etc, remainder gas will be in some symbolic form like `x+keccak_256[10]-10*calldata_10` .
Also LaserEvm has a list of global states in its worklist. so you can check that corresponding global state's machine state.

dmuhs:
@norhh In the context of the EVM tests the worklist is always empty after the message call has been executed. As for the gas calculation, should I then make the base cost a BitVec, add everything up and then solve? How would I go about dynamic costs such as
```python
def call_cost(value: int, contract_created: bool):
    if not isinstance(value, int):
        is_value_tx = False
    else:
        is_value_tx = True if value != 0 else False

    return 700 + is_value_tx * 9000 + contract_created * 25000
```

Maybe add the contract creation as implication? There's still the case that the `DEAD` function from the yellowpaper that checks whether a state for an address already exists isn't really available in our current view.

norhh:
regarding the messagecall, can you try returning the max and min gas for the message call execution? If something is popped from the worklist and nothing is added into it then we have reached a deadend for that branch, so we can just log it.

norhh:
Can you also add the `vmSha3Test` in the test_types list in `evm_test` it seems to be missed from the merge.

dmuhs:
Just had another look at the comments and tried to concreticize some thoughts on whether to do the gas calculation in a symbolic or concrete way:

| Symbolic | Concrete |
|-------------|-------------|
| \+ precise | \+ quick |
| \- overhead to solve | \+ easy to implement |
| \- handling of math operations (e.g. log, ceil) | \- imprecise (lower bound) |

I don't think the overhead of symbolic calculation is worth the value the feature can deliver. It'd add a lot more complexity to Mythril just to make the number a little more precise.
Sitting down a little more, it seems feasible however to calculate an upper bound for each dynamic gas cost function based on standards the EVM or Solidity set.

So I'll build in a min and max gas usage to avoid dealing with symbolic complexity or context and then give a range for each issue.

dmuhs:
```
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/transaction/concolic.py:52: in execute_message_call
    return laser_evm.exec()
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/svm.py:157: in exec
    new_states, op_code = self.execute_state(global_state)
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/svm.py:180: in execute_state
    global_state
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/instructions.py:284: in evaluate
    return instruction_mutator(global_state)
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/instructions.py:243: in wrapper
    new_global_states = self.call_on_state_copy(func, func_obj, global_state)
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/instructions.py:215: in call_on_state_copy
    return func(func_obj, global_state_copy)
.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/instructions.py:747: in sha3_
    state.mem_extend(index, length)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

self = <mythril.laser.ethereum.state.MachineState object at 0x7f8382948898>, start = 282695530364541492...9945699220681916415, size = 2

    def mem_extend(self, start, size):
        """
        Extends the memory of this machine state
        :param start: Start of memory extension
        :param size: Size of memory extension
        """
        if self.memory_size > start + size:
            return
        m_extend = start + size - self.memory_size
>       self.memory.extend(bytearray(m_extend))
E       OverflowError: cannot fit 'int' into an index-sized integer

.tox/py36/lib/python3.6/site-packages/mythril/laser/ethereum/state.py:344: OverflowError
```

It seems that the large inputs of the SHA3 tests break the machine stack extension method. This usually happens when the int value grows too large and doesn't fit into an index integer anymore, because Python scales up the size internally. The gas calculation doesn't touch the machine state's memory, so that's a weird bug. II'll check it out. :+1:

norhh:
When memory is extended then I think gas is consumed right? in that case this test is expected to result in an OutOfGas exception right?

PS: previously it worked because the post_condition check in the test was above to remove all the OOG exceptions which were expected to be caught.

tests/laser/evm_testsuite/evm_test.py
```diff
@@ -64,22 +64,23 @@ def test_vmtest(
...
         origin_address=action["origin"],
         code=action["code"][2:],
-        gas=action["gas"],
+        gas_limit=int(action["gas"], 16),
         data=binascii.a2b_hex(action["data"][2:]),
         gas_price=int(action["gasPrice"], 16),
         value=int(action["value"], 16),
     )
 
+    if post_condition == {}:
```
norhh: there should also be a check to see whether any vmException was raised or not right?

mythril/laser/ethereum/instructions.py
```diff
@@ -49,6 +50,154 @@
 keccak_function_manager = KeccakFunctionManager()
 
 
+# opcode -> (min_gas, max_gas)
+OPCODE_COST_FUNCTIONS = {
```
norhh: Can you move this to a seperate file related to gas consumption?

tests/laser/evm_testsuite/evm_test.py
dmuhs: Internally VmExceptions are caught, so we don't have an indicator on the test level. If we go for testing the instructions in the future and always go through the `Instruction.evaluate` function, we could check for all of these

dmuhs:
> When memory is extended then I think gas is consumed right? in that case this test is expected to result in an OutOfGas exception right?
> 
> PS: previously it worked because the post_condition check in the test was above to remove all the OOG exceptions which were expected to be caught.

@norhh In the current state to avoid handling symbolic values in gas cost functions, we simply estimate the storage cost to a fixed value. Regarding the test cases, it seems that some legitimate cases don't have a post condition, e.g. in the SHA3 tests. It's valid (and often fairly expensive) calculations, but most of them bring enough gas along to account for them. Because we don't scale up the gas cost with the input size of `SHA3`, some tests don't run out of gas. Together with the valid ones, such as `sha3_BigOffset.json` they will reserve memory at a large index. `mem_extend` will then fill up a list with a ton of zeros and try to enter the value at the specified location. The value of that location can be bigger than a `long`, however, which results in Python failing to use it as an index value. Thus the OverflowError.

norhh:
Gas is also consumed while extending the memory right?I mean the `mem_extend` function also consumes gas.

tests/laser/evm_testsuite/evm_test.py
```diff
@@ -64,40 +86,51 @@ def test_vmtest(
...
+        # this currently exceeds our estimations
+        gas_min_max = [
+            (s.mstate.min_gas_used, s.mstate.max_gas_used) for s in final_states
+        ]
+        gas_ranges = [g[0] <= gas_used <= g[1] for g in gas_min_max]
+        assert all(map(lambda g: g[0] <= g[1], gas_min_max))
+        assert any(gas_ranges)
+
+    if any((v in test_name for v in ["error", "oog"])):
+        # no more work to do if error happens or out of gas
```
norhh: If this happens and also the post_condition={} only then the test is true.
if OOG/vmException is thrown and post_condtion!={} or post_condition={} and nothing is thrown then the test should raise an error.

dmuhs: We don't get to see the internal exceptions, because they're caught [here](https://github.com/ConsenSys/mythril-classic/blob/5a7802643bf89d6c834da71dca25052d0f4ce16a/mythril/laser/ethereum/svm.py#L187). As mentioned earlier, we could cover that behaviour in unit tests for the `Instruction` class by testing closer to the `evaluate` method. Apart from some crazy mocking I wouldn't know how to dig it out of the scenario of the VM Tests.

dmuhs:
> Gas is also consumed while extending the memory right?I mean the `mem_extend` function also consumes gas.

Yeah. Because we don't always have concrete values, we cannot give a concrete value for the gas cost of extending the memory. That's why in the gas cost dict you'll find something like this:
```python
...
    # assume 1KB memory r/w cost as upper bound
    "MLOAD": (3, 96),
    "MSTORE": (3, 98),
    "MSTORE8": (3, 98),
    # assume 64 byte r/w cost as upper bound
    "SLOAD": (400, 400),
    "SSTORE": (5000, 25000),
...
```

mythril/laser/ethereum/instructions.py
dmuhs: Fixed :+1:

tests/laser/evm_testsuite/evm_test.py
norhh: Should we be catching the vmException in mythril? all ethereum VMs directly raise that to say that something is wrong with the bytecode. @JoranHonig  what do you think?

norhh:
native contracts also consume gas right?

nbanmp:
How are you planning to handle the `gas`, `gaslimit`, and `gasprice` opcodes? `gasprice` can probably remain an unconstrained BitVec, but we have values for `gas` and `gaslimit` now.

The first thing that came to mind when glancing at this was a [ctf challenge](https://blog.ret2.io/2018/05/16/practical-eth-decompilation/) that constrained the amount of gas you could use, and you had to construct an input that did not consume too much gas (Note that this ctf challenge won't work too well with mythril, it uses hashes and other hard to symbolically handle things). This is a very specific thing that would be more difficult to support, but it's still something to keep in mind because sometimes contracts do have different behavior depending on the amount of gas used.

dmuhs:
@norhh Yes, native contracts also consume gas. This should be handled by the `Instructions.call_` function, shouldn't it? At least that seems to be the only place where `native_contracts` from `natives.py` is called.

@nbanmp I agree, `gas` and `gaslimit` could carry some value, I didn't realize that. Right now we have the minimal and maximal gas estimation. I could let them return the min value to make sure we don't miss anything by estimating too high. Will test that :+1:

norhh:
@dmuhs each native contract consumes seperate gas + gas for the call right?

norhh:
@dmuhs 
`they will reserve memory at a large index. mem_extend will then fill up a list with a ton of zeros and try to enter the value at the specified location`
exactly!, this operation consumes some killer amount of gas:
`(new_size**2)*512 + new_size*3 - ((prev_size**2)*512 +prev_size*3)` .
This is used to avoid killing EVM's memory.

dmuhs:
@norhh Due to the potentially symbolic size values we aren't really able to calculate that (without using SMT solving at least). Where did you get the formula from, btw? I used [this](https://docs.google.com/spreadsheets/d/1n6mRqkBz3iWcOlRem_mO09GtSKEKrAsfO7Frgx18pNU/edit#gid=0) for the simpler estimations and the yellowpaper for the rest.

As for the native contracts, I didn't find any resource indicating that there is separate gas consumption apart from the instructions how we already implement them. Generally you don't find that many resources for that topic. If you have anything lying around, I'd appreciate it.

norhh:
what better resource can anyone have which is better than ethereum's code itself :wink: . [goethereum](https://github.com/ethereum/go-ethereum/blob/master/core/vm/contracts.go), [pyethereum](https://github.com/ethereum/pyethereum/blob/develop/ethereum/specials.py) links contain the respective native contracts files . And also list.extend() can't take symbolic values(will raise some TypeError which might be caught somewhere), so you can just check that for  concrete values or the values which can be concertized directly(by using util.get_concrete_int() ).

mythril/laser/ethereum/instructions.py
```diff
@@ -76,11 +79,32 @@ def increment_states_pc(self, states: List[GlobalState]) -> List[GlobalState]:
...
+        global_state.mstate.max_gas_used += max_gas
+        return global_state
+
     def __call__(self, func: Callable) -> Callable:
         def wrapper(
             func_obj: "Instruction", global_state: GlobalState
         ) -> List[GlobalState]:
             new_global_states = self.call_on_state_copy(func, func_obj, global_state)
+            new_global_states = [
+                self.accumulate_gas(state) for state in new_global_states
```
JoranHonig: Do you pay gas before or after the instruction execution?

mythril/laser/ethereum/instructions.py
```diff
@@ -1012,15 +1046,22 @@ def jump_(self, global_state: GlobalState) -> List[GlobalState]:
             )
 
         new_state = copy(global_state)
+        # add JUMP gas cost
+        min_gas, max_gas = OPCODE_GAS["JUMP"]
```
JoranHonig: Shouldn't you still pay gas, even if the jump is invalid?

dmuhs: Yeah, true. On an invalid jump, the whole gas is actually burnt to punish the user. From the yellowpaper:

> Just as with contract creation, if the execution halts in an exceptional fashion (i.e.  due to an exhausted gas supply, stack underflow, invalid jump destination or invalid instruction), then no gas is refunded to the caller and the state is reverted to the point immediately prior to balance transfer (i.e. σ).

mythril/laser/ethereum/instructions.py
```diff
@@ -1332,15 +1391,16 @@ def callcode_(self, global_state: GlobalState) -> List[GlobalState]:
...
 
         transaction = MessageCallTransaction(
-            global_state.world_state,
-            environment.active_account,
-            environment.address,
-            call_data=call_data,
+            world_state=global_state.world_state,
             gas_price=environment.gasprice,
-            call_value=value,
+            gas_limit=2300,  # G_callstipend
```
JoranHonig: isn't the stipend a lower bound? Or am I misunderstanding that?

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -40,53 +40,74 @@ def __init__(
...
+    """Basic transaction class holding common data."""
 
     def __init__(
         self,
         world_state: WorldState,
-        callee_account: Account,
-        caller: ExprRef,
+        callee_account: Account = None,
+        caller: ExprRef = None,
         call_data=None,
```
JoranHonig: Not every transaction has calldata (creation transactions)

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -40,53 +40,74 @@ def __init__(
...
+        self.code = code
+
+        self.caller = caller
+        self.callee_account = callee_account
+        self.call_data = (
+            call_data
+            if isinstance(call_data, Calldata)
+            else Calldata(self.id, call_data)
         )
         self.call_data_type = (
```
JoranHonig: I think this is an unused variable now. Can you check that (and remove it if so)

mythril/laser/ethereum/svm.py
```diff
@@ -154,17 +154,18 @@ def _get_covered_instructions(self) -> int:
             )
         return total_covered_instructions
 
-    def exec(self, create=False) -> None:
+    def exec(self, create=False, track_gas=False) -> None:
```
JoranHonig: Now the return type is incorrect

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -178,11 +151,7 @@ def initial_global_state(self) -> GlobalState:
             self.code,
             calldata_type=self.call_data_type,
         )
-
-        global_state = GlobalState(self.world_state, environment, None)
-        global_state.environment.active_function_name = "constructor"
-
-        return global_state
+        return super().initial_global_state_from_environment(environment)
```
JoranHonig: what about active_function = "constructor"

mythril/laser/ethereum/transaction/transaction_models.py
JoranHonig: However, I do see that the previous implementation also had this in there

mythril/laser/ethereum/transaction/transaction_models.py
dmuhs: That's set as the default argument for `initial_global_state_from_environment`. I'll make it explicit, though for better readability

mythril/laser/ethereum/transaction/transaction_models.py
JoranHonig: Wouldn't it be more correct to make call_data a variable only in the message call transaction?
Although  your solution would probably also result in correct behaviour

mythril/laser/ethereum/instructions.py
dmuhs: Yeah. The gas stipend is actually deducted from the call gas costs for a non-zero value transfer. That doesn't really affect the gas limit. That rather comes from the call parameters. I've added the gas value, which is popped from the stack in the latest changes :+1:

mythril/laser/ethereum/instructions.py
dmuhs: If you look at the pyethereum VM code, it seems like the gas is checked after the elements have been popped from the stack and before any state changes are made, e.g. [here](https://github.com/ethereum/pyethereum/blob/develop/ethereum/vm.py#L763). That shouldn't really affect us as we discard the global states anyway. At points where expensive memory operations can trigger an overflow, there is either an explicit gas check in the mstate's `mem_extend`, or in the state changing function.

mythril/laser/ethereum/instructions.py
JoranHonig: So this should in theory be above the jumpdest check? (Not that it really matters for laser)

mythril/laser/ethereum/svm.py
JoranHonig: Why is this resolved>

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -178,11 +149,9 @@ def initial_global_state(self) -> GlobalState:
             self.code,
             calldata_type=self.call_data_type,
```
JoranHonig: There should also be no calldatatype

mythril/laser/ethereum/svm.py
JoranHonig: The latest commits hadn't shown yet

mythril/laser/ethereum/transaction/transaction_models.py
JoranHonig: Keep it there (I think its actually an unused variable, so we can remove it in an other pr)

mythril/laser/ethereum/instructions.py
```diff
@@ -608,8 +608,15 @@ def sha3_(self, global_state: GlobalState) -> List[GlobalState]:
             if is_expr(op0):
                 op0 = simplify(op0)
             state.stack.append(BitVec("KECCAC_mem[" + str(op0) + "]", 256))
+            state.min_gas_used += OPCODE_GAS["SHA3"][0]
+            state.max_gas_used += OPCODE_GAS["SHA3"][1]
             return [global_state]
 
+        min_gas, max_gas = OPCODE_GAS["SHA3_FUNC"](state.memory_size + index, length)
```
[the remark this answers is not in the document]
JoranHonig: @norhh can you confirm that this is resolved now?

mythril/laser/ethereum/svm.py
dmuhs: https://github.com/ConsenSys/mythril-classic/pull/660/commits/e47c7a35b268a3842342aa58b866cdf14d986b83
I updated the type hint in this commit

mythril/laser/ethereum/transaction/transaction_models.py
dmuhs: :+1: Not going to touch it

mythril/laser/ethereum/instructions.py
dmuhs: It would burn the gas after jumpdest check in case it failed :)

mythril/laser/ethereum/gas.py
```diff
@@ -0,0 +1,175 @@
...
+        gas_value = opcodes.GRIPEMD160BASE + word_num * opcodes.GRIPEMD160WORD
+    elif contract == "identity":
+        gas_value = opcodes.GIDENTITYBASE + word_num * opcodes.GIDENTITYWORD
+    else:
+        raise ValueError("Unknown contract type {}".format(contract))
+    return gas_value, gas_value
+
+
+def calculate_sha3_gas(length: int):
+    gas_val = 30 + opcodes.GSHA3WORD * ceil32(length)
```
norhh: it's gas per word right? shouldn't it be ceil32(length)//32 or something? I am not totally sure

dmuhs: Good catch. Fixed it, testing now and then I'll push it with the latest merge :+1:

JoranHonig: It doesn't seem to be fixed yet @dmuhs

JoranHonig: Now it is :+1:

PR #664, opened 2018-11-03
Add EnvironmentInfo tests

norhh:
- Added EnvironmentalInfo tests
- Made some fixes in the corresponding places like the Calldata class and few instructions like calldataload etc


tests/laser/evm_testsuite/evm_test.py
```diff
@@ -95,9 +98,13 @@ def test_vmtest(
 
         for index, value in details["storage"].items():
             expected = int(value, 16)
-            if type(account.storage[int(index, 16)]) != int:
-                actual = model.eval(account.storage[int(index, 16)])
+            actual = account.storage[int(index, 16)]
+            if type(actual) not in [str, bytes, int]:
```
JoranHonig: in this case actual is an ExprRef right?

mythril/laser/ethereum/util.py
```diff
@@ -89,10 +89,10 @@ def get_concrete_int(item):
 
 
 def concrete_int_from_bytes(_bytes, start_index):
-
-    # logging.debug("-- concrete_int_from_bytes: " + str(_bytes[start_index:start_index+32]))
+    _bytes = [
```
JoranHonig: Maybe rename to  `concrete_bytes`

mythril/laser/ethereum/util.py
```diff
@@ -89,10 +89,10 @@ def get_concrete_int(item):
...
+        _byte.as_long() if type(_byte) == BitVecNumRef else _byte for _byte in _bytes
+    ]
     b = _bytes[start_index : start_index + 32]
```
JoranHonig: While you are busy can you make the variable names in this function a bit more verbose. integer_bytes seems more clear than `b` And the int.from_bytes can directly be returned vs creating a variable val

mythril/laser/ethereum/state.py
```diff
@@ -90,19 +90,23 @@ def get_word_at(self, index: int):
 
     def __getitem__(self, item):
         if isinstance(item, slice):
+            start, step, stop = item.start, item.step, item.stop
             try:
+                if start is None:
```
JoranHonig: The following would be possible here too right?
```suggestion
                start = start or 0
```

JoranHonig: You could even make a oneliner:
```
start, step, stop = item.start or 0, item.step or 1, item.stop or self.calldatasize
```

JoranHonig: Or just make it three lines

mythril/laser/ethereum/state.py
```diff
@@ -90,19 +90,23 @@ def get_word_at(self, index: int):
 
     def __getitem__(self, item):
         if isinstance(item, slice):
+            start, step, stop = item.start, item.step, item.stop
```
JoranHonig: What required the change here?

mythril/laser/ethereum/instructions.py
```diff
@@ -775,8 +775,8 @@ def mload_(self, global_state):
             data = global_state.new_bitvec("mem[" + str(simplify(op0)) + "]", 256)
             state.stack.append(data)
             return [global_state]
-
         try:
+            state.mem_extend(offset, 32)
             data = util.concrete_int_from_bytes(state.memory, offset)
         except IndexError:  # Memory slot not allocated
```
JoranHonig: This except isn't needed anymore

mythril/laser/ethereum/state.py
norhh: if step in slice is unintialized then it is None, in many cases it was uninitialized, which led to Z3Exception

mythril/laser/ethereum/state.py
JoranHonig: Sad

mythril/laser/ethereum/instructions.py
```diff
@@ -775,11 +775,9 @@ def mload_(self, global_state):
...
             state.stack.append(data)
             return [global_state]
-
         try:
+            state.mem_extend(offset, 32)
             data = util.concrete_int_from_bytes(state.memory, offset)
-        except IndexError:  # Memory slot not allocated
-            data = global_state.new_bitvec("mem[" + str(offset) + "]", 256)
         except TypeError:  # Symbolic memory
             data = state.memory[offset]
```
JoranHonig: Should this use a slice?

mythril/laser/ethereum/instructions.py
```diff
@@ -811,7 +809,7 @@ def mstore_(self, global_state):
         try:
             # Attempt to concretize value
             _bytes = util.concrete_int_to_bytes(value)
-            state.memory[mstart : mstart + len(_bytes)] = _bytes
+            state.memory[mstart : mstart + 32] = _bytes
```
JoranHonig: What happens when _bytes does not have 32 bytes?

mythril/laser/ethereum/instructions.py
```diff
@@ -811,7 +809,7 @@ def mstore_(self, global_state):
         try:
             # Attempt to concretize value
             _bytes = util.concrete_int_to_bytes(value)
```
JoranHonig: Extract should be used here vs concretizing and getting bytes that way. @norhh Could you create an issue for this?

mythril/laser/ethereum/instructions.py
JoranHonig:
```suggestion
            data = state.memory[offset:offset+32]
```

mythril/laser/ethereum/instructions.py
norhh: fortunately that doesn't happen as util.concrete_int_to_bytes gives a 32 length word.So this line was obsolete.Maybe adding an assert there might be a good thing to do

mythril/laser/ethereum/instructions.py
norhh: unfortunately everything should be concatenated there right?Straight concatenation lead to some failing tests which raised some Z3 exception, it might be good to raise a separate issue here.What do you think

mythril/laser/ethereum/instructions.py
norhh: added #684

mythril/laser/ethereum/instructions.py
norhh: can you explain why extract should be used? Created an issue #685

mythril/laser/ethereum/instructions.py
JoranHonig: Sure, go ahead and create that

mythril/laser/ethereum/instructions.py
JoranHonig: The current process to get bytes is:
symbol -> concrete int -> bytes

Which breaks for symbolic values, in which case could do:
symbol -> [Extract(0, 7, value), ....]

PR #667, opened 2018-11-05
Solver per branch

norhh:


mythril/laser/ethereum/instructions.py
```diff
@@ -1056,6 +1056,7 @@ def jumpi_(self, global_state):
                 states.append(new_state)
             else:
                 logging.debug("Pruned unreachable states.")
+        del global_state
```
JoranHonig: won't this be automatically be garbage collected?

norhh: I think so but I am not totally sure, I added del everywhere as a desperate attempt to reduce memory and this got left off.

mythril/analysis/modules/dependence_on_predictable_vars.py
```diff
@@ -48,7 +48,7 @@ def execute(statespace):
 
         found = []
         for var in vars:
-            for constraint in call.node.constraints + [call.to]:
+            for constraint in call.node.constraints[:] + [call.to]:
```
JoranHonig: Is the [:] necessary here? (same for ln 78)

JoranHonig:
@norhh maybe :innocent:

Issue #226, opened 2018-05-31
Refactor: Rebuild laser and mythril to work with callbacks[MYTHC-3]


JoranHonig:
## Description
Currently, laser-ethereum ( the symbolic analysis backend ) performs symbolic execution on a smart contract and the result of this execution (the statespace) is passed to analysis modules to be analyzed.

In the future, we want to remodel the analysis of smart contracts to work using callbacks. For example, the integer module would hook ADD and MUL instructions to find locations that can overflow, while also hooking JUMPI and SSTORE to find if the values are used there.

This means that analysis and execution happen simultaneously and we don't do any double work

### Benefits
- Only need to keep the last state of a path in memory, rather than needing to keep a copy of all historical states at each instruction

child of #203

[[[MYTHC-3]](https://diligence.atlassian.net/browse/MYTHC-3)](https://diligence.atlassian.net/browse/MYTHC-3)

JoranHonig:
Initially, we should let laser retain the passed states. 
Doing this will allow the current analysis modules to keep working, making their transition easier.

JoranHonig:
This implementation also requires the taint analysis to be changed

FFengIll:
Some extra suggestion: (for I am doing some work like it)
Makes stack element as a individual class like Slot/Element or something.
The possible benefits are:
* python `magic` to override some operators, e.g. __add__, __lt___ and so on
```python
class Slot:
    def __iadd__(self,other):
       #inplace add
       self.value = self.value + other.value
       return self.
```
* python `property` to insulate solver and some others, e.g. (main)
```python
if type(condi) == BoolRef
# can be changed to
if condi.is_bool_ref()
```
* make it possible to use mythril but work for another vm / bytecode (minor)
* may more

I am not sure which one is the better, but wish to have a quick discuss.

JoranHonig:
After a lot of triaging I built the following example code that has the same problem:
```
pragma solidity ^0.4.16;
/**
 * @title SafeMath
 * @dev Math operations with safety checks that throw on error
 */
library SafeMath {
  function sub(uint256 a, uint256 b) internal constant returns (uint256) {
    assert(b <= a);
    return a - b;
  }

  function add(uint256 a, uint256 b) internal constant returns (uint256) {
    uint256 c = a + b;
    assert(c >= a);
    return c;
  }
}

contract IntegerOverflow2  {
    using SafeMath for uint256;
    uint256 public count = 7;
    mapping(address => uint256) balances;


[...]

PR #677, opened 2018-11-08
Add a symbol manager

hzzhang:
The symbol manager provides a global symbol factory to create symbols
and manage properties attached symbols. A symbol created from the
symbol factory is identified by its name and composed of
- the raw symbol object from Z3,
- additional properties attached to that symbol. The properties are
  managed and used in Mythril, and are never used by the constraint
  solvers.

Three APIs are provided by mythril.laser.ethereum.symbol_manager:
1. sym_get_bv(name, width): create or get a bit vector symbol of the
   specified name and bit width;
2. sym_set_prop(sym_name, prop_name, prop_value): set a property of
   a symbol;
3. sym_get_prop(sym_name, prop_name): get a property of a symbol.

It can be used to address the requirement of taint analysis in #533 
without a SMT abstraction layer.


mythril/laser/ethereum/symbol_manager.py
```diff
@@ -0,0 +1,180 @@
+from z3 import BitVec, BitVecRef, is_bv
+
+
+class Sym:
```
JoranHonig: What do you think about renaming this to Symbol?

hzzhang: It's fine. This code is picked from my SMT abstraction work where I chose 'Sym' to just avoid name conflict with pySMT which happens to provide a function 'Symbol'. Anyway, the conflict can be avoided by using a full name of pySMT Symbol in the future.

JoranHonig:
So in this case to do taint analysis the instructions would need to have specific taint propagation logic in them.

Ideally what I would like is that laser does not work with the raw symbols but with the Symbol object, and the implementation of symbol defines how taint is propagated. 

For example some very trivial version of __add__()
```python3
define __add__(self, other: Symbol):
    new_symbol = add_symbol_logic(self, other)
    new_symbol._props = self._props
```
What do you think about implementing it in such a way?

hzzhang:
> Also (correct me if I'm wrong) but it doesn't seem to express properties over expressions

Right, it only attach properties to symbols or constant in Z3 terms.

> So in this case to do taint analysis the instructions would need to have specific taint propagation logic in them.

Right. On each instruction, the taint analysis will need to traverse every operand to check whether it contains any tainted symbols.

> 
> Ideally what I would like is that laser does not work with the raw symbols but with the Symbol object, and the implementation of symbol defines how taint is propagated.

Then, we may need to wrap every types of Z3 expressions used in Mythril, i.e., a SMT abstraction is still needed. 

> 
> For example some very trivial version of **add**()
> 
> ```python
> define __add__(self, other: Symbol):
>     new_symbol = add_symbol_logic(self, other)
>     new_symbol._props = self._props
> ```

No quite understand how this works. To my understanding, during the symbolic execution, laser creates all symbols and expressions, which are used in the later taint analysis. But during the symbolic execution, laser does not know which symbols will be interested by the taint analysis, so it does not know where the taint property needs to be set.

> 
> What do you think about implementing it in such a way?
>

JoranHonig:
@hzzhang Right, that is currently how taint analysis works. However in light of #226, we want to execute the analysis modules along with symbolic execution, in which case we want to stop using the old taint module and move towards tracking taint during the execution.

hzzhang:
In such cases, we may still need a complete SMT abstraction that can at least cover all types of Z3 expressions used by the current laser. Neither pySMT nor claripy can satisfy the requirement. It looks like we may need to implement another one from scratch, though we can make it very simple and only support Z3 at the beginning.

Unfortunately, I'm very busy recently and will be in the same status in the next 3-4 weeks, so I will not have enough time to working on it :(

JoranHonig:
@hzzhang Thanks for the update! In that case we'll find someone else to pick this up

Issue #675, opened 2018-11-08
Add a special case for passing original calldata with symbolic length


b-mueller:
## Description

As described in #674, Mythril currently doesn't support the delegatecall proxy pattern where the original calldata is passed through. This is because we don't support calldata with a symbolic offset or length.

I suppose that we could add a special case for `offset = 0` and `length = calldatasize` where we provide the original calldata to the callee.

nbanmp:
We likely won't be able to handle it "correctly" with the current memory design. I forgot that calldata is copied to memory, then copied from memory to the new calldata. I'll see what I can do about getting a good special case.

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__This issue now has a funding of 2.5 ETH (529.46 USD @ $211.78/ETH)  attached to it as part of the ConsenSys fund.__

 * If you would like to work on this issue you can 'start work' [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ConsenSys/mythril-classic/675/1721).
* Want to chip in? Add your own contribution [here](https://gitcoin.co/issue/ConsenSys/mythril-classic/675/1721).
* Questions? Checkout <a href='https://gitcoin.co/help'>Gitcoin Help</a> or the <a href='https://gitcoin.co/slack'>Gitcoin Slack</a>
* $57,217.94 more funded OSS Work available on the [Gitcoin Issue Explorer](https://gitcoin.co/explorer)

gitcoinbot:
Issue Status: 1. **Open** 2. Started 3. Submitted 4. Done 

<hr>

__Workers have applied to start work__.


These users each claimed they can complete the work by 12 months from now.
Please review their action plans below:


**1) [nbanmp](https://gitcoin.co/profile/nbanmp) has applied to start work _(Funders only: [approve worker](https://gitcoin.co/issue/ConsenSys/mythril-classic/675/1721?mutate_worker_action=approve&worker=nbanmp) | [reject worker](https://gitcoin.co/issue/ConsenSys/mythril-classic/675/1721?mutate_worker_action=reject&worker=nbanmp))_.**

Add a special case: Check if the offset is dependent only calldatasize, and if so, it probably is using the delegatecall pattern.

Learn more [on the Gitcoin Issue Details page](https://gitcoin.co/issue/ConsenSys/mythril-classic/675/1721).

gitcoinbot:
[...]

PR #683, opened 2018-11-09
Implement improvement for #675

nbanmp:


mythril/laser/ethereum/call.py
```diff
@@ -149,28 +147,44 @@ def get_call_data(global_state, memory_start, memory_size, pad=True):
...
+            call_data = global_state.environment.calldata
+        else:
+            # TODO: This only allows for either fully concrete or fully symbolic calldata.
+            # Improve management of memory and callata to support a mix between both types.
+            calldata_from_mem = state.memory[
+                util.get_concrete_int(memory_start) : util.get_concrete_int(
+                    memory_start + memory_size
+                )
+            ]
+            i = 0
```
JoranHonig: Maybe we could put the logic determining starting_calldata in a separate function, keeps this function from growing to big

mythril/laser/ethereum/call.py
```diff
@@ -149,28 +147,44 @@ def get_call_data(global_state, memory_start, memory_size, pad=True):
...
+            starting_calldata = []
+            while i < len(calldata_from_mem):
+                elem = calldata_from_mem[i]
+                if isinstance(elem, int):
+                    starting_calldata.append(elem)
```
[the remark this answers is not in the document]
nbanmp: concrete_int_to_bytes always returns a word.

The goal of this while loop is to copy each byte from memory to the new `starting_calldata`, and not die when a word happens to be in memory rather than it being properly byte indexed. Have we fixed that yet? If so, then we can reduce the entire while loop to one or two lines.

(side note, this wasn't code added in this pr, it was just all indented in this pr)

mythril/laser/ethereum/call.py
nbanmp: I think it's small enough (and will only get smaller with fixes taking into account correct memory byte indexing) as is, and I don't have any reason to expect it to get more complicated or grow, or that it will be used elsewhere. Currently that logic is ~11 lines, and it will probably decrease to ~2.

mythril/laser/ethereum/call.py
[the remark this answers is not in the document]
JoranHonig: I think that has been fixed, can you try reducing the while loop then? That also resolves the previous comment

mythril/laser/ethereum/call.py
JoranHonig: Well, IMO this function has a fair bit of nesting and is getting quite lengthy. However, as you say this can be improved with the correct memory implementation

b-mueller:
Moved to #739

Issue #607, opened 2018-10-23
fix --solc-args for truffle


norhh:
## Description
truffle option isn't working for solc-args, for more info refer #497

norhh:
@roynalnaruto can you check it with the current master, i think this was fixed 2-3 days back?

norhh:
or the latest version of `v0.18.13`

norhh:
@roynalnaruto I am closing this issue as this was recently fixed in another issue, If you still have an issue with mythril then feel free to open the issue(or comment requesting to open the issue).

PR #686, opened 2018-11-10
Fix truffle source mappings and signatures

### LINKED ISSUE #597: File outside of allowed directories
## Description

If i run myth on our truffle project i get `File outside of allowed directories`. 

```
/Users/sebastian/ocean/keeper-contracts/contracts/auth/OceanAuth.sol:3:1: Error: Source "/Users/sebastian/ocean/keeper-contracts/contracts/OceanMarket.sol" not found: File outside of allowed directories.
import '../OceanMarket.sol';
^--------------------------^
```
I am using the most recent version 18.12. It was working with 18.11.

$ myth -V
Mythril version v0.18.12

## How to Reproduce

clone 
`https://github.com/oceanprotocol/keeper-contracts`

run 
`myth -x --truffle`

I also tried --sold-args "--allow-paths xxx" but got the same results.

## Expected behavior

should output the results

## Environment

macOS 10.13.6

$ python --version
Python 3.7.0

$ pip --version
pip 18.1 from /usr/local/lib/python3.7/site-packages/pip (python 3.7)

$ solc --version
solc, the solidity compiler commandline interface
Version: 0.4.25+commit.59dbf8f1.Darwin.appleclang

norhh:
Quick fixes to truffle module.
fixes #597 , #607

### DIFF
--- a/mythril/support/truffle.py
+++ b/mythril/support/truffle.py
@@ -4,6 +4,7 @@ import re
 import sys
 import json
 import logging
+from ethereum.utils import sha3
 from mythril.ether.ethcontract import ETHContract
 from mythril.ether.soliditycontract import SourceMapping
 from mythril.exceptions import CriticalError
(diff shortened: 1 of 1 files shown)


dmuhs:
Nice work. Both issues are fixed for me with this - even though I could not verify with https://github.com/oceanprotocol/keeper-contracts to the end, because there's some crazy memory write in one of the contracts that makes my RAM overflow.
Works as expected on other truffle projects, though :+1:

norhh:
Same here, crashed due to ram overflow. :cry:

PR #687, opened 2018-11-11
Refactor logging and error messages (slightly)

b-mueller:
`-v1` and `-v2` were inconsistent. Change debug levels so that `-v1` only outputs information related to progress and coverage.

### DIFF
--- a/mythril/analysis/modules/external_calls.py
+++ b/mythril/analysis/modules/external_calls.py
@@ -64,7 +64,7 @@ def execute(statespace):
 
         if call.type == "CALL":
 
-            logging.info(
+            logging.debug(
                 "[EXTERNAL_CALLS] Call to: %s, value = %s, gas = %s"
                 % (str(call.to), str(call.value), str(call.gas))
             )
--- a/mythril/analysis/solver.py
+++ b/mythril/analysis/solver.py
@@ -21,7 +21,7 @@ def get_model(constraints, minimize=(), maximize=()):
     if result == sat:
         return s.model()
     elif result == unknown:
-        logging.info("Timeout encountered while solving expression using z3")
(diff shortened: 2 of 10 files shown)


norhh:
there is also the line 179 in call.py `logging.info("Unsupported symbolic calldata offset")` that should be in debug right?(I am no expert here)
And also I think there are some logging.info statements in mythril.py, are we ignoring them as they are of one time display?

dmuhs:
I agree with Nikhil here. In `mythril.py` there are some statements that aren't really informational but rather helpful for debugging:
- [mythril.py:227](https://github.com/ConsenSys/mythril-classic/blob/debug_levels/mythril/mythril.py#L227)
- [mythril.py:246](https://github.com/ConsenSys/mythril-classic/blob/debug_levels/mythril/mythril.py#L246)
- [mythril.py:281](https://github.com/ConsenSys/mythril-classic/blob/debug_levels/mythril/mythril.py#L281)
- [mythril.py:287](https://github.com/ConsenSys/mythril-classic/blob/debug_levels/mythril/mythril.py#L287)

With some statements I'm also not sure whether `logging.error` would be a better pick, e.g. [mythril.py:417](https://github.com/ConsenSys/mythril-classic/blob/debug_levels/mythril/mythril.py#L417)

I'd actually love to see an overhaul of the logging system in general. Currently we're only using the default root logger. With more components, especially when we rely on callbacks soon, the log output will become more and more cluttered. Having names loggers (with different configurable levels) can be of great help there. A use case would be the development of new analysis modules. Each module can have its own logger and if you want to debug it, you can simply enable it without having other module or core component logs in the way.

This is obviously way too broad for this PR, but plays in the same direction.

Issue #685, opened 2018-11-10
Refactor: Memory to z3 array


norhh:
## Description
Currently the mload and mstore try to get concrete bytes from and to memory. This needs to be changed so it allows symbolic memory here.

Example:
```python
try:
    state.mem_extend(offset, 32)
    data = util.concrete_int_from_bytes(state.memory, offset)
```

Additionally currently memory is implemented as a python list of bytes. Refactoring the module as a z3 array will probably be more flexible for our requirements. Additionally by implementing a Memory class we can make the code independent of the specific memory class

### Requirements
- Memory is modeled Similar to Calldata and uses a z3 K type internally. 
  - This will allow us to easily support symbolic reads.
- Memory locations are initialized to 0
- Memory class implementation implements __getitem__ for both slices and indices
- mem_extend shouldn't be necessary with a z3 K type array
- calls.py is refactored to work with new memory class
- instructions.py (mload and mstore) are refactored to work with new memory class

**Out of scope**
- support symbolic writes

PR #703, opened 2018-11-15
Refactor: Memory to z3 Array (Issue #685)

nbanmp:
Memory is now byte indexed
Memory is a z3 Array

Still need to write unit tests for new memory

### DIFF
--- a/mythril/laser/ethereum/state/memory.py
+++ b/mythril/laser/ethereum/state/memory.py
@@ -75,10 +75,8 @@ class Memory:
                     raise IndexError("Invalid Memory Slice")
                 if step is None:
                     step = 1
-                dataparts = []
 
-                for i in range(start, stop, step):
-                    dataparts.append(self[i])
+                dataparts = [self[i] for i in range(start, stop, step)]
 
             except Z3Exception:
                 raise IndexError("Invalid Memory Slice")
--- a/mythril/laser/ethereum/state/memory.py
+++ b/mythril/laser/ethereum/state/memory.py
@@ -45,7 +45,7 @@ class Memory:
                 )
             )
 
-    def write_word_at(self, index: int, value: Union[int, BitVecRef, BoolRef]):
+    def write_word_at(self, index: int, value: Union[int, BitVecRef, bool, BoolRef]):
         """
         Writes a 32 byte word to memory at the specified index`
         :param index: index to write to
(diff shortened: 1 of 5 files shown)


mythril/laser/ethereum/state/memory.py
```diff
@@ -0,0 +1,110 @@
...
+                if start is None:
+                    start = 0
+                if stop is None:  # 2**256 is just a bit too big
+                    raise IndexError("Invalid Memory Slice")
+                if step is None:
+                    step = 1
+                dataparts = []
+
+                for i in range(start, stop, step):
+                    dataparts.append(self[i])
```
norhh: we can directly use a list comprehensive here.
`dataparts = [self[i] for i in range(start, stop, step)]`

mythril/laser/ethereum/state/memory.py
```diff
@@ -0,0 +1,108 @@
...
+
+    def write_word_at(self, index: int, value: Union[int, BitVecRef, BoolRef]):
+        """
+        Writes a 32 byte word to memory at the specified index`
+        :param index: index to write to
+        :param value: the value to write to memory
+        """
+        try:
+            # Attempt to concretize value
+            _bytes = util.concrete_int_to_bytes(value)
```
norhh: Unfortunately value is an element taken from stack, it can be either `ExprRef, int, bytes, or boolean`. If it is boolean this might fails(it can be also be boolean as we can see it in the is_zero opcode) .One way might be is to handle boolean in here, 
another way is to reject boolean types in the stack and also adding an assert there in `MachineStack::push()` to fix to `Union[int, BitVecRef, BoolRef]` .

nbanmp: I'll just add support for bool here for now. I looked into requiring that only BitVecRefs are in the stack, and it's not trivial, so when/if it happens, it should probably be in a different PR.

norhh:
@nbanmp , whenever memory is extended, the gas should be consumed.

nbanmp:
After the work I've done on this, I don't think we should actually make the switch to a z3 array, and instead we should just use a list wrapped in a class that handles the important logic for stored symbolic values. It is probably a cleaner and faster solution to #685, as long as we don't need symbolic indexes.

As such, I'm going to close this for now and open a new pr later this week with a better implementation.

PR #707, opened 2018-11-16
Add 4byte signatures and small DB optimizations

dmuhs:
This PR will
- add a ton more signatures from 4byte.directory to Mythril
- adapt the report tests to reflect the actual signature names
- make the signature DB a singleton, which loads the file only once to avoid long waits
- ... which in result speeds up all tests that deal with signature resolution

As our time should go into other issues right now, I'd also hear your opinions on whether we should introduce a lightweight DB like sqlite3 instead of a plain JSON file for the signatures. That would save us the hassle of dealing with concurrent access and memory overhead for larger signature files.


mythril/analysis/modules/external_calls.py
```diff
@@ -85,7 +85,7 @@ def execute(statespace):
                     if "calldata" in target:
                         description += "an address provided as a function argument. "
                     else:
-                        description += "the address of the transaction sender. "
+                        description += "the address of the transaction senderself. "
```
JoranHonig: I think this is an accidental self.

dmuhs: Fixed. What a *self-centered* mistake. \*badum-tss\* :drum:

mythril/mythril.py
```diff
@@ -94,11 +94,12 @@ def __init__(
 
         self.mythril_dir = self._init_mythril_dir()
 
-        self.sigs = signatures.SignatureDb(
-            enable_online_lookup=self.enable_online_lookup
-        )
+        self.sigs = {}
```
norhh: If we initialize it here then line 105/106 might throw an error right, `self.sigs.signatures_file`

dmuhs: Good catch! In the old and new `SignatureDB` constructor `signatures_file` is set to None, though, so I think we can remove that part of the logging statement completely, because it won't show a path anyway

dmuhs: Whoops, overlooked that `open` is called in the constructor :smile:

dmuhs:
Fixed @norhh's comment on the exception handling by delegating the generation of exception texts to the `SignatureDB` instance, which holds the file path. Rebased with develop and ready to review again :+1:

PR #709, opened 2018-11-16
Detection Module Refactoring

dmuhs:
This PR will:
- introduce a base class for detection modules
- refactor the existing modules to use that class
- add a helper method for aggregating (and filtering) existing analysis modules
- add an entrypoint for modules (either `post` or `callback`) to determine when they are executed
- only execute modules at the specified entrypoint

This is in preparation for the callback refactoring in which now only the detection module logic has to be adapted to use global states and the relevant entrypoints have to be changed to `callback`


mythril/analysis/security.py
```diff
@@ -4,19 +4,29 @@
 import logging
 
 
-def fire_lasers(statespace, module_names=None):
-
-    issues = []
+def get_detection_modules(entrypoint, except_modules=None):
+    except_modules = [] if except_modules is None else except_modules
+    except_modules.append("base")  # always exclude base class file
```
JoranHonig:
```suggestion
    except_modules = ["base"].extend(except_modules)
```

JoranHonig: and make except modules have the default ```()```

mythril/laser/ethereum/svm.py
```diff
@@ -70,13 +76,33 @@ def __init__(
...
+        self.pre_hooks = defaultdict(list)
+        self.post_hooks = defaultdict(list)
+
+        self.register_detection_modules()
 
         logging.info(
             "LASER EVM initialized with dynamic loader: " + str(dynamic_loader)
         )
 
+    def register_detection_modules(self):
```
JoranHonig: I would like to keep the logic that doesn't have to do with symbolic execition but just reasons about the execution, outside of laser. WDYT

JoranHonig: So what I'd like to happen is that the equivalent class of the sym exec wrapper, registers the hooks with the laser evm instance. But that the laser does not get dependencies towards mythril.analysis

mythril/analysis/security.py
dmuhs: `list.extend` returns `None` and changes the list in place. So that wouldn't have any effect. But if you want to keep the default argument an iterable, we could do something like
```python
def get_detection_modules(entrypoint, except_modules=()):
    except_modules = list(except_modules) + ["base"]
```
That way we always have a list with at least `base` in it to call all familiar methods on, while the default argument stays immutable (which is *really* important. :smile: )

mythril/laser/ethereum/svm.py
dmuhs: I'll refactor it into a utility method :+1:

dmuhs: To elaborate a little more: I think it's reasonable to keep the modules that reason about execution outside of Laser. That would be a semantic separation, which is one way of looking at it (and the better one for us regarding potential future refactorings). The other one would be the view that the detection modules change the state of Laser and (to satisfy the isolation property) should be included.

I'll add the `register_post_hook` and `register_pre_hook` methods to Laser that will allow you to register any of them. That way we have the same functionality without the analysis module dependencies.

PR #713, opened 2018-11-17
Introduce additional constraints in Ether Thief module

b-mueller:
An "ETH relay" contract is often found on the mainnet. This contract contains a function that sends its balance to a fixed address ([example](https://etherscan.io/address/0x94e934139ae32efa180beaecc7e92dbe141de9bd#code)). This PR narrows the win condition by adding the following constraints:

- The receiver address must be is identical to the sender address;
- The sender address can be chosen arbitrarily.

### DIFF
--- a/mythril/analysis/modules/ether_thief.py
+++ b/mythril/analysis/modules/ether_thief.py
@@ -36,7 +36,7 @@ class EtherThief(DetectionModule):
         )
 
     def execute(self, state_space):
-        logging.debug("Executing module: " + format(self.name))
+        logging.debug("Executing module: %s", self.name)
         issues = []
 
         for k in state_space.nodes:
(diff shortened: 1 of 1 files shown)


mythril/analysis/modules/ether_thief.py
```diff
@@ -8,37 +8,35 @@
...
-                "- The transaction sender does not match contract creator\n"
-                "- The sender has not previously sent any ETH to the contract account.\n\n"
-                "This is somewhat coarse and needs to be refined in the future."
-            ),
+            description=DESCRIPTION,
         )
 
     def execute(self, state_space):
-        logging.debug("Executing module: ETHER_THIEF")
+        logging.debug("Executing module: " + format(self.name))
```
dmuhs: +1 on dynamically adding it in. The `logging` factory has a neat twist where it resolves the name lazily - right before writing out. It's best practice to pass in the variable name without inserting it as such
```suggestion
        logging.debug("Executing module: %s", self.name)
```
Sure, it's not a format string, but that just shows how desperately we need a refactoring of the STL. :slightly_smiling_face:

Issue #714, opened 2018-11-17
"-m / --modules" argument is ignored


b-mueller:
## Description

Mythril always runs all analysis modules even if specific modules are provided via `-m`.

## How to Reproduce

```
 $MYTH -m ether_send -x wallet_01_ok.sol
==== Integer Overflow ====
SWC ID: 101
Type: Warning
Contract: Wallet
Function name: deposit()
PC address: 1000
Estimated Gas Usage: 1152 - 1627
This binary add operation can result in integer overflow.

--------------------
In file: wallet_01_ok.sol:17

balances[msg.sender] + msg.value

--------------------

==== Exception state ====
SWC ID: 110
Type: Informational
Contract: Wallet
Function name: deposit()
PC address: 1008
Estimated Gas Usage: 1177 - 1652
A reachable exception (opcode 0xfe) has been detected. This can be caused by type errors, division by zero, out-of-bounds array access, or assert violations. Note that explicit `assert()` should only be used to check invariants. Use `require()` for regular input checking. 
--------------------
In file: wallet_01_ok.sol:17

assert(balances[msg.sender] + msg.value > balances[msg.sender])

--------------------
```

## Expected behavior

If the user provides a list of modules, Mythril should run the selected modules only.

## Environment

- Mythril version: develop
- Solidity compiler and version: 0.4.24+commit.e67f0147.Darwin.appleclang
- Python version: 3.6.4
- OS and Version: Mac OS Mojave

b-mueller:
@dmuhs just realized that the "modules" parameter is now used as a blacklist of modules *not* to run. Why is that?

PR #716, opened 2018-11-18
Revert module loading behavour

b-mueller:
Change `get_detection_modules(entrypoint)` to treat modules argument as a list of modules to be loaded to fix behavior of the `-m` argument. `ModuleNotFound` exception is now thrown if a non-existent module name is passed.

@dmuhs I'm not sure why you implemented the `exclude_modules` arg before but I didn't see it used anywhere. Should we re-add this?


dmuhs:
WRT #714 this was probably due to the old code in `security.py`:
```python
         if not module_names or module.__name__ in module_names:
             logging.info("Executing " + str(module))
             issues += module.execute(statespace)
```
which I probably misread as `not (module_names or module.__name__ in module_names)` making `module_names` a blacklist.

mythril/analysis/security.py
```diff
@@ -28,16 +29,22 @@ def get_detection_module_hooks():
...
+
     _modules = []
 
-    for loader, name, _ in pkgutil.walk_packages(modules.__path__):
-        module = loader.find_module(name).load_module(name)
-        if (
-            module.__name__ not in except_modules
-            and module.detector.entrypoint == entrypoint
-        ):
+    if not len(include_modules):
```
dmuhs: I think the `len()` is redundant here, because `bool(list()) => False`

PR #719, opened 2018-11-18
Make transaction count fixed

b-mueller:
Transaction count is now a fixed parameter.

I also took the `-t` flag away from Truffle and assigned it to `--transaction-count` instead:

```
$ myth -t3 -v1 -x source.sol 
$ myth --transaction-count 3 -v1 -x source.sol 
```


tests/mythril_dir/config.ini
```diff
@@ -0,0 +1,4 @@
+[defaults]
```
JoranHonig: This shouldn't be here right :wink:

mythril/interfaces/cli.py
```diff
@@ -389,6 +389,7 @@ def main():
                     max_depth=args.max_depth,
                     execution_timeout=args.execution_timeout,
                     create_timeout=args.create_timeout,
+                    transaction_count=1,
```
JoranHonig: Why is it 1 here by default?

JoranHonig: Also I'd say that for the graph functionality it doesn't really make sense to do multi transactional analysis.
I'd think that symbolic storage is fine here.
What do you think?

mythril/laser/ethereum/svm.py
```diff
@@ -151,17 +147,12 @@ def _execute_transactions(self, address):
...
-        for i in range(self.max_transaction_count):
-            initial_coverage = self._get_covered_instructions()
+        for i in range(self.transaction_count):
 
             self.time = datetime.now()
             logging.info("Starting message call transaction, iteration: {}".format(i))
             execute_message_call(self, address)
 
-            end_coverage = self._get_covered_instructions()
-            if end_coverage == initial_coverage:
```
JoranHonig: We can keep the initial_coverage and end_coverage so we can log.info the increased instruction coverage. WDYT?

mythril/interfaces/cli.py
b-mueller:
```
Also I'd say that for the graph functionality it doesn't really make sense to do multi transactional analysis.
```

This answers your first question - that's why I set it to the fixed value 1 instead of args.transaction_count. Does that make sense?

tests/mythril_dir/config.ini
b-mueller: What, this appeared *again*? :(

PR #730, opened 2018-11-20
Refactor package and module structure

b-mueller:
Refactor packages and modules:

- Merge `ether` package into `ethereum`
- Create a new `solidity` package for everything related to Solidity support
- Rename `ETHContract` to `EVMContract`
- Remove unused `EVM` class

Besides cleaning things up a bit, this is also in anticipation of adding support for new interpreters (`EWASMContract) and languages (Vyper), even though that will need some more prep work.


b-mueller:
> Rename solidity module to contract_source

I was thinking along these lines as well. How about `lang`?

`
from mythril.ethereum.lang.solidity import SolidityContract
`

b-mueller:
Other suggestions sound all good.

JoranHonig:
`lang` is awesome, I wasn't happy with `contract_source` myself

PR #752, opened 2018-11-24
Improve Suicide and Ether Thief modules

b-mueller:
Rewrite the SUICIDE module along the lines of the Ether thief module. Now uses the solver to determine whether the 'to' address can be set.

Also removes redundant use of solver in Ether Thief.


JoranHonig:
Merging this with #743 might be a bit hard on the suicide module :sweat_smile:

Issue #744, opened 2018-11-23
Refactor test samples so they compile with solc 0.5.0


b-mueller:
## Description

[CircleCI tests are failing](https://circleci.com/gh/ConsenSys/mythril-classic/1750) since the latest major release of solc because some  input contracts fail to compile.

Refactor the samples for compatibility with solc 0.5.0.

JoranHonig:
@dmuhs This can be closed right?

PR #754, opened 2018-11-25
Fix test contracts for solc 0.5.0

dmuhs:
This PR will fix the compiler errors for the latest version of solc 0.5.0 (#744). As a side effect, all vulnerabilities we're trying to test for are still there - the code changes make their presence just a little more explicit and occasionally warn about it. No breaking changes for our test scenarios, though.

Edit: It now also contains fresher solidity examples that all compile with 0.5.0.

### DIFF
--- a/mythril/analysis/analysis_utils.py
+++ b/mythril/analysis/analysis_utils.py
@@ -17,6 +17,7 @@ def get_non_creator_constraints(state: GlobalState) -> (List, bool):
         state.world_state.transaction_sequence[0], ContractCreationTransaction
     ):
         creator = state.world_state.transaction_sequence[0].caller
+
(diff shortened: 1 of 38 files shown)


norhh:
Looks like now the tests should be ported to running with default solc which is `0.5.0` over `0.4.24`(which was being downloaded for tests as a quick fix)  :sweat_smile: thats the reason for the failing tests.

dmuhs:
@norhh The current version of `py-solc` doesn't support 0.5.0 yet. So I changed the installation routine a little to also take `/usr/bin/solc` (the default installation path if you are using the Ubuntu PPA) into account if it exists. All green now. :)

PR #755, opened 2018-11-25
Bernd 3

b-mueller:
Merging changes to suicide and ether_send modules without the refactor to concrete callers.


tests/mythril_dir/config.ini
```diff
@@ -0,0 +1,4 @@
+[defaults]
+leveldb_dir = /Users/bernhardmueller/Library/Ethereum/geth/chaindata
```
cleanunicorn: This probably needs to be removed from the repo

PR #756, opened 2018-11-25
Add optional address argument to load_from_bytecode

b-mueller:


tests/mythril_dir/config.ini
```diff
@@ -0,0 +1,4 @@
+[defaults]
+leveldb_dir = /Users/bernhardmueller/Library/Ethereum/geth/chaindata
```
cleanunicorn: This should be removed from repo

PR #757, opened 2018-11-26
Make analysis more awesome than it already is

b-mueller:
This PR introduces a number of changes significantly reduce false positives in source code / on-chain analysis. The only source of FPs left so far appears to be unresolved external calls (which is to be expected).

Note this break some old modules such as "external calls", specifically if they do regex comparisons on constraints.

**Changelog:**

- Check for add overflows in Ether Thief (we don't care about attacks that require more 2^^256 ETH to be sent
- Use UGT comparison in Ether Thief & and fix a bug in summing call values
- Reduce complexity of analysis modules and eliminates the analysis_utils class
- Return `BitVecVal(0x0)` when initialized storage is accessed and access mode is set to "concrete"
- Set concrete addresses for creator and attacker
- Eliminate false positives resulting from mappings that use the sender address as an index


tests/testdata/outputs_expected/calls.sol.o.text
```diff
@@ -18,26 +8,6 @@ Estimated Gas Usage: 1352 - 35963
 The return value of an external call is not checked. Note that execution continue even if the called contract throws.
 --------------------
 
-==== Message call to external contract ====
```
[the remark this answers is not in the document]
b-mueller: Not 100% sure but I think it's because the call module only reports calls to symbolic addresses. That module needs to be refactored anyway though.

dmuhs:
Can we make the JSON test reports pretty again so we can use line numbers in diffs?

b-mueller:
> Can we make the JSON test reports pretty again so we can use line numbers in diffs?

@dmuhs didn't the last commit fix that?

b-mueller:
Also, what tool do you use to prettify JSON? Using `black` breaks the tests for me.

tests/testdata/inputs/metacoin.sol.o
```diff
@@ -1 +0,0 @@
-60606040526004361061004c576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff16806327e235e314610051578063412664ae1461009e575b600080fd5b341561005c57600080fd5b610088600480803573ffffffffffffffffffffffffffffffffffffffff169060200190919050506100f8565b6040518082815260200191505060405180910390f35b34156100a957600080fd5b6100de600480803573ffffffffffffffffffffffffffffffffffffffff16906020019091908035906020019091905050610110565b604051808215151515815260200191505060405180910390f35b60006020528060005260406000206000915090505481565b6000816000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002054101561016157600090506101fe565b816000803373ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060008282540392505081905550816000808573ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff16815260200190815260200160002060008282540192505081905550600090505b929150505600a165627a7a72305820c860d60246e215343f02c5025aeef4ad1f207b0a7d2dec05e43f6ecaaebe9cec0029
```
JoranHonig: why did this test get removed?

tests/testdata/outputs_expected/metacoin.sol.o.json
```diff
@@ -1,5 +1,18 @@
 {
   "error": null,
-  "issues": [],
+  "issues": [
```
JoranHonig: This shouldn't be possible

tests/testdata/outputs_expected/overflow.sol.o.markdown
```diff
@@ -23,15 +23,3 @@ The subtraction can result in an integer underflow.
 ### Description
 
 The subtraction can result in an integer underflow.
-
```
JoranHonig: This is not correct right. It is still possible to have an overflow here

tests/testdata/outputs_expected/returnvalue.sol.o.markdown
```diff
@@ -1,29 +1,5 @@
...
-- Contract: Unknown
-- Function name: `callchecked()`
-- PC address: 196
-- Estimated Gas Usage: 599 - 1210
-
-### Description
-
-This contract executes a message call to to another contract. Make sure that the called contract is trusted and does not execute user-supplied code.
-
-## Message call to external contract
```
JoranHonig: these should still be detected right?

tests/testdata/outputs_expected/overflow.sol.o.markdown
b-mueller: It's really weird. I'm getting correct results on source code analysis but the tests generate incomplete results.

```
$ $MYTH -x tests/testdata/input_contracts/overflow.sol 
==== Integer Underflow ====
SWC ID: 101
Type: Warning
Contract: Over
Function name: sendeth(address,uint256)
PC address: 497
Estimated Gas Usage: 784 - 1069
The subtraction can result in an integer underflow.

--------------------
In file: tests/testdata/input_contracts/overflow.sol:14

balances[msg.sender] - _value

--------------------

==== Integer Underflow ====
SWC ID: 101
Type: Warning
Contract: Over
Function name: sendeth(address,uint256)
PC address: 579
Estimated Gas Usage: 1317 - 1792
The subtraction can result in an integer underflow.

--------------------
In file: tests/testdata/input_contracts/overflow.sol:15

balances[msg.sender] -= _value

--------------------

==== Integer Overflow ====
SWC ID: 101
Type: Warning
Contract: Over
Function name: sendeth(address,uint256)
PC address: 655
Estimated Gas Usage: 6840 - 27505
This binary add operation can result in integer overflow.

--------------------
In file: tests/testdata/input_contracts/overflow.sol:16

balances[_to] += _value

--------------------
```

Same for calls.sol:

```
$ $MYTH -mexternal_calls -x tests/testdata/input_contracts/calls.sol 
==== Message call to external contract ====
SWC ID: 107
Type: Informational
Contract: Caller
Function name: thisisfine()
PC address: 725
Estimated Gas Usage: 662 - 1273
This contract executes a message call to to another contract. Make sure that the called contract is trusted and does not execute user-supplied code.
--------------------
In file: tests/testdata/input_contracts/calls.sol:16

fixed_address.call("")

--------------------

==== Message call to external contract ====
SWC ID: 107
Type: Warning
Contract: Caller
Function name: callstoredaddress()
PC address: 900
Estimated Gas Usage: 706 - 1317
This contract executes a message call to an address provided as a function argument. Generally, it is not recommended to call user-supplied addresses using Solidity's call() construct. Note that attackers might leverage reentrancy attacks to exploit race conditions or manipulate this contract's state.
--------------------
In file: tests/testdata/input_contracts/calls.sol:29

stored_address.call("")

--------------------

==== Message call to external contract ====
SWC ID: 107
Type: Informational
Contract: Caller
Function name: reentrancy()
PC address: 1036
Estimated Gas Usage: 728 - 1339
This contract executes a message call to to another contract. Make sure that the called contract is trusted and does not execute user-supplied code.
--------------------
In file: tests/testdata/input_contracts/calls.sol:20

fixed_address.call("")

--------------------

==== State change after external call ====
SWC ID: 107
Type: Warning
Contract: Caller
Function name: reentrancy()
PC address: 1098
Estimated Gas Usage: 728 - 1339
The contract account state is changed after an external call. Consider that the called contract could re-enter the function before this state change takes place. This can lead to business logic vulnerabilities.
--------------------
In file: tests/testdata/input_contracts/calls.sol:21

statevar = 0

--------------------

==== Message call to external contract ====
SWC ID: 107
Type: Warning
Contract: Caller
Function name: calluseraddress(address)
PC address: 1147
Estimated Gas Usage: 402 - 683
This contract executes a message call to an address provided as a function argument. Generally, it is not recommended to call user-supplied addresses using Solidity's call() construct. Note that attackers might leverage reentrancy attacks to exploit race conditions or manipulate this contract's state.
--------------------
In file: tests/testdata/input_contracts/calls.sol:25

addr.call("")

--------------------
```

But some of this is missing in outputs_current.

Unfortunately I don't understand well how the tests are set up :( Maybe @norhh can help fix this?

JoranHonig:
@b-mueller We already have a ToD module that kind of currently depends on having multiple callers

tests/testdata/inputs/metacoin.sol.o
b-mueller: Actually I'm not sure why I removed it, I guess I was just annoyed by it giving me errors :)

tests/testdata/outputs_expected/returnvalue.sol.o.markdown
b-mueller: ^^ see above.

b-mueller:
So when trying to address the call issues I ended up completely rewriting the external calls module and also refactoring it to callback mode (also the Ether Thief module).

It partially works, but there's still a bug and I can't figure out. I'm not 100% sure if I'm doing it wrong or if there's a fundamental issue somewhere else.

The callback module detects stuff sometimes, but it often doesn't work for very simple examples. For example, it returns nothing for this code (and informational issue should be returned here):

```solidity
pragma solidity 0.5.0;

contract Caller {

	function simplecall() public {
		address fixed_address = address(0x552254CbAaF32613C6c0450CF19524594eF84044);
	    fixed_address.call.value(0)('');
	}

}
```

The reason is that Z3 can't find a model for the "CALL" state even with no extra constraints added. Checking the constraints on that node also shows a large amount of constraints on the node.

constraints

@JoranHonig @norhh any ideas?

b-mueller:
I tried merging #761, didn't improve it 😭

b-mueller:
This makes *no sense* whatsoever and it's killing me.

If I remove the gas constraint, then the `transfer` is detected but not the low-level call:

```solidity
pragma solidity 0.4.24;

contract Caller {

    function simplecall() public {
        address fixed_address = address(0x552254CbAaF32613C6c0450CF19524594eF84044);

        // Not detected:

        fixed_address.call(bytes4(sha3("func(address,uint256)")));

        // Not detected:

        Callee callee = new Callee();
        callee.callme();

    }
}
```

```bash
$ $MYTH -mexternal_calls -x ~/Desktop/call.sol  
==== External call ====
SWC ID: 107
Type: Informational
Contract: Caller
Function name: simplecall()
PC address: 189
Estimated Gas Usage: 607 - 888
The contract executes a function call to an external address. Verify that the code at this address is trusted and immutable.
--------------------
In file: /Users/bernhardmueller/Desktop/call.sol:13

msg.sender.transfer(address(this).balance)

--------------------
```

The node with the low-level call contains two constraints that apparently simplify to `false`:

```
UGT(0, 0 + 0 + call_value4)  ->   False

1271270613000041655817448348132275889066893754095 ==
1461501637330902918203684832716283019655932542975 &
486029827593346413276326003332628986382196949060    -> False
```

Where do those come from?

mythril/analysis/modules/ether_thief.py
```diff
@@ -22,7 +22,57 @@
...
+
+    if instruction["opcode"] != "CALL":
+        return []
+
+    call_value = state.mstate.stack[-3]
+    target = state.mstate.stack[-2]
+
+    eth_sent_total = BitVecVal(0, 256)
+
+    constraints = node.constraints
```
norhh: This should Have been copied, after some painful debugging realized that problem was in this file :rofl:

mythril/analysis/modules/ether_thief.py
```diff
@@ -22,7 +22,57 @@
...
+
+    for tx in state.world_state.transaction_sequence:
+        if tx.caller == 0xDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF:
+
+            # There's sometimes no overflow check on balances added.
+            # But we don't care about attacks that require more 2^^256 ETH to be sent.
+
+            constraints += [BVAddNoOverflow(eth_sent_total, tx.call_value, False)]
```
norhh: This is where that unwanted constraints were added

b-mueller:
FYI, I removed the old delegatecall unit tests in `analysis` since the delegatecall module is going to be refactored heavily (or might be merged into the external_calls module) and I don't want to be continuously updating this. Also, delegate call variants will be included in the SWC so we have duplicate effort.

Issue #556, opened 2018-10-09
Refactor: Remove unused Calldatatype


JoranHonig:
## Description
It seems like the calldatatype variable is not really used anymore:
```python 
class Environment:
    """
    The environment class represents the current execution environment for the symbolic executor
    """

    def __init__(
        self,
        active_account: Account,
        sender: ExprRef,
        calldata: Calldata,
        gasprice: ExprRef,
        callvalue: ExprRef,
        origin: ExprRef,
        code=None,
        calldata_type=CalldataType.SYMBOLIC,
    ):
        # Metadata

        self.active_account = active_account
        self.active_function_name = ""

        self.address = BitVecVal(int(active_account.address, 16), 256)

        # Ib
        self.code = active_account.code if code is None else code

        self.sender = sender
        self.calldata = calldata
        self.calldata_type = calldata_type
        self.gasprice = gasprice
        self.origin = origin
        self.callvalue = callvalue

    def __str__(self) -> str:
        return str(self.as_dict)

    @property
    def as_dict(self) -> Dict:
        return dict(
            active_account=self.active_account,
            sender=self.sender,
            calldata=self.calldata,
            gasprice=self.gasprice,
            callvalue=self.callvalue,
            origin=self.origin,
            calldata_type=self.calldata_type,
        )

```

### Requirements
- Verify that calldatatype is indeed unused
- Removal of calldata_type variable
- Removal of CalldataType enum

PR #761, opened 2018-11-28
Refactor of calldata

### LINKED ISSUE #759: Refactor: Calldata without arrays
## Description 
Currently our Calldata implementation uses arrays to facilitate symbolic reads of the calldata. However with the move to an abstraction layer vs directly using z3 we'll be losing that particular functionality.

This issue tracks the move from the current implementation, to one independent of arrays.

### Requirements
- Calldata doesn't use z3 arrays
- Calldata still supports symbolic reads

JoranHonig:
TODO:
- [x] Documentation strings
- [x] Moar tests
closes #759

### DIFF
--- a/mythril/laser/ethereum/state/calldata.py
+++ b/mythril/laser/ethereum/state/calldata.py
@@ -34,7 +34,7 @@ class BaseCalldata:
         return self[offset: offset + 32]
 
     def __getitem__(self, item: Union[int, slice]) -> Any:
-        if isinstance(item, int):
+        if isinstance(item, int) or isinstance(item, ExprRef):
             return self._load(item)
 
         if isinstance(item, slice):
--- a/mythril/laser/ethereum/state/calldata.py
+++ b/mythril/laser/ethereum/state/calldata.py
@@ -1,14 +1,6 @@
 from enum import Enum
 from typing import Union, Any
-from z3 import (
-    BitVecVal,
-    BitVecRef,
-    BitVec,
-    simplify,
-    Concat,
-    If,
-    ExprRef
-)
+from z3 import BitVecVal, BitVecRef, BitVec, simplify, Concat, If, ExprRef
 from z3.z3types import Z3Exception
 
 from mythril.laser.ethereum.util import get_concrete_int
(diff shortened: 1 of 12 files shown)


mythril/laser/ethereum/state/calldata.py
```diff
@@ -21,84 +11,133 @@ class CalldataType(Enum):
     SYMBOLIC = 2
```
nbanmp: Can we remove `CalldataType`? I believe the only place it is used is line 1270 of instructions.py, and that whole if statement is always false and should be deleted...

mythril/laser/ethereum/state/calldata.py
```diff
@@ -21,84 +11,133 @@ class CalldataType(Enum):
...
                     current_index = simplify(current_index + step)
             except Z3Exception:
                 raise IndexError("Invalid Calldata Slice")
 
-            values, constraints = zip(*dataparts)
-            result_constraints = []
-            for c in constraints:
-                result_constraints.extend(c)
-            return simplify(Concat(values)), result_constraints
+            return simplify(Concat(parts))
```
nbanmp: I think it makes more sense and is more consistent with Python to return a list when `Calldata` is accessed as a slice rather than concatenating everything together. The concatenation should instead happen in `get_word_at`.

wdyt?

tests/testdata/input_contracts/overflow.sol
```diff
@@ -11,7 +11,7 @@ contract Over {
   }
 
   function sendeth(address _to, uint _value) public returns (bool) {
-    require(balances[msg.sender] - _value >= 0);
+    // require(balances[msg.sender] - _value >= 0);
```
nbanmp: What's this?

I suppose this require is a tautology, but you could go all out and delete it if you wanted. But isn't this line included to point out the logical error that a developer might make?

mythril/laser/ethereum/state/calldata.py
JoranHonig: See #556 ^^

tests/testdata/input_contracts/overflow.sol
JoranHonig: > What's this?
 Unintended :sweat_smile:

PR #762, opened 2018-11-28
Initial StateAnnotation Implementation

JoranHonig:
This is an initial implementation of the StateAnnotation, a class which will be used by callback analysis modules to reason about traces


mythril/laser/ethereum/state/world_state.py
```diff
@@ -2,20 +2,22 @@
...
+from mythril.laser.ethereum.state.annotation import StateAnnotation
 
 
 class WorldState:
     """
     The WorldState class represents the world state as described in the yellow paper
     """
 
-    def __init__(self, transaction_sequence=None):
+    def __init__(self, transaction_sequence=None, annotations=None):
```
dmuhs:
```suggestion
    def __init__(self, transaction_sequence=None, annotations: List[StateAnnotation] = None) -> None:
```

mythril/laser/ethereum/state/world_state.py
```diff
@@ -67,6 +73,13 @@ def create_initialized_contract_account(self, contract_code, storage) -> None:
         new_account.storage = storage
         self._put_account(new_account)
 
+    def annotate(self, annotation: StateAnnotation):
```
dmuhs:
```suggestion
    def annotate(self, annotation: StateAnnotation) -> None:
```

mythril/laser/ethereum/state/world_state.py
```diff
@@ -67,6 +73,13 @@ def create_initialized_contract_account(self, contract_code, storage) -> None:
...
+        self._annotations.append(annotation)
+
+    @property
+    def annotations(self):
```
dmuhs:
```suggestion
    def annotations(self) -> List[StateAnnotation]:
```

dmuhs:
Great work! :fire: I have proposed some more type hints. I've also noticed that we are also missing some hints for default parameter values. We can address that in a separate PR, though and align it with the `mypy` integration.

PR #764, opened 2018-11-28
Memory refactor to store symbolic values (#685)

### LINKED ISSUE #685: Refactor: Memory to z3 array

nbanmp:
- Refactors memory into a new class
- Store symbolic values in memory
- Does not use a z3 Array. Better for claripy and easier and more logical to implement
- Keeps mem_extend (needed for gas anyway, and also needed still to resize list)

fixes #685


tests/testdata/outputs_expected/kinds_of_calls.sol.o.json
```diff
@@ -1,54 +1 @@
...
... of an external call is not checked. Note that execution continue even if the called contract throws.", "function": "_function_0x9b58bc26", "max_gas_used": 35913, "min_gas_used": 1161, "swc-id": "104", "title": "Unchecked CALL return value", "type": "Informational"}, {"address": 1038, "contract": "Unknown", "debug": "<DEBUG-DATA>", "description": "This contract executes a message call to an address provided as a function argument. Generally, it is not recommended to call user-supplied addresses using Solidity's call() construct. Note that attackers might leverage reentrancy attacks to exploit race conditions or manipulate this contract's state.", "function": "_function_0xeea4c864", "max_gas_used": 1223, "min_gas_used": 471, "swc-id": "107", "title": "Message call to external contract", "type": "Warning"}, {"address": 1046, "contract": "Unknown", "debug": "<DEBUG-DATA>", "description": "The return value of an external call is not checked. Note that execution continue even if the called contract throws.", "function": "_function_0xeea4c864", "max_gas_used": 35938, "min_gas_used": 1186, "swc-id": "104", "title": "Unchecked CALL return value", "type": "Informational"}], "success": true}
```
dmuhs: Can we make this pretty again for readability? :slightly_smiling_face:

nbanmp: https://github.com/ConsenSys/mythril-classic/pull/764/files#diff-1bd4444f8e8f9884576dd475243c55f2 😉

dmuhs: Just noted. Pardon my stale notifications. :laughing:

nbanmp:
Interesting. The svm test case timed out in python 3.6, and not for python 3.5. Running it locally, I completed the svm test case in 4:12 with the current develop, and 6:18 with my pr merged. My guess is that on the less powerful circleci machine, execution time is borderline with the timeout.

nbanmp:
The time tests took to run:

Before refactors: 13:31 <https://circleci.com/gh/ConsenSys/mythril-classic/1840>
After calldata refactor: 24:58 <https://circleci.com/gh/ConsenSys/mythril-classic/1858>
After memory refactor: 17:32 <https://circleci.com/gh/ConsenSys/mythril-classic/1890>
Both calldata and memory refactors: >26:10 (Timed out during python 3.6 svm test case and did not complete) <https://circleci.com/gh/ConsenSys/mythril-classic/1901>

Any suggestions?

cc @JoranHonig because you worked on calldata.

mythril/laser/ethereum/state/memory.py
```diff
@@ -0,0 +1,99 @@
...
+        """
+        try:
+            # Attempt to concretize value
+            if isinstance(value, bool):
+                _bytes = (
+                    int(1).to_bytes(32, byteorder="big")
+                    if value
+                    else int(0).to_bytes(32, byteorder="big")
+                )
+            _bytes = util.concrete_int_to_bytes(value)
```
norhh: Shouldn't this be in an `else`? As `value=True/False` might raise an AttributeError here.

.circleci/config.yml
```diff
@@ -48,6 +48,7 @@ jobs:
           name: Unit-testing
           command: tox
           working_directory: /home/mythril
+          no_output_timeout: 15m
```
norhh: We can remove this now right?

mythril/laser/ethereum/state/memory.py
```diff
@@ -0,0 +1,102 @@
...
+        :param index: integer representing the index to access
+        :return: 32 byte word at the specified index
+        """
+        try:
+            return util.concrete_int_from_bytes(
+                bytes([util.get_concrete_int(b) for b in self[index : index + 32]]), 0
+            )
+        except TypeError:
+            return simplify(
+                Concat(
```
norhh: I recently realized that `Concat()` increases the bitwidth and operations like UDiv need same bitwidth of both lhs and rhs, so can you change the bitwidth in here?

nbanmp: :sweat_smile: I had a 256 instead of an 8 here. It should be good now, and if not, the assert will catch it.

Issue #771, opened 2018-11-29
Latest Docker Image No Longer Respects --solv


mickayz:
## Description

Latest Docker image running Mythril version v0.19.8 does not respect "--solv"

## How to Reproduce

Run the following:

```console
docker run --rm -v $(pwd):/tmp mythril/myth -x /tmp/DonationScan.sol -mether_thief --verbose-report --solv 0.4.24
ERROR:root:Solc has experienced a fatal error (code 1).

/tmp/DonationScan.sol:1:1: Error: Source file requires different compiler version (current compiler is 0.5.0+commit.1d4f565a.Linux.g++ - note that nightly builds are considered to be strictly less than the released version
pragma solidity ^0.4.24;
^----------------------^
```

## Environment

Latest Docker image - Mythril version v0.19.8

mickayz:
I think this is the offending commit:

https://github.com/ConsenSys/mythril-classic/commit/c4127fd229c40e145f4d2826571af81bede09d05#diff-45953b4ee9eebfcb91aedc01cb804a29

PR #773, opened 2018-11-29
Fix for --solv regression

mickayz:
Addresses Issue #771

Note that this fix removes the check for the Ubuntu PPA Solidity bin


norhh:
Hi @mickayz , can you ran install [black](https://pypi.org/project/black/)  and run `black .` in your directory and then make a commit, black is code checker and formatter.

norhh:
Hi @mickayz thanks for contributing.  can you ran install black and run black . in your directory and then add  a commit to this PR , black is a code formatter.

mythril/ethereum/util.py
```diff
@@ -79,7 +79,7 @@ def solc_exists(version):
             ".py-solc/solc-v" + version,
             "bin/solc",
         ),  # py-solc setup
-        "/usr/bin/solc",  # Ubuntu PPA setup
+        #"/usr/bin/solc",  # Ubuntu PPA setup
```
cleanunicorn: Should this be removed instead of commented out?
```suggestion

```

dmuhs:
Hey @mickayz. Sorry, I'm a little late to the party. First of all thanks for contributing! The added lookup path for `/usr/bin/solc` was added, because currently `py-solc` does not provide solc version 0.5.0. Removing that line will result in failing test cases, because of the outdated compiler (and other parts of Mythril now fixing the version to 0.5.0).
I have opened an issue regarding the new version [in the py-solc repo](https://github.com/ethereum/py-solc/issues/63). Until this is fixed, we will need the additional lookup.

mickayz:
Sorry, I'm limited on time and couldn't run tests of this code.  Was trying for a quick fix since this is blocking for my work.

@dmuhs, currently the changes recently made make it so that if that if /usr/bin/solc is present (as is the case on the Docker image), then solc_exists() will always return at least one value, regardless of the version argument that was passed.  This does not work if you are running Mythril against a v0.4.x contract. 

Is /usr/bin/solc always guaranteed to be a v0.5.x version only? If so, can there be a quick check inside solc_exists() that only adds that path if the version param is v0.5 or greater?

Additionally, the recent changes to mythril/Mythril.py makes it so that the result of solc.install_solc("v" + version) is not stored in the solc_binary, meaning that Mythril will still run against the v0.5 solc despite having just downloaded the v0.4 version.

Right now this is blocking using Mythril against any v0.4.x contracts if a v0.5.x version of solc is installed on the system, as is the case with the Docker image.

Additionally, I think its important that unit tests be included for not only "--solv v0.5" but also "--solv v0.4.25" so that this regression does not occur again.

PR #778, opened 2018-12-03
Add array based implementation of calldata

JoranHonig:
For speed


tests/laser/evm_testsuite/evm_test.py
```diff
@@ -76,6 +76,8 @@ def test_vmtest(
     # Arrange
     if test_name == "gasprice":
         return
+    if "calldatacopy0_return" not in test_name:
```
norhh: this will run only those tests which have `calldatacopy0_return` in them right?

JoranHonig: oops

mythril/laser/ethereum/state/calldata.py
```diff
@@ -49,12 +63,16 @@ def __getitem__(self, item: Union[int, slice]) -> Any:
                 )
                 parts = []
                 while simplify(current_index != stop):
-                    parts.append(self._load(current_index))
+                    element = simplify(self._load(current_index))
```
norhh: How about putting simplify() in the self._load() itself?

mythril/laser/ethereum/state/calldata.py
```diff
@@ -32,11 +45,12 @@ def calldatasize(self) -> ExprRef:
...
     def get_word_at(self, offset: int) -> ExprRef:
         """ Gets word at offset"""
-        return self[offset : offset + 32]
+        parts = self[offset : offset + 32]
+        return simplify(Concat(parts))
 
     def __getitem__(self, item: Union[int, slice]) -> Any:
         if isinstance(item, int) or isinstance(item, ExprRef):
-            return self._load(item)
+            return simplify(self._load(item))
```
norhh: Putting simplify in _load() might kill the simplify here too

mythril/laser/ethereum/state/calldata.py
JoranHonig: Why?

PR #785, opened 2018-12-04
SignatureDB behaviour changes and test architecture

### LINKED ISSUE #736: Incorrect reporting of function names
## Description

Function names are reported incorrectly for the [Ethernaut token smart contract](https://ethernaut.zeppelin.solutions/level/0x6545df87f57d21cb096a0bfcc53a70464d062512).

## How to Reproduce

```
$ myth -x Desktop/ethernaut-token.sol 
==== Integer Underflow ====
SWC ID: 101
Type: Warning
Contract: Token
Function name: **ambiguous** many_msg_babbage(bytes1)
PC address: 469
Estimated Gas Usage: 758 - 1043
The subtraction can result in an integer underflow.

--------------------
In file: Desktop/ethernaut-token.sol:13

balances[msg.sender] - _value

--------------------

==== Integer Underflow ====
SWC ID: 101
Type: Warning
Contract: Token
Function name: **ambiguous** many_msg_babbage(bytes1)
PC address: 551
Estimated Gas Usage: 1291 - 1766
The subtraction can result in an integer underflow.

--------------------
In file: Desktop/ethernaut-token.sol:14

balances[msg.sender] -= _value

--------------------

==== Integer Overflow ====
SWC ID: 101
Type: Warning
Contract: Token
Function name: **ambiguous** many_msg_babbage(bytes1)
PC address: 627
Estimated Gas Usage: 6814 - 27479
This binary add operation can result in integer overflow.

--------------------
In file: Desktop/ethernaut-token.sol:15

balances[_to] += _value

--------------------
```

## Environment

- Mythril version: develop
- Solidity compiler and version: 0.4.24+commit.e67f0147.Darwin.appleclang
- Python version: 3.6.4
- OS and Version: Mac OS Mojave

### LINKED ISSUE #765: Copy the signatures.db from mythril root to the .mythril directory
## Description
- Copy the signatures.db from mythril root to the .mythril directory if there is no `signatures.db` in `~/.mythril/`
- Make sure all the signatures used in the tests are present in the signatures.db

This might help to resolve various signature problems with tests(forever :v: ), and would also be helpful for the users.

dmuhs:
This PR will
- add a shortcut to the signature DB to avoid costly DB lookups when Solidity code was given and the function signatures are given (by adding a local cache)
- make the signature DB thread-safe to avoid race conditions in multithreaded tests
- change the output of ambiguous signatures to reflect all possible candidates (fixes #736)
- correctly initialise the signature DB file in the user directory with the default one from the repo (fixes #765)
- delete the now unused `mythril_dir` from the test cases (and consequentially make us handle one less DB file :slightly_smiling_face:)


norhh:
wait! the report tests run on multiple processes right?

norhh:
The report tests run on multiple processes right?

JoranHonig:
@norhh Yes. Would that break the lock?

dmuhs:
@JoranHonig Nope. That's exactly why I built in the lock. To share the Solidity file's signatures across multiple Disassembly objects (of which we instantiate quite a lot), the DB has to become a singleton with a local cache as shortcut to lookups. Access to the Singleton in the phase where it's first created is subject to a race condition and can result in the second thread overwriting the DB created by the first thread. Then you'll get funky stuff like invalid object references and missing tables. The sync lock on the singleton prevents that on creation.
Following threads will then just fetch the same DB object and concurrency is handled by the signature DB for reads and writes. :+1:

norhh:
The lock seems to for threading right? as it was imported from `threading` and the tests run on multiple processes so is this lock(the lock in the threading lib) extensible for processes too?

PR #787, opened 2018-12-05
Enable creation bytecode option

cryptobarbossa:
Enable runtime and creation bytecode options to be displayed with the appropriate header.


JoranHonig:
@cryptobarbossa Awesome thanks! But there is a case where both runtime and creation bytecode are available (if a solidity contract is provided). And I think that this pr doesn't really account for that,

Could you change the logic to something that checks whether the bytecode is available vs whether the  argument `args.code` is used?

cryptobarbossa:
@JoranHonig Thanks for highlighting that. Added.

mythril/interfaces/cli.py
```diff
@@ -364,10 +365,16 @@ def main():
...
-                0
-            ].get_easm()  # or mythril.disassemble(mythril.contracts[0])
-            sys.stdout.write(easm_text)
+            output = ""
+            easm_text = mythril.contracts[0].get_easm()
+            # or mythril.disassemble(mythril.contracts[0])
+
+            if easm_text:
+                output += "Runtime Disassembly: \n" + easm_text
+            if code:
```
JoranHonig:
```suggestion
            if mythril.contracts[0].code:
```

mythril/interfaces/cli.py
```diff
@@ -364,10 +365,16 @@ def main():
...
         elif args.disassemble:
-            easm_text = mythril.contracts[
-                0
-            ].get_easm()  # or mythril.disassemble(mythril.contracts[0])
-            sys.stdout.write(easm_text)
+            output = ""
+            easm_text = mythril.contracts[0].get_easm()
+            # or mythril.disassemble(mythril.contracts[0])
+
+            if easm_text:
```
JoranHonig:
```suggestion
            if mythril.contracts[0].code::
```

JoranHonig: And then you can move `easm_text = mythril.contracts[0].get_easm()` to this if statement

mythril/interfaces/cli.py
```diff
@@ -364,10 +365,16 @@ def main():
...
+                output += "Runtime Disassembly: \n" + easm_text
+            if code:
+                output += "Disassembly: \n" + mythril.contracts[0].get_creation_easm()
+
+            sys.stdout.write(output)
```
JoranHonig: Why not print?

cryptobarbossa:
Neat solutions @JoranHonig

mythril/interfaces/cli.py
nbanmp: I think you want keep `sys.stdout.write` here. The `print()` adds an extra newline.

PR #788, opened 2018-12-05
Initial smt abstraction PR

JoranHonig:
Todo:
- [ ] documentation
- [ ] value property
- [ ] is_symbolic property

### DIFF
--- a/mythril/laser/smt/bitvec.py
+++ b/mythril/laser/smt/bitvec.py
@@ -24,19 +24,19 @@ class BitVec(Expression):
         return BitVec(self.raw / other.raw, annotations=union)
 
 
-def UREM(a: BitVec, b: BitVec) -> BitVec:
+def URem(a: BitVec, b: BitVec) -> BitVec:
     union = a.annotations + b.annotations
     return BitVec(z3.URem(a.raw, b.raw), annotations=union)
 
 
-def SREM(a: BitVec, b: BitVec) -> BitVec:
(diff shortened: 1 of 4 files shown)


mythril/laser/smt/bitvec.py
```diff
@@ -0,0 +1,94 @@
...
+    union = a.annotations + b.annotations
+    return BitVec(z3.SRem(a.raw, b.raw), annotations=union)
+
+
+def UDiv(a: BitVec, b: BitVec) -> BitVec:
+    union = a.annotations + b.annotations
+    return BitVec(z3.UDiv(a.raw, b.raw), annotations=union)
+
+
+def UDiv(a: BitVec, b: BitVec) -> BitVec:
```
norhh: Am I missing something or are there 2 UDivs here?

JoranHonig: the sdiv is implemented in __truediv__

JoranHonig: but you are right,, there are two udivs ^^'

b-mueller:
Look goods.  I was just wondering if we should be adding some tests for `_Z3SymbolFactory`? I'm fine with adding those tests though due to time constraints.

JoranHonig:
@b-mueller, most of this will be tested in my next pr where I'm going to be integrating this into mythril as a whole. 
Because there will probably still be some changes to this code, I'd rather fully test it at that point

PR #803, opened 2018-12-07
Callback refactor for dependence on predictable vars

### LINKED ISSUE #793: Refactor: Callback dependence_on_predictable_vars.py
## Description
The delegate call module `mythril/analysis/modules/dependence_on_predictable_vars.py` should be refactored to work as a callback module.

An example refactored module is `mythril/analysis/modules/suicide.py`

child of #226

norhh:
fixes #793


mythril/analysis/call_helpers.py
```diff
@@ -0,0 +1,49 @@
+from mythril.analysis.ops import VarType, Call, get_variable
+
+
+def get_call_from_state(state):
```
nbanmp: Thanks for this helper. :smile: Can you add some type annotations?

mythril/analysis/call_helpers.py
```diff
@@ -0,0 +1,49 @@
...
+        gas, to, meminstart, meminsz, memoutstart, memoutsz = (
+            get_variable(stack[-1]),
+            get_variable(stack[-2]),
+            get_variable(stack[-3]),
+            get_variable(stack[-4]),
+            get_variable(stack[-5]),
+            get_variable(stack[-6]),
+        )
+
+        return Call(state.node, state, None, op, to, gas)
```
nbanmp: Do we want to check if `meminstart` and `meminsz` are concrete here and potentially create the Call with the calldata from memory?

mythril/analysis/modules/dependence_on_predictable_vars.py
```diff
@@ -14,174 +14,180 @@ def __init__(self):
...
                 "based on predictable variables such as block.coinbase, "
                 "block.gaslimit, block.timestamp, block.number"
             ),
+            entrypoint="callback",
         )
+        self._issues = []
 
-    def execute(self, statespace):
+    @property
+    def issues(self):
```
nbanmp: Can we get some fun type annotations here, `execute`, `solve`, `_analyze_states`, and in the other functions that might need them?

mythril/analysis/modules/dependence_on_predictable_vars.py
```diff
@@ -14,174 +14,180 @@ def __init__(self):
...
-                                )
-                                issues.append(issue)
-                                break
+                    break
+            else:
+                r = re.search(r"storage_([a-z0-9_&^]+)", str(constraint))
+                if r:  # block.blockhash(storage_0)
+
+                    """
+                    We actually can do better here by adding a constraint blockhash_block_storage_0 == 0
```
nbanmp: Why don't we take the opportunity refactoring the module to do this?

mythril/analysis/modules/dependence_on_predictable_vars.py
norhh: Done

mythril/analysis/call_helpers.py
norhh: Done

mythril/analysis/modules/dependence_on_predictable_vars.py
norhh: We could, but unfortunately the storage is concrete currently(previously it was unconstrained) so that should be changed before doing something here.We can change but we can't test it without changing that, so currently this entire if condition is redundant :cry:

mythril/analysis/call_helpers.py
nbanmp: I don't think we currently need it, but do we need/use it in the above case either? There might be a case where it's useful to extract the calldata for a delegatecall or different opcode, but it's not currently needed. You make the call here (that was also a pun btw).

norhh: Just to clarify this `Call` is the class in the `mythril.analysis.ops` which comes handy to manage Call type opcodes right?

nbanmp: That's what I understood.

norhh: It's easy to maintain various variables with call, so lets leave it that way, WDYT?

nbanmp: That's fine. :ok_hand:

mythril/analysis/modules/dependence_on_predictable_vars.py
```diff
@@ -16,172 +17,180 @@ def __init__(self):
...
+detector = PredictableDependenceModule()
+
+
+def _analyze_states(state: GlobalState) -> list:
+    issues = []
+    call = get_call_from_state(state)
+    if call is None:
+        return []
+    if "callvalue" in str(call.value):
+        log.debug("[DEPENDENCE_ON_PREDICTABLE_VARS] Skipping refund function")
```
nbanmp: We don't need to print `[DEPENDENCE_ON_PREDICTABLE_VARS]` with named loggers anymore.

norhh: Oh! cool, it was previously like that, so i left it that way.

PR #813, opened 2018-12-10
Extention of smt capabilities

### LINKED ISSUE #533: SMT abstraction

JoranHonig:
Implemented features:
- Implement the use of the symbol factory
- Implement missing features in the smt lib

Notable things that are not fully there:

- Sometimes integers get passed into smt functions, right now where this is done I added some code in the abstraction layer to deal with that. Ideally this will not be necessary in the future
- Not every part of z3 has been abstracted, namely sat result, exceptions and  models. This should be done in a later pr

closes #533


tests/laser/evm_testsuite/evm_test.py
```diff
@@ -133,10 +134,10 @@ def test_vmtest(
...
                 expected = int(value, 16)
                 actual = account.storage[int(index, 16)]
-                if isinstance(actual, ExprRef):
-                    actual = model.eval(actual)
+                if isinstance(actual, Expression):
+                    actual = actual.value
                     actual = (
-                        1 if actual == True else 0 if actual == False else actual
+                        1 if actual is True else 0 if actual is False else actual
                     )  # Comparisons should be done with == than 'is' here as actual can be a BoolRef
```
dmuhs: Is this comment still valid?

JoranHonig: no, I'll remove it

PR #831, opened 2018-12-15
Report issues when exception occurs

### LINKED ISSUE #810: report all issues till then when a failure occurs
## Description
Whenever an uncaught exception is raised then mythril should report all the issues it has encountered till then and also warn the user regarding the exception.This might be more useful after callback refactor.

## Background
Bugs might keep popping up, so people don't have to wait till the bug is fixed and atleast they would get some partial report.

nbanmp:
closes #810


mythril/mythril.py
```diff
@@ -481,23 +481,31 @@ def fire_lasers(
...
+                    ),
+                    max_depth=max_depth,
+                    execution_timeout=execution_timeout,
+                    create_timeout=create_timeout,
+                    transaction_count=transaction_count,
+                    modules=modules,
+                )
+                issues = fire_lasers(sym, modules)
+            except KeyboardInterrupt:
+                log.critical("Keyboard Interrupt")
```
norhh: Is this critical? :rofl:

mythril/mythril.py
```diff
@@ -481,23 +481,31 @@ def fire_lasers(
...
+                issues = retrieve_callback_issues(modules)
+            except Exception:
+                log.critical(
+                    "Exception occurred, aborting analysis\n" + traceback.format_exc()
```
norhh: We can also mention here to report this to mythril github page.

mythril/mythril.py
nbanmp: Possibly. It does kind of kill the program. If not critical, what level do you think it should be?

mythril/mythril.py
```diff
@@ -481,23 +481,31 @@ def fire_lasers(
...
+                )
+                issues = retrieve_callback_issues(modules)
```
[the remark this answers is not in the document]
nbanmp: Issues are gathered differently depending on whether or not there was an exception. `finally` executes whether or not an exception is thrown, so I'm not sure it would be appropriate in this situation. This StackOverflow question makes me think it would be nicer just to leave the one extra line in: <https://stackoverflow.com/questions/34796808/share-python-code-when-handling-multiple-exceptions>.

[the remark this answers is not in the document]
JoranHonig: Okay, lets just keep it in there then

PR #832, opened 2018-12-16
Some minor fixes

norhh:
- Checks whether call object is None or not in delegatecall module
- Converts python bool to z3 bool to support `Or()` in sstore.

### DIFF
--- a/mythril/laser/ethereum/instructions.py
+++ b/mythril/laser/ethereum/instructions.py
@@ -3,7 +3,6 @@ import logging
 
 from copy import copy, deepcopy
 from typing import Callable, List, Union
-from functools import reduce
 
 from ethereum import utils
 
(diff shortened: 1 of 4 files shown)


mythril/laser/ethereum/instructions.py
```diff
@@ -1004,7 +1004,7 @@ def sstore_(self, global_state: GlobalState) -> List[GlobalState]:
             keccak_keys = filter(keccak_function_manager.is_keccak, storage_keys)
 
             results = []
-            new = False
+            new = Bool(False)
```
JoranHonig: We should use the symbol factory for this. But it doesn't have a method for Bool yet, could you add that?

norhh: Done

mythril/laser/smt/__init__.py
```diff
@@ -56,6 +66,17 @@ class _SmtSymbolFactory(SymbolFactory):
...
+    @staticmethod
+    def Bool(value: bool, annotations=None):
+        """
+        Creates a Bool with concrete value
+        :param value: The boolean value
+        :param annotations: The annotations to initialize the bool with
+        :return: The freshly created Bool()
+        """
+        raw = z3.Bool(value)
+        return Bool(raw)
```
JoranHonig: the annotations need to be passed as a parameter right?

norhh: argh!

PR #835, opened 2018-12-17
Pop arguments for staticcall

nbanmp:
Now also adds support for native contracts called with staticcall and fixes some incorrect natives test cases.


mythril/laser/ethereum/instructions.py
```diff
@@ -1609,7 +1571,31 @@ def delegatecall_post(self, global_state: GlobalState) -> List[GlobalState]:
     def staticcall_(self, global_state: GlobalState) -> List[GlobalState]:
         # TODO: implement me
```
norhh: can you remove this?

nbanmp: It's not implemented. Only natives are implemented.

mythril/laser/ethereum/instructions.py
```diff
@@ -1609,7 +1571,31 @@ def delegatecall_post(self, global_state: GlobalState) -> List[GlobalState]:
...
+            )
+            global_state.mstate.stack.append(
+                global_state.new_bitvec("retval_" + str(instr["address"]), 256)
+            )
+            return [global_state]
+
         global_state.mstate.stack.append(
             global_state.new_bitvec("retval_" + str(instr["address"]), 256)
         )
+        if 0 < int(callee_address, 16) < 5:
```
norhh: can this line be somehow be merged into native_call, or other way is can you use a constant for the 5, as in the future we must increase the native contracts count, so we might need to change this number in multiple places, so localizing it might be better.

PR #837, opened 2018-12-18
Unchecked retval

### LINKED ISSUE #794: Refactor: Callback unchecked_retval.py
## Description
The delegate call module `mythril/analysis/modules/unchecked_retval.py` should be refactored to work as a callback module.

This module is used to check a property that holds over a trace, therefore this one requires the use of a custom StateAnnotation

child of #226

nbanmp:
Closes #794.


mythril/analysis/modules/delegatecall.py
```diff
@@ -18,7 +18,8 @@ def __init__(self):
         super().__init__(
             name="DELEGATECALL Usage in Fallback Function",
             swc_id=DELEGATECALL_TO_UNTRUSTED_CONTRACT,
-            hooks=["DELEGATECALL"],
+            pre_hooks=["DELEGATECALL"],
+            post_hooks=[],
```
norhh: I think you can just default post_hooks to [] in the constructor itself rather than adding it in each place, now it doesn't matter anyway.

nbanmp: I'll change that and do the same with `pre_hooks` too. It would make the code a bit nicer.

PR #848, opened 2018-12-24
Configure call to webhook in Circle CI build

s0b0lev:
This variables should be added into Circle CI ENV variables:
```
ORGANIZATION 
WEBHOOK_PROJECT
CIRCLE_TOKEN
```

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -77,9 +77,7 @@ jobs:
       - run:
           name: Call webhook
           command: |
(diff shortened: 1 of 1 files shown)


norhh:
@s0b0lev , can you mention the exact variable values that are to be added?

norhh:
@s0b0lev , any updates?

s0b0lev:
@norhh I am still looking for a reason why environment variables are not injected into build. Also, I didn't find any other variables which was declared in Circle CI build env as well.

s0b0lev:
@norhh , thanks to @fgimenez , issue is within settings, which is not allowed to inject env variables on forks. So I should use mythril-classic branch, and it should work. I will close this PR and create another one.

PR #852, opened 2018-12-27
Add JSONv2 standard format and refactor reporting

### LINKED ISSUE #844: Implement the new JSON output format
## Description

Update JSON output to follow the new specification:

```
{
    "issues": [
        {
            "swcID": "",
            "swcTitle": "",
            "description": {
                "head": "", 
                "tail": ""
            },
            "severity": "",
            "locations": [ 
                {
                 "sourceMap": "", 
                }, ...
            ],
            "extra": {} 
        }
    ],
    "sourceType": "", 
    "sourceFormat": ""
    "sourceList": [ "", ... ], 
    "meta": {}
}
```

### Details

**1. When the input is one or more Solidity files return a SourceList as follows:**

```
    "sourceType": "solidity-file", 
    "sourceFormat": "text"
    "sourceList": [ "file1.sol, "file2.sol" ,...]
```
 
Source maps for each issue follow the standard solc format:

```
"sourceMap": "24:693:0"
```

**2. When the input is bytecode (`-c`):**

```
"sourceType": "raw-bytecode"
"sourceFormat": "evm-byzantium-bytecode",
"sourceList": [ "0xeda9b01a853deb0dbc2380aa9c3053529aec777e2aae6818f5819cd5feebb1ff" ],
```

Note that the entries in "sourceList" are sha3 hashes of the contract bytecode(s).

Sourcemap again follows the solc format. Mythril usually only returns a simple bytecode offset per issue, e.g.:

```
"sourceMap": "10"
```

**3. When the input is an Ethereum address bytecode (`-a`):**

Mapping is the same as for raw bytecode but `sourceType` changes to `ethereum-address`.

b-mueller:
@norhh should be all done, however I noticed that code snippets are missing from the text reports. Haven't been able to figure out why yet so I also didn't update the expected report results (a lot of the descriptions have changed).

Would appreciate it if one of you guys could fix this up 😉

fixes #844


norhh:
Fixed :+1: . And yes that was my mistake :sweat_smile:

b-mueller:
Awesome!

norhh:
I wanted to ask this, should we change the json option to standard format or should we write a new option for the same?

b-mueller:
I think having a new option is fine.

norhh:
currently we are modifying the current json option, so should we change that?

b-mueller:
Oh, we are? I didn't notice that.

mythril/analysis/modules/delegatecall.py
```diff
@@ -78,13 +78,18 @@ def _concrete_call(
...
+        description_tail="The smart contract forwards the received calldata via delegatecall. Note that callers"
+        "can execute arbitrary functions in the callee contract and that the callee contract "
+        "can access the storage of the calling contract. "
+        "Make sure that the callee contract is audited properly.",
         gas_used=(state.mstate.min_gas_used, state.mstate.max_gas_used),
     )
 
     issue.description = (
-        "This contract forwards its call data via DELEGATECALL in its fallback function. "
+        ""
```
nbanmp: This line can be deleted without any side effect, right?

mythril/analysis/modules/dependence_on_predictable_vars.py
```diff
@@ -159,19 +164,20 @@ def _analyze_states(state: GlobalState) -> list:
 
                     index = r.group(1)
                     if index and solve(call):
-                        description += (
-                            "block.blockhash() is calculated using a value from storage "
+                        description = (
+                            "Block.blockhash() is calculated using a value from storage "
```
nbanmp: When starting a sentence with a segment of code, is it correct to capitalize the code? Probably add a word or two in front of `block.blockhash()`.

mythril/analysis/modules/deprecated_ops.py
```diff
@@ -17,8 +17,9 @@ def _analyze_state(state):
 
     if instruction["opcode"] == "ORIGIN":
         log.debug("ORIGIN in function " + node.function_name)
-        title = "Use of tx.origin"
-        description = (
+        title = "Use of tx.origin is deprecated."
+        description_head = "Use of tx.origin is deprecated."
+        description_tail = (
             "The function `{}` retrieves the transaction origin (tx.origin) using the ORIGIN opcode. "
```
nbanmp: The function name was removed from the description elsewhere (at least in predictable vars). Why was it removed there and not removed here?

tests/testdata/outputs_expected/calls.sol.o.json
```diff
@@ -1,102 +1 @@
...
..."title": "Unchecked Call Return Value"}, {"SourceMap": null, "address": 912, "contract": "Unknown", "debug": "<DEBUG-DATA>", "description": "A call to a user-supplied address is executed.\nThe callee address of an external message call can be set by the caller. Note that the callee can contain arbitrary code and may re-enter any function in this contract. Review the business logic carefully to prevent averse effects on thecontract state.", "function": "calluseraddress(address)", "max_gas_used": 616, "min_gas_used": 335, "severity": "Medium", "swc-id": "107", "title": "External Call To User-Supplied Address"}, {"SourceMap": null, "address": 912, "contract": "Unknown", "debug": "<DEBUG-DATA>", "description": "The return value of a message call is not checked.\nExternal calls return a boolean value. If the callee contract halts with an exception, 'false' is returned and execution continues in the caller. It is usually recommended to wrap external calls into a require statement to prevent unexpected states.", "function": "calluseraddress(address)", "max_gas_used": 35336, "min_gas_used": 1055, "severity": "Low", "swc-id": "104", "title": "Unchecked Call Return Value"}], "success": true}
```
nbanmp: Can we pretty-print these files please?

mythril/analysis/modules/delegatecall.py
b-mueller: Deleted.

mythril/analysis/modules/deprecated_ops.py
b-mueller: Sloppiness 😅

mythril/analysis/modules/integer.py
```diff
@@ -168,20 +170,24 @@ def _handle_jumpi(self, state):
...
             issue = Issue(
                 contract=node.contract_name,
                 function_name=node.function_name,
                 address=ostate.get_current_instruction()["address"],
                 swc_id=INTEGER_OVERFLOW_AND_UNDERFLOW,
                 bytecode=ostate.environment.code.bytecode,
                 title="Integer Overflow",
-                _type="Warning",
+                severity="High",
+                description_head="The {} can overflow.".format(annotation.operator),
```
JoranHonig: The title and description head should be handled the same as how an SSTORE is handled

mythril/analysis/modules/multiple_sends.py
```diff
@@ -71,28 +71,34 @@ def _analyze_state(state: GlobalState):
...
+            description_tail = (
+                "Consecutive calls are executed at the following bytecode offsets:\n"
+            )
+
+            for call in calls:
+                description_tail += "Offset: {}\n".format(
+                    call.state.get_current_instruction()["address"]
+                )
+
+                description_tail = (
```
JoranHonig: description_tail is overwritten here

JoranHonig: Should probably not be indented :smile:

mythril/analysis/report.py
```diff
@@ -18,23 +23,27 @@ def __init__(
...
         debug="",
     ):
 
         self.title = title
         self.contract = contract
         self.function = function_name
         self.address = address
-        self.description = description
-        self.type = _type
+        self.description_head = (description_head,)
```
JoranHonig: why does this need to be a tuple?

mythril/analysis/report.py
```diff
@@ -55,11 +64,12 @@ def as_dict(self):
             "contract": self.contract,
             "description": self.description,
             "function": self.function,
-            "type": self.type,
+            "severity": self.severity,
             "address": self.address,
             "debug": self.debug,
             "min_gas_used": self.min_gas_used,
             "max_gas_used": self.max_gas_used,
+            "SourceMap": self.source_mapping,
```
JoranHonig: should be `sourceMap` right?

mythril/support/source_support.py
```diff
@@ -0,0 +1,31 @@
+from mythril.solidity.soliditycontract import SolidityContract
+from mythril.ethereum.evmcontract import EVMContract
+
+
+class Source:
```
JoranHonig: This class still requires documentation strings

mythril/ethereum/evmcontract.py
```diff
@@ -23,6 +27,15 @@ def __init__(
         self.creation_disassembly = Disassembly(
             creation_code, enable_online_lookup=enable_online_lookup
         )
+        try:
```
JoranHonig: I think this fits better as an @property function. Separates this logic from the constructor and allows us to create a well documented acessor

mythril/analysis/report.py
```diff
@@ -113,16 +136,45 @@ def as_json(self):
 
     def as_swc_standard_format(self):
```
JoranHonig: should the format be named jsonv2 or swc_standard_format

mythril/analysis/report.py
```diff
@@ -72,24 +82,37 @@ def as_dict(self):
...
-    def __init__(self, verbose=False):
+    def __init__(self, verbose=False, source=None):
         self.issues = {}
         self.verbose = verbose
-        pass
+        self.solc_version = ""
+        self.meta = {}
+
+        if source is not None:
+            self.source = source
```
JoranHonig:
```suggestion
             self.source = source or Source()
```

mythril/analysis/report.py
```diff
@@ -72,24 +82,37 @@ def as_dict(self):
...
+        else:
+            self.source = Source()
+
+        self.source_type = self.source.source_type
```
JoranHonig: Shouldn't these be @property functions? If source changes self.source_type should also change right?

mythril/analysis/report.py
```diff
@@ -113,16 +136,45 @@ def as_json(self):
...
-                    "bytecodeOffset": issue.address,
-                    "codeHash": issue.bytecode_hash,
+                    "swcID": issue.swc_id,
+                    "swcTitle": title,
+                    "description": {
+                        "head": issue.description_head,
+                        "tail": issue.description_tail,
+                    },
+                    "severity": issue.severity,
+                    "locations": [{"sourceMap": "%d:1:%d" % (issue.address, idx)}],
```
JoranHonig: if the source format is bytecode then the location should only be the bytecode offset

mythril/solidity/soliditycontract.py
```diff
@@ -104,8 +105,12 @@ def get_source_info(self, address, constructor=False):
         disassembly = self.creation_disassembly if constructor else self.disassembly
         mappings = self.constructor_mappings if constructor else self.mappings
         index = helper.get_instruction_index(disassembly.instruction_list, address)
+        if index is None:
+            index = helper.get_instruction_index(
+                self.creation_disassembly.instruction_list, address
+            )
+            mappings = self.constructor_mappings
```
JoranHonig: this overwrites mappings, which it probably shouldn't

mythril/ethereum/evmcontract.py
```diff
@@ -23,6 +27,15 @@ def __init__(
...
+            keccak = sha3.keccak_256()
+            keccak.update(bytes.fromhex(self.code[2:]))
+            self.bytecode_hash = "0x" + keccak.hexdigest()
```
JoranHonig: It should also have a hash for the creation bytecode right?

mythril/support/truffle.py
```diff
@@ -145,7 +145,11 @@ def get_sigs_from_truffle(sigs, contract_data):
 
 def get_mappings(source, deployed_source_map):
     mappings = []
+    prev_item = ""
     for item in deployed_source_map:
```
JoranHonig:
```suggestion
     for item in [e for e in deployed_source_map if e]:
```

JoranHonig: I think this allows you to not use the prev_item stuff here

mythril/solidity/soliditycontract.py
```diff
@@ -115,12 +120,14 @@ def get_source_info(self, address, constructor=False):
...
         )
         lineno = mappings[index].lineno
-
-        return SourceCodeInfo(filename, lineno, code)
+        return SourceCodeInfo(filename, lineno, code, mappings[index].solc_mapping)
 
     def _get_solc_mappings(self, srcmap, constructor=False):
         mappings = self.constructor_mappings if constructor else self.mappings
+        prev_item = ""
         for item in srcmap:
```
JoranHonig:
```suggestion
         for item in [e for e in srcmap if e]:
```

mythril/analysis/report.py
```diff
@@ -113,16 +136,45 @@ def as_json(self):
...
-            "issues": [
+
+        _issues = []
+        sourceList = []
+
+        for key, issue in self.issues.items():
+
+            if issue.bytecode_hash not in sourceList:
+                idx = len(sourceList)
+                sourceList.append(issue.bytecode_hash)
```
JoranHonig: If we do this then why bother with the bytecode_hash in the solidity source file?

mythril/analysis/modules/integer.py
b-mueller: Can you clarify that?

rocky:
See Issue #865

Also it looks like `-o jsonv2` should be added to the help. Thanks.

b-mueller:
I added a couple more commits. Originally wanted to create a separate branch for that but I messed up locally, anyway it's not a big change, just a nicer way of displaying transaction data.

mythril/analysis/report.py
b-mueller: It's not really a swc standard so I guess `jsonv2` makes more sense.

mythril/support/truffle.py
norhh: But in case of mappings we can't ignore when e=="", as that would miss a mapping right?(when e=="" then solidity says to use the previous map for that point too) We need to add them to the mappings list. This might ignore such mappings.

mythril/analysis/report.py
norhh: aren't bytecode offset and address same?

mythril/solidity/soliditycontract.py
norhh: might result in same problem as above

mythril/analysis/report.py
norhh: Yes, it might require in solidity source file but not in the issues

norhh: even if there are no issues i think the bytecode hash needs to be printed in the source list

mythril/analysis/modules/integer.py
```diff
@@ -124,6 +124,29 @@ def _make_bitvec_if_not(stack, index):
...
+                _type.lower(),
+                _type.lower(),
+            )
+        )
+
+    @staticmethod
+    def _get_title(_type):
+        return "Integer {}".format(_type)
+
     def _handle_sstore(self, state):
```
norhh: Both handle_sstore and handle_jump look same currently, is there any intention to make them divergent, if thats not the case then we can remove one function(in a different PR, definitely not in this one)

tests/report_test.py
```diff
@@ -77,7 +77,9 @@ def _assert_empty(changed_files, postfix):
 
 
 def _assert_empty_json(changed_files, postfix=".json"):
-    """ Asserts there are no changed files and otherwise builds error message"""
+    """Asserts there are no changed files and otherwise builds error
+    message."""
+    postfix = ".json"
```
JoranHonig: You shouldn't overwrite the postfix here, I think that changes some test behaviour

PR #853, opened 2018-12-27
Implement sigalrm timeout

### LINKED ISSUE #673: Add option to specify maximum total execution time
## Description

Add an option to specify maximum overall runtime in seconds. Mythril should terminate after the specified time and return all issues discovered as well as the number of transactions executed and code coverage achieved.

## Background

This feature is needed for usability reasons. For example, during SDLC, users might want to run best-effort security analysis that run no longer than a minute.

## Dependencies

A key dependency refactoring the security analysis modules to callbacks (#226).

nbanmp:
Closes #673

Tell Windows I'm sorry.

### DIFF
--- a/mythril/laser/ethereum/svm.py
+++ b/mythril/laser/ethereum/svm.py
@@ -70,7 +70,7 @@ class LaserEVM:
         self.max_depth = max_depth
         self.transaction_count = transaction_count
 
-        self.execution_timeout = execution_timeout
+        self.execution_timeout = execution_timeout or 0
         self.create_timeout = create_timeout
 
         self.time = None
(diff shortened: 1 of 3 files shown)


mythril/laser/ethereum/svm.py
```diff
@@ -172,13 +182,7 @@ def _get_covered_instructions(self) -> int:
...
         final_states = []
         for global_state in self.strategy:
-            if self.execution_timeout and not create:
-                if (
-                    self.time + timedelta(seconds=self.execution_timeout)
-                    <= datetime.now()
-                ):
-                    return final_states + [global_state] if track_gas else None
-            elif self.create_timeout and create:
+            if self.create_timeout and create:
```
norhh: is having this nested if necessary, both of these can be combined right? It might reduce some nesting.

PR #863, opened 2019-01-03
Cache the ether_thief addresses whose issues were already raised

norhh:
gives a reasonable speed up on an average case. speedup is proportional to number of paths to the vulnerability.So this kind of optimization will give a killer speedup for integer module if used there.

### DIFF
--- a/mythril/analysis/modules/base.py
+++ b/mythril/analysis/modules/base.py
@@ -25,6 +25,7 @@ class DetectionModule:
                 self.name,
             )
         self.entrypoint = entrypoint
+        self.cache_addresses = {}
 
     def execute(self, statespace):
         raise NotImplementedError()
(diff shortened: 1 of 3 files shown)


mythril/analysis/modules/base.py
```diff
@@ -25,6 +25,16 @@ def __init__(
...
+        self._issues = []
+        self.cache_addresses = {}
+
+    @property
+    def issues(self):
+        return self._issues
+
+    def reset_module(self):
+        self._issues = []
+        self.cache_addresses = {}
```
JoranHonig: `self._cache_addresses`? this is not something that needs to be publicly accessed right?

norhh: Changed

Issue #689, opened 2018-11-12
Implement transfer of ETH between accounts


b-mueller:
## Description

Keeping track of account balances is a prerequisite for implementing more precise analysis. Specifically, without this feature it is impossible to analyze the flow of ETH between contracts over multiple transactions (e.g. can the sender withdraw more ETH than they paid in)?

Account balances are currently not updated when CALL-family instructions are executed. AFAIK the correct behavior is to subtract call.value from the sender and add it to the receiver unless the transaction is reverted. This should not be too difficult to implement:

1. Create an individual symbolic variable for each account created;
2. Add / subtract balance as needed in message call pre/post-processing.

b-mueller:
Note - labeling this as "high priority" because IMO this change is needed to make the ether_send module meaningful.

b-mueller:
Perhaps as a first step, if `call_` encounters an existing callee account without code, we could update balances and put `True` on the stack instead of raising `TransactionStartSignal`?

JoranHonig:
Account balance modelling is largely implemented. 
What still seems to be missing is the balance change for "user" initiated transactions

JoranHonig:
The approach is somewhat different from what @b-mueller proposed, as it leverages array theories to be a bit quicker.

JoranHonig:
This is done

PR #868, opened 2019-01-04
Prune non mutated states 

### LINKED ISSUE #867: Pruning non mutated worldstates
## Description
If there is a path leading to a new world state and this path does not mutate the world state, then the resulting worldstate should not be considered for further analysis, as this will be unnecessary.

JoranHonig:
## Description
This is an initial effort to implement pruning non mutated states.
I also introduced a Plugin system to laser, and intend to flesh this out a bit more. Things like #849 should fit neatly into this bracket, we can then also support command line selection of the different plugins

closes #867


mythril/laser/ethereum/plugins/mutation_pruner.py
```diff
@@ -0,0 +1,56 @@
...
+        """
+
+        @symbolic_vm.pre_hook("SSTORE")
+        def mutator_hook(global_state: GlobalState):
+            global_state.annotate(MutationAnnotation())
+
+        @symbolic_vm.laser_hook("add_world_state")
+        def world_state_filter_hook(global_state: GlobalState):
+            if isinstance(
+                global_state.current_transaction, ContractCreationTransaction
```
norhh: can you explain why is this avoided for creation transaction?It isn't clear to me.

mythril/laser/ethereum/plugins/mutation_pruner.py
```diff
@@ -0,0 +1,56 @@
...
+
+    def initialize(self, symbolic_vm: LaserEVM):
+        """Initializes the mutation pruner
+
+        Introduces hooks for SSTORE operations
+        :param symbolic_vm:
+        :return:
+        """
+
+        @symbolic_vm.pre_hook("SSTORE")
```
nbanmp: It's worth noting that call instructions also mutate the contract's state if their value is > 0. We don't accurately model contract balances though (#689), so this doesn't necessarily have to be implemented here immediately, but we should be aware of it.

mythril/laser/ethereum/svm.py
```diff
@@ -480,6 +483,47 @@ def _new_node_state(
...
+
+        :param hook_type:
+        :param hook_dict:
+        """
+        if hook_type == "pre":
+            entrypoint = self.pre_hooks
+        elif hook_type == "post":
+            entrypoint = self.post_hooks
+        else:
+            raise ValueError(
```
nbanmp: The exceptions raised on a failure to find a hook are inconsistent. It's a `ValueError` in `register_hooks`, and a `NotImplementedError` for `register_laser_hooks`. Is this appropriate, or should we use the same exception for both?

mythril/laser/ethereum/svm.py
```diff
@@ -480,6 +483,47 @@ def _new_node_state(
...
+                "Invalid hook type %s. Must be one of {pre, post}", hook_type
+            )
+
+        for op_code, funcs in hook_dict.items():
+            entrypoint[op_code].extend(funcs)
+
+    def register_laser_hooks(self, hook_type: str, hook: Callable):
```
nbanmp: Docstring please!

mythril/laser/ethereum/plugins/mutation_pruner.py
b-mueller: Calls with >0 value do need to be considered mutated even though we don't model balance. Otherwise we can get false negatives on the Ether Thief check with counts Ether sent vs. Ether received over all transactions. At least It think so...

b-mueller:
Some unscientific experiments on my Mac:

```
                    DEVELOP                 PRUNE
                    nstates     time        nstates     time        Improvement (%)
-------------------------------------------------------------------------------
calls.sol           11          2m24s       3           5s          69%
exceptions.sol       9          1m54s                   --- failed ---
suicide.sol          2          0.9s        0           1.8s        50%
token.sol            3          9.5s        1           9.6s        0%
weak_random.sol    162          lol         152         lol         N/A
```

Looks promising! Notes:

For `exceptions.sol` no contract is created with the patch:

```
 $MYTH -x exceptions.sol -v4
mythril.support.signatures [INFO]: Using signature database at /Users/bernhardmueller/.mythril/signatures.db
mythril.analysis.security [INFO]: Found 1 detection modules
mythril.laser.ethereum.svm [INFO]: LASER EVM initialized with dynamic loader: <mythril.support.loader.DynLoader object at 0x10ffaef28>
mythril.analysis.security [INFO]: Found 10 detection modules
mythril.analysis.security [INFO]: Found 10 detection modules
mythril.laser.ethereum.svm [INFO]: Starting contract creation transaction
mythril.laser.ethereum.svm [INFO]: Finished contract creation, found 0 open states
mythril.laser.ethereum.svm [WARNING]: No contract was created during the execution of contract creation Increase the resources for creation execution (--max-depth or --create-timeout)
mythril.laser.ethereum.svm [INFO]: Starting message call transaction, iteration: 0, 0 initial states
mythril.laser.ethereum.svm [INFO]: Number of new instructions covered in tx 0: 0
mythril.laser.ethereum.svm [INFO]: Starting message call transaction, iteration: 1, 0 initial states
mythril.laser.ethereum.svm [INFO]: Number of new instructions covered in tx 1: 0
mythril.laser.ethereum.svm [INFO]: Finished symbolic execution
mythril.laser.ethereum.svm [INFO]: 3 nodes, 2 edges, 19 total states
mythril.analysis.security [INFO]: Starting analysis
mythril.analysis.security [INFO]: Found 1 detection modules
mythril.analysis.security [INFO]: Executing Transaction Order Dependence
mythril.analysis.security [INFO]: Found 10 detection modules
mythril.analysis.security [INFO]: Found 10 detection modules
```

mythril/laser/ethereum/plugins/mutation_pruner.py
JoranHonig: Because you don't want to apply this technique to creation transactions. By their nature creation transactions mutate the world state, or they finish unsuccesfully

mythril/laser/ethereum/plugins/mutation_pruner.py
JoranHonig: There are a few other instructions that mutate the state as well (suicide for example), however I'm not sure if they "matter" for multi-transactional analysis and opening up new possible paths

mythril/laser/ethereum/svm.py
JoranHonig: Nice, ofc!

mythril/laser/ethereum/plugins/mutation_pruner.py
nbanmp: @JoranHonig Why was this marked resolved? Did we decide that calls with >0 value do not need to be considered mutated?

mythril/laser/ethereum/svm.py
```diff
@@ -305,17 +308,17 @@ def execute_state(
             return [new_global_state], op_code
 
         except TransactionEndSignal as end_signal:
-            transaction, return_global_state = (
-                end_signal.global_state.transaction_stack.pop()
-            )
+            transaction, return_global_state = end_signal.global_state.transaction_stack[
+                -1
+            ]
```
nbanmp: I'm slightly confused for the reasoning for this change. Why was popping from the transaction stack bad?

JoranHonig: Because the callback is issued after this pop.
That means that for a creation transaction it will receive a global state with a transaction stack of [], and it will be impossible to determine the transaction that the global state was from

nbanmp: Ah. So is it then popped after the callback?

JoranHonig: I don't think it has to be popped at all (and I think it isn't)

mythril/laser/ethereum/plugins/mutation_pruner.py
```diff
@@ -0,0 +1,56 @@
...
+
+    This optimization inhibits path explosion caused by "clean" behaviour
+
+    The basic operation of this plugin is as follows:
+     - Hook all mutating operations and introduce a MutationAnnotation to the global state on execution of the hook
+     - Hook the svm EndTransaction on execution filter the states that do not have a mutation annotation
+
+    """
+
+    def __init__(self):
```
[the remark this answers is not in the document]
JoranHonig: no

mythril/laser/ethereum/plugins/mutation_pruner.py
```diff
@@ -0,0 +1,56 @@
...
+        :param symbolic_vm:
+        :return:
+        """
+
+        @symbolic_vm.pre_hook("SSTORE")
+        def mutator_hook(global_state: GlobalState):
+            global_state.annotate(MutationAnnotation())
+
+        @symbolic_vm.laser_hook("add_world_state")
+        def world_state_filter_hook(global_state: GlobalState):
```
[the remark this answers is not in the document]
JoranHonig: No, inner functions shouldn't be  (I think)

mythril/laser/ethereum/plugins/mutation_pruner.py
```diff
@@ -0,0 +1,56 @@
...
+    def initialize(self, symbolic_vm: LaserEVM):
+        """Initializes the mutation pruner
+
+        Introduces hooks for SSTORE operations
+        :param symbolic_vm:
+        :return:
+        """
+
+        @symbolic_vm.pre_hook("SSTORE")
+        def mutator_hook(global_state: GlobalState):
```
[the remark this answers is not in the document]
JoranHonig: No, inner functions shouldn't be  (I think)

mythril/laser/ethereum/plugins/mutation_pruner.py
JoranHonig: I don't think that this is a problem. 
If we have a transaction sequence of 
A -> B -> C
Where C does nothing but put money in the contract, then you find
A -> B -> C -> D 
Where D has the withdrawal. Since the vuln in D is independent of the balance change in C it must also be possible to find:
A -> B -> D
Therefore, if we abandon execution of C we should still find transaction D

mythril/laser/ethereum/plugins/mutation_pruner.py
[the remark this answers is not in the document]
nbanmp: In that case, should line 520 in svm not have a docstring?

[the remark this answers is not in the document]
JoranHonig: I don't think it's necessary.

mythril/laser/ethereum/plugins/mutation_pruner.py
JoranHonig: Ok I added a check so only traces with a constraint callvalue == 0 are ignored

nbanmp: Do we also want to hook `CALL` opcodes?

PR #869, opened 2019-01-05
Cache integer overflows/underflows, selfdestruct module

norhh:
also removes _cache_addresses from base as we aren't using it for all the modules.


mythril/analysis/modules/integer.py
```diff
@@ -156,14 +169,30 @@ def _handle_sstore(self, state):
             issue.description = "This binary {} operation can result in {}.\n".format(
                 annotation.operator, title.lower()
             )
+            address = ostate.get_current_instruction()["address"]
```
JoranHonig: can we make a private function for this as the same logic is being done at ln 68

mythril/analysis/modules/integer.py
```diff
@@ -192,17 +221,34 @@ def _handle_jumpi(self, state):
                 gas_used=(state.mstate.min_gas_used, state.mstate.max_gas_used),
             )
 
-            issue.description = "This binary {} operation can result in integer overflow.\n".format(
-                annotation.operator
+            issue.description = "This binary {} operation can result in {}.\n".format(
+                annotation.operator, title.lower()
             )
+            address = ostate.get_current_instruction()["address"]
```
JoranHonig: same here, this logic would be (effectively) equivalent to that at ln 68 right?

PR #870, opened 2019-01-05
Implement returndatacopy and returndatasize

nbanmp:
Also includes a few minor bugfixes


mythril/laser/ethereum/instructions.py
```diff
@@ -1057,9 +1057,45 @@ def returndatacopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
+        try:
+            concrete_memory_offset = helper.get_concrete_int(memory_offset)
+        except TypeError:
+            log.debug("Unsupported symbolic memory offset in RETURNDATACOPY")
+            return [global_state]
+
+        try:
+            concrete_return_offset = helper.get_concrete_int(return_offset)
+        except TypeError:
+            log.debug("Unsupported symbolic return offset in RETURNDATACOPY")
```
JoranHonig: If we return [globalstate] for these catches, then we don't need the ifstatement at 1090. Additionally we won't need ln 1066, 1067 if we do that, as everything would always be initialized after the try catches

mythril/laser/ethereum/instructions.py
```diff
@@ -1728,7 +1773,7 @@ def call_post(self, global_state: GlobalState) -> List[GlobalState]:
         instr = global_state.get_current_instruction()
 
         try:
-            callee_address, _, _, value, _, _, memory_out_offset, memory_out_size = get_call_parameters(
+            callee_address, callee_account, call_data, value, gas, memory_out_offset, memory_out_size = get_call_parameters(
```
JoranHonig: why did you add this change?

mythril/laser/ethereum/instructions.py
```diff
@@ -1694,14 +1739,14 @@ def call_(self, global_state: GlobalState) -> List[GlobalState]:
...
             return [global_state]
-        global_state.mstate.stack.append(
-            global_state.new_bitvec("retval_" + str(instr["address"]), 256)
-        )
 
         native_result = native_call(
             global_state, callee_address, call_data, memory_out_offset, memory_out_size
         )
         if native_result:
+            global_state.mstate.stack.append(
```
JoranHonig: Don't we know the success/failure of the native call?

mythril/laser/ethereum/instructions.py
```diff
@@ -1057,9 +1057,45 @@ def returndatacopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
+            concrete_size is None
+            or global_state.last_return_data is None
+            or concrete_return_offset is None
+        ):
+            return [global_state]
+
+        global_state.mstate.mem_extend(concrete_memory_offset, concrete_size)
+        for i in range(concrete_size):
+            global_state.mstate.memory[concrete_memory_offset + i] = (
+                global_state.last_return_data[concrete_return_offset + i]
```
JoranHonig: what if there is no last return data? returndatasize does have logic handling this

mythril/laser/ethereum/instructions.py
norhh: there was an extra `_` there compared to the return statement of the get_call_parameters. Probably we shouldn't return too many parameters :sweat_smile:, like returning a struct might be good.

mythril/laser/ethereum/instructions.py
nbanmp: It's checked in the above if.

mythril/laser/ethereum/instructions.py
nbanmp: There's currently a `# TODO: maybe use BitVec here constrained to 1` at the bottom of `native_call()`, which is probably a better place to put this anyway.

mythril/laser/ethereum/instructions.py
nbanmp: Yeah, `get_call_parameters` returns 7 parameters, and 8 were expected, so a `ValueError` was being raised and caught every time.

mythril/laser/ethereum/instructions.py
JoranHonig: Ah I missed that

mythril/laser/ethereum/instructions.py
JoranHonig: Why did you resolve this?

nbanmp: Because I added `return [global_state]` to these catches and fixed lines 1090, 1066, and 1067.

PR #871, opened 2019-01-07
Implement special case for the delegatecall proxy pattern

### LINKED ISSUE #675: Add a special case for passing original calldata with symbolic length

nbanmp:
closes #675

### DIFF
--- a/mythril/laser/ethereum/call.py
+++ b/mythril/laser/ethereum/call.py
@@ -4,6 +4,7 @@ parameters for the new global state."""
 
 import logging
 from typing import Union, List
+from z3 import Z3Exception
 
 from mythril.laser.ethereum import natives
 from mythril.laser.ethereum.gas import OPCODE_GAS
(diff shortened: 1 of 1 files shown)


mythril/laser/ethereum/call.py
```diff
@@ -179,16 +180,26 @@ def get_call_data(
...
+            uses_entire_calldata = simplify(
+                memory_size - global_state.environment.calldata.calldatasize == 0
             )
-        ]
-        call_data = ConcreteCalldata(transaction_id, calldata_from_mem)
-        log.debug("Calldata: " + str(call_data))
+        except Z3Exception:
+            pass
+
+        if uses_entire_calldata:
```
norhh: does this part ever get executed?

mythril/laser/ethereum/call.py
```diff
@@ -179,16 +180,26 @@ def get_call_data(
     )
 
     try:
-        calldata_from_mem = state.memory[
-            util.get_concrete_int(memory_start) : util.get_concrete_int(
-                memory_start + memory_size
+        uses_entire_calldata = False
+        try:
+            uses_entire_calldata = simplify(
```
norhh: is the "try" statement in line 182 needed to be covered for these lines? As i think TypeError should be caught from lines 194:197 and not from these lines right?

mythril/laser/ethereum/call.py
```diff
@@ -179,16 +180,26 @@ def get_call_data(
...
+                memory_size - global_state.environment.calldata.calldatasize == 0
             )
-        ]
-        call_data = ConcreteCalldata(transaction_id, calldata_from_mem)
-        log.debug("Calldata: " + str(call_data))
+        except Z3Exception:
```
norhh: I can't get the reason why Z3Exception is caught here? can you explain.

nbanmp: :sweat_smile: A previous version had `if simplify(...): `, which would raise an exception. It's unneeded now.

PR #872, opened 2019-01-07
Fix bug accessing on chain contracts with addresses starting with 0s

nbanmp:

PR #881, opened 2019-01-11
Refactor timeouts

### LINKED ISSUE #875: --execution-timeout is not proper for rubixi.sol
## How to Reproduce

For example:

```console
$ myth -x solidity_examples/rubixi.sol --execution-timeout 1
takes long time
...
$
```

## Expected behavior
execution should finish in 1s

## Environment
Please modify for your setup

- Mythril version: output from  `develop`
- Solidity compiler and version: `0.5.0`
- Python version: `3.6`
- OS and Version: ubuntu 17.10

JoranHonig:
With this pr the timeouts will work using a ThreadPoolExecutor instead of sigalarms, which has tuned out to cause some problems: #875, #872, #849 
To my understanding, this is caused by z3 also handling this signal.

@nbanmp you have been involved with the development of this feature and the bug reports, could you check if this pr fixes the problem/goes in the correct direction?

closes #875


norhh:
It isn't working for rubixi.sol example in my laptop. can you check `time ./myth -x solidity_examples/rubixi.sol --execution-timeout 10` or even `--execution-timeout 1` doesn't work.

mythril/laser/ethereum/svm.py
```diff
@@ -180,6 +171,22 @@ def sym_exec(
...
+            f_results = executor.submit(
+                self._sym_exec,
+                main_address=main_address,
+                creation_code=creation_code,
+                contract_name=contract_name,
+            )
+            try:
+                return f_results.result(timeout)
+            except TimeoutError:
+                return
```
nbanmp: It might be nice to log something here so that it is clear that a timeout occurred? WDYT?

nbanmp: In `rubixi.sol`, as @norhh pointed out, execution still hangs. It turns out that the `TimeoutError` is being raised and caught, but execution gets stuck returning, presumably waiting patiently for the `with` to gracefully terminate `executor`. We might need to kill the thread a little more aggressively.

mythril/laser/ethereum/svm.py
```diff
@@ -229,12 +236,15 @@ def exec(self, create=False, track_gas=False) -> Union[List[GlobalState], None]:
         """
         final_states = []
         for global_state in self.strategy:
-            if (
-                self.create_timeout
-                and create
-                and self.time + timedelta(seconds=self.create_timeout) <= datetime.now()
-            ):
-                return final_states + [global_state] if track_gas else None
+            if self.execution_timeout and not create:
```
nbanmp: Is this still needed?

mythril/laser/ethereum/svm.py
JoranHonig: You are right, unfortunately Executors & futures don't allow this so I used ThreadPool instead

mythril/laser/ethereum/svm.py
JoranHonig: Yes and no. This code doesn't do much now. But there is a problem when you want to have multiple transactions, as now the execution timeout is used for the entire thing. In another pr I adapt this so it uses a timeout of execution_timeout/#transactions as a limit per transaction

norhh:
I was trying it out with BECToken.sol, it seems to throw some weird Z3exceptions and segfaults without returning issues.

PR #885, opened 2019-01-14
Add z3 mypy stubs; Add types to mythril/laser/smt/*.py

cd1m0:
This PR:

- adds mypy typing stubs for the z3 module
- adds more types to files under mythril/laser/smt
- fixes a small issue in mythril/laser/smt/__init__.py where z3.Bool() was called rather than z3.BoolVal()

You should be able to check the new types by running this from the root of the repo:

MYPYPATH=mypy-stubs/ mypy --disallow-untyped-defs --disallow-any-unimported mythril/laser/smt/*.py


mythril/laser/smt/bitvec.py
```diff
@@ -61,7 +61,7 @@ def __add__(self, other) -> "BitVec":
         union = self.annotations + other.annotations
         return BitVec(self.raw + other.raw, annotations=union)
 
-    def __sub__(self, other) -> "BitVec":
+    def __sub__(self, other: Union[int, "BitVec"]) -> "BitVec":
```
norhh: `other` can't be `int`. If it is int somewhere, in that case we need to correct it :sweat_smile:

mythril/laser/smt/solver.py
```diff
@@ -1,15 +1,16 @@
...
 
 
 class Solver:
     """An SMT solver object."""
 
-    def __init__(self):
+    def __init__(self) -> None:
         """"""
-        self.raw = z3.Solver()
+        self.raw = z3.Solver()  # type: Union[z3.Solver, z3.Optimize]
```
norhh: Any reason why this should also have z3.Optimize?

mythril/laser/smt/bitvec.py
```diff
@@ -92,7 +92,7 @@ def __truediv__(self, other) -> "BitVec":
         union = self.annotations + other.annotations
         return BitVec(self.raw / other.raw, annotations=union)
 
-    def __and__(self, other) -> "BitVec":
+    def __and__(self, other: Union[int, "BitVec"]) -> "BitVec":
         """Create an and expression.
```
norhh: same here, i think other shouldn't be an int

mythril/laser/smt/__init__.py
```diff
@@ -65,24 +75,30 @@ class _SmtSymbolFactory(SymbolFactory):
...
+    ) -> Union[BitVec, z3.BitVecRef]:
         """Creates a new bit vector with a concrete value."""
         raw = z3.BitVecVal(value, size)
         return BitVec(raw, annotations)
 
     @staticmethod
-    def BitVecSym(name: str, size: int, annotations=None):
+    def BitVecSym(
+        name: str, size: int, annotations: Annotations = None
+    ) -> Union[BitVec, z3.BitVecRef]:
```
norhh: this doesn't return z3.BitVecRef right?

mythril/laser/smt/__init__.py
```diff
@@ -65,24 +75,30 @@ class _SmtSymbolFactory(SymbolFactory):
...
         """
-        raw = z3.Bool(value)
+        raw = z3.BoolVal(value)
         return Bool(raw, annotations)
 
     @staticmethod
-    def BitVecVal(value: int, size: int, annotations=None):
+    def BitVecVal(
+        value: int, size: int, annotations: Annotations = None
+    ) -> Union[BitVec, z3.BitVecRef]:
```
norhh: this too doesn't return z3.BitVecRef right?

mythril/laser/smt/solver.py
cd1m0: The issue is that the (mythril) Solver class is a super class of the (mythril) Optimize class, and they both define self.raw. Unfortunately, in z3 z3.Optimize is not a subclass of z3.Solver.

So the only way I could think of to unify this was with a union.

Something else that may work could be something along these lines:

T=TypeVar("T", bound=Union[z3.Solver, z3.Optimize])
class BaseSolver(Generic[T]):
    raw:  T

class Solver(BaseSolver[z3.Solver]):
    ...

class Optimize(BaseSolver[z3.Optimize]):
   ...

But then we would have to update the type of *all* places where Solver is expected to BaseSolver

mythril/laser/smt/__init__.py
cd1m0: The BitVecSym function is defined in the parent SymbolFactory class. In the _SmtSymbolFactory subclass BitVecSym returns a BitVec and in the _Z3SymbolFactory subclass BitVecSym returns a z3.BitVecRef.

However, the signature of BitVecSym in the base class must be compatible with all of the implementing classes, ergo why I chose to make it a union.

However now that you point it out, we CAN be more precise here! We can do the following:

```
T=TypeVar("T", bound=Union[BitVec, z3.BitVecRef])
class SymbolFactory(Generic[T]):
    def BitVecSym(..)->T:
...
class _SmtSymbolFactory(SymbolFactory[BitVec]):
    ...
class _Z3SymbolFactory(SymbolFactory[z3.BitVecRef]):
    ...
```

That may make this simpler. Do you want me to try that instead?

mythril/laser/smt/__init__.py
cd1m0: See above comment for reply.

mythril/laser/smt/__init__.py
```diff
@@ -19,14 +19,20 @@
...
+Annotations = Optional[List[Any]]
+
+
 class SymbolFactory:
     """A symbol factory provides a default interface for all the components of mythril to create symbols"""
 
     @staticmethod
-    def Bool(value: bool, annotations=None):
+    def Bool(
+        value: "__builtins__.bool", annotations: Annotations = None
```
norhh: are bool and __builtins__.bool  different? as we aren't importing bool, is there any other problem here?

mythril/laser/smt/bitvec.py
```diff
@@ -139,44 +139,48 @@ def __gt__(self, other) -> Bool:
         union = self.annotations + other.annotations
         return Bool(self.raw > other.raw, annotations=union)
 
-    def __eq__(self, other) -> Bool:
+    #MYPY: fix complains about overriding __eq__
```
norhh: Usually there would be a space after # in comments from PEP8 standards.

mythril/laser/smt/bitvec.py
```diff
@@ -139,44 +139,48 @@ def __gt__(self, other) -> Bool:
...
-            return Bool(self.raw == other, annotations=self.annotations)
+            return Bool(cast(z3.BoolRef, self.raw == other), annotations=self.annotations)
 
         union = self.annotations + other.annotations
-        return Bool(self.raw == other.raw, annotations=union)
+        # MYPY: fix complaints due to z3 overriding __eq__
+        return Bool(cast(z3.BoolRef, self.raw == other.raw), annotations=union)
 
-    def __ne__(self, other) -> Bool:
+    #MYPY: fix complains about overriding __ne__
```
norhh: There should be a space after #, according to PEP8 standards

mythril/laser/smt/bool.py
```diff
@@ -46,27 +46,31 @@ def value(self) -> Union[bool, None]:
         else:
             return None
 
-    def __eq__(self, other):
+    #MYPY: complains about overloading __eq__ # noqa
```
norhh: Same here, no extra space after #.

mythril/laser/smt/bool.py
```diff
@@ -46,27 +46,31 @@ def value(self) -> Union[bool, None]:
...
         """
         if isinstance(other, Expression):
-            return Bool(self.raw == other.raw, self.annotations + other.annotations)
-        return Bool(self.raw == other, self.annotations)
+            return Bool(cast(z3.BoolRef, self.raw == other.raw),
+                        self.annotations + other.annotations)
+        return Bool(cast(z3.BoolRef, self.raw == other), self.annotations)
 
-    def __ne__(self, other):
+    #MYPY: complains about overloading __ne__ # noqa
```
norhh: Space after # for the sake of PEP8.

mythril/laser/smt/solver.py
norhh: Yes, that would be much cleaner.

mythril/laser/smt/__init__.py
cd1m0: In the context of init, for some !&@# reason mypy was confusing bool with myhril.laser.smt.bool. The whole thing is very strange, as we are _not_ importing mythril.laser.smt.bool, but just importing things from it. I suppose inside __init__.py the names of .py files in the current directory are present? I am not sure. I spent half an hour trying to figure out mypy thought bool refered to mythril.laser.smt.bool, and in the end gave up and disambiguated using __builin__.bool

mythril/laser/smt/solver.py
cd1m0: So this is a bit tricky. As we have to by Python 3.5 compliant, I cannot use class variable type annotations like this:

```
class BaseSolver(Generic[T]):
      raw: T
```

Instead I need to add a dummy init and annotate:

```
class BaseSolver(Generic[T]):
    def __init__(self):
        self.raw = ?  # type: T
```

Now the issue becomes what the ? is. If i put None, then I must make the type Optional[T], which is incorrect and adds its own problems. An extremly ugly hack that works is:

```
class BaseSolver(Generic[T]):
    def __init__(self):
        self.raw = self.raw  # type: T
```
Ideas?

cd1m0: Never mind, the answer is obvious..

mythril/laser/smt/bool.py
cd1m0: Fixed in next commit

mythril/laser/smt/bool.py
cd1m0: Fixed in next commit

mythril/laser/smt/bitvec.py
cd1m0: Fixed in next commit

mythril/laser/smt/bitvec.py
cd1m0: Fixed in next commit

mythril/laser/smt/__init__.py
cd1m0: For some reason mypy was confused in the context of __init__.py and thought 'bool' referes to the mythril.laser.smt.bool module, rather than the builtin type. I used __builtins__.bool here to disambiguate.

mythril/laser/smt/solver.py
cd1m0: I fixed it in the next commit, adding a BaseSolver class from which both Solver and Optimize refer. Please note that I also chagned the signatures of the BaseSolver.{add,append} methods (perviously Solver.{add, append}) to make them tighter.

mythril/laser/smt/__init__.py
norhh: yes, that would be great.

mythril/laser/smt/bitvec.py
```diff
@@ -130,7 +130,7 @@ def __lt__(self, other) -> Bool:
         union = self.annotations + other.annotations
         return Bool(self.raw < other.raw, annotations=union)
 
-    def __gt__(self, other) -> Bool:
+    def __gt__(self, other: "BitVec") -> Bool:
```
norhh: It can also support an int here right?

mythril/laser/smt/bitvec.py
```diff
@@ -121,7 +121,7 @@ def __xor__(self, other) -> "BitVec":
         union = self.annotations + other.annotations
         return BitVec(self.raw ^ other.raw, annotations=union)
 
-    def __lt__(self, other) -> Bool:
+    def __lt__(self, other: "BitVec") -> Bool:
```
norhh: same here

mythril/laser/smt/bitvec.py
```diff
@@ -112,7 +112,7 @@ def __or__(self, other) -> "BitVec":
         union = self.annotations + other.annotations
         return BitVec(self.raw | other.raw, annotations=union)
 
-    def __xor__(self, other) -> "BitVec":
+    def __xor__(self, other: "BitVec") -> "BitVec":
```
norhh: It can also support an int here right?

cd1m0:
Not based on the way the code is written:

    def __gt__(self, other: "BitVec") -> Bool:
        """Create a signed greater than expression.

        :param other:
        :return:
        """
        union = self.annotations + other.annotations
        return Bool(self.raw > other.raw, annotations=union)

An int doesn't have an annotations member.

cd1m0:
Not based on the way the code is written:

    def __lt__(self, other: "BitVec") -> Bool:
        """Create a signed less than expression.

        :param other:
        :return:
        """
        union = self.annotations + other.annotations
        return Bool(self.raw < other.raw, annotations=union)

An int doesn't have an annotations member.

cd1m0:
Not based on the way the code is written:

    def __xor__(self, other: "BitVec") -> "BitVec":
        """Create a xor expression.

        :param other:
        :return:
        """
        union = self.annotations + other.annotations
        return BitVec(self.raw ^ other.raw, annotations=union)

ints don't have annotations

PR #890, opened 2019-01-17
Create module for SWC-124: Arbitrary Write

### LINKED ISSUE #861: Analysis Module: SWC-124 - Write to Arbitrary Storage Location
Create an analysis module for SWC-124 - Write to Arbitrary Storage Location. SWC link: <https://smartcontractsecurity.github.io/SWC-registry/docs/SWC-124>

This can be done by adding pre-hooks at `sstore` instructions and checking if it is possible to constrain the location argument to an arbitrary value.

nbanmp:
closes #861 

After putting this module together, I realized that mythril adds unconstrained bitvecs to the stack fairly often when we don't know the correct value to add to the stack is. This might result in way too many false positives, and as such, we might not want to merge it.

### DIFF
--- a/mythril/analysis/modules/arbitrary_write.py
+++ b/mythril/analysis/modules/arbitrary_write.py
@@ -1,0 +1,93 @@
+from mythril.analysis import solver
+from mythril.analysis.report import Issue
+from mythril.analysis.swc_data import WRITE_TO_ARBITRARY_STORAGE
+from mythril.exceptions import UnsatError
+from mythril.analysis.modules.base import DetectionModule
+from mythril.laser.ethereum.state.global_state import GlobalState
+import logging
+import json
+
+log = logging.getLogger(__name__)
+
+DESCRIPTION = """
+Check if an arbitrary storage location can be written to by anyone.
+"""
+
+
+class ArbitraryWriteModule(DetectionModule):
+    """This module checks if an arbitrary storage location in the contact can be written to by anyone."""
+
+    def __init__(self):
+        super().__init__(
+            name="Write to Arbitrary Storage",
+            swc_id=WRITE_TO_ARBITRARY_STORAGE,
+            description=DESCRIPTION,
+            entrypoint="callback",
+            pre_hooks=["SSTORE"],
+        )
+        self._cache_address = {}
+
+    def reset_module(self):
+        """
+        Resets the module
+        :return:
+        """
+        super().reset_module()
+        self._cache_address = {}
+
+    def execute(self, state: GlobalState):
+        """
+
+        :param state:
+        :return:
+        """
+        self._issues.extend(self._analyze_state(state))
+        return self.issues
+
+    def _analyze_state(self, state):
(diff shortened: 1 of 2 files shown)


mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,93 @@
...
+    """This module checks if an arbitrary storage location in the contact can be written to by anyone."""
+
+    def __init__(self):
+        super().__init__(
+            name="Write to Arbitrary Storage",
+            swc_id=WRITE_TO_ARBITRARY_STORAGE,
+            description=DESCRIPTION,
+            entrypoint="callback",
+            pre_hooks=["SSTORE"],
+        )
```
norhh: Can you add a `-> None`  in this part

mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,93 @@
...
+        self._cache_address = {}
+
+    def reset_module(self):
```
norhh: this returns None. `-> None`

mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+        Resets the module
+        :return:
+        """
+        super().reset_module()
+        self._cache_address = {}
+
+    def execute(self, state: GlobalState):
```
norhh: Returns a list of issues

mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,93 @@
...
+        """
+
+        :param state:
+        :return:
+        """
+        self._issues.extend(self._analyze_state(state))
+        return self.issues
+
+    def _analyze_state(self, state):
```
norhh: returns a list of issues.

JoranHonig:
@nbanmp Does this module perhaps give a ton of false positives when there is a mapping?

nbanmp:
@JoranHoing From my relatively small amount of testing so far, yes, it looks like the issues are often returned when there is a mapping.

JoranHonig:
:upside_down_face: I think this pr will have to wait until we have made a neat implementation for storage & uninterpreted functions for hashes

mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,93 @@
...
+
+    def __init__(self):
+        super().__init__(
+            name="Write to Arbitrary Storage",
+            swc_id=WRITE_TO_ARBITRARY_STORAGE,
+            description=DESCRIPTION,
+            entrypoint="callback",
+            pre_hooks=["SSTORE"],
+        )
+        self._cache_address = {}
```
norhh: This should also be typehinted as `Dict[int, bool]` or something like that.

nbanmp:
Closing until storage implementation is refactored.

PR #897, opened 2019-01-24
Reachability Check

### LINKED ISSUE #509: Do reachability check on branch in laser svm
If we have some kind of branch in the cfg we should prune unreachable branches.
Right now we only prune branches where the condition simplifies to True

norhh:
fixes #509, Also cleans up the some part of `jumpi_`  instruction for readability


mythril/laser/ethereum/state/constraints.py
```diff
@@ -1,77 +1,98 @@
...
+    def pop(self, index: int = -1) -> None:
         """
 
-        :param index:
+        :param index: Index to be popped from the list
         """
         raise NotImplementedError
 
-    def __copy__(self):
+    def get_list(self) -> List[Bool]:
```
JoranHonig: How about
```suggestion
    @property
    def list(self) -> List[Bool]:
```

mythril/laser/ethereum/state/constraints.py
```diff
@@ -1,77 +1,98 @@
...
-        :param solver:
-        :param possibility:
+        :param constraint_list: List of constraints
+        :param is_possible: Whether it is possible to satisfy the constraints or not
         """
         super(Constraints, self).__init__(constraint_list or [])
-        self.solver = solver
-        self.__possibility = possibility
+        self.__default_timeout = 10
+        self.__is_possible = is_possible
```
JoranHonig: Why are these variables defined with double underscore?

mythril/laser/ethereum/state/constraints.py
```diff
@@ -1,77 +1,98 @@
...
+        :return: True/False based on the existence of solution of constraints
         """
-        return True
+        if self.__is_possible is not None:
+            return self.__is_possible
+        solver = Solver()
+        solver.set_timeout(self.__default_timeout)
+        for constraint in self[:]:
+            solver.add(constraint)
+        self.__is_possible = solver.check() != unsat
```
JoranHonig:
```suggestion
        self.__is_possible = solver.check() == sat
```
Unknown should also be categorized as not possible I think

mythril/laser/ethereum/state/constraints.py
norhh: They are super secret variables which can't even be acccesed by obj.__<name>

mythril/laser/ethereum/state/constraints.py
norhh: for low timeouts such as 10ms there might be a case where it might not be solvable right, but solvable at the time of 100s.

mythril/laser/ethereum/state/constraints.py
JoranHonig: ^^, isn't that already the case if you do _
As I understand it, this would only make sure that classes inheriting this class would also not be able to access the variable

mythril/laser/ethereum/state/constraints.py
norhh: But As the constraints seems to increase lot of `unknown` answers seems to popup due to the timeout(after increasing timeout to 100ms) so we might just miss those paths right?

JoranHonig: Maybe we need to put a higher timeout there

norhh: increasing it to 100ms itself slows mythril by 2X , so should we increase the timeout?

JoranHonig:
@norhh I think the get_model function should maybe be adapted with an additional parameter that if true makes get_model ignore the time_handler

mythril/laser/ethereum/state/constraints.py
```diff
@@ -1,77 +1,99 @@
...
+        :param is_possible: Whether it is possible to satisfy the constraints or not
         """
         super(Constraints, self).__init__(constraint_list or [])
-        self.solver = solver
-        self.__possibility = possibility
+        self._default_timeout = 100
+        self._is_possible = is_possible
 
-    def check_possibility(self):
+    def check_possibility(self) -> bool:
```
nbanmp: Thoughts about making adding `@property` and renaming to `is_possible`? Would that make this a little clearer?

norhh: That would make it pretty clear 👍

PR #900, opened 2019-01-25
Change solver's timeout based on execution-timeout

### LINKED ISSUE #899: Latest mythril-dev docker image fails on analysis
## Description
Latest `mythril/myth-dev` docker image [fails on analysis](https://circleci.com/workflow-run/5a531854-ae08-481b-9595-277d33dec6ef) with 
```
  File "/usr/local/lib/python3.6/dist-packages/mythril-0.19.11-py3.6.egg/mythril/laser/ethereum/svm.py", line 283, in execute_state
```

## How to Reproduce
The failure occurs on the latest `mythril-dev` docker image so
```
docker pull mythril/myth-dev
docker run -it -v "$PWD"/build:/build mythril/myth-dev myth --truffle
```
The sample CI builds used here run against the [ColonyNetwork](https://github.com/JoinColony/colonyNetwork) repo.

## Expected behavior
`mythril-dev` image was analysing our project correctly 4 days ago as seen in [this build](https://circleci.com/gh/JoinColony/colonyNetwork/5945)

## Environment
- Mythril version: v0.19.11
- Solidity compiler and version: 0.4.23
- Python version: 2.7.15
- OS and Version: Mac OS High Sierra

norhh:
Also fixes #899 , was required to fix to test for various other contracts.

### DIFF
--- a/mythril/laser/ethereum/time_handler.py
+++ b/mythril/laser/ethereum/time_handler.py
@@ -1,7 +1,8 @@
 import time
+from mythril.support.support_utils import Singleton
 
 
-class TimeHandler:
+class TimeHandler(object, metaclass=Singleton):
     def __init__(self):
         self._start_time = None
         self._execution_time = None
--- a/mythril/laser/ethereum/time_handler.py
+++ b/mythril/laser/ethereum/time_handler.py
@@ -1,0 +1,17 @@
+import time
+
+
+class TimeHandler:
+    def __init__(self):
+        self._start_time = None
+        self._execution_time = None
+
+    def start_execution(self, execution_time):
+        self._start_time = int(time.time() * 1000)
+        self._execution_time = execution_time * 1000
+
+    def time_remaining(self):
+        return self._execution_time - (int(time.time() * 1000) - self._start_time)
+
+
+time_handler = TimeHandler()
--- a/mythril/laser/smt/bool.py
+++ b/mythril/laser/smt/bool.py
@@ -84,6 +84,7 @@ class Bool(Expression[z3.BoolRef]):
 def And(*args: Bool) -> Bool:
     """Create an And expression."""
     union = []
+    args = [arg if isinstance(arg, Bool) else Bool(arg) for arg in args]
     for arg in args:
         union.append(arg.annotations)
     return Bool(z3.And([a.raw for a in args]), union)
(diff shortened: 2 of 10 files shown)


mythril/laser/smt/bool.py
```diff
@@ -84,6 +84,7 @@ def __bool__(self) -> bool:
 def And(*args: Bool) -> Bool:
     """Create an And expression."""
     union = []
+    args = [arg if isinstance(arg, Bool) else Bool(arg) for arg in args]
```
norhh: This is against the corresponding type hint of `Bool.And()`, will change it in #894 .

mythril/laser/ethereum/time_handler.py
```diff
@@ -0,0 +1,17 @@
...
+
+    def start_execution(self, execution_time):
+        self._start_time = int(time.time() * 1000)
+        self._execution_time = execution_time * 1000
+
+    def time_remaining(self):
+        return self._execution_time - (int(time.time() * 1000) - self._start_time)
+
+
+time_handler = TimeHandler()
```
JoranHonig: What do you think about using the singleton meta class here? 

```

class Singleton(type):
    """A metaclass type implementing the singleton pattern."""

    _instances = {}

    @synchronized(lock)
    def __call__(cls, *args, **kwargs):
        """Delegate the call to an existing resource or a a new one.

        This is not thread- or process-safe by default. It must be protected with
        a lock.

        :param args:
        :param kwargs:
        :return:
        """
        if cls not in cls._instances:
            cls._instances[cls] = super(Singleton, cls).__call__(*args, **kwargs)
        return cls._instances[cls]
```

JoranHonig: Its currently in the signatures file, but it can/should be moved to some util's class if we use it here as well

norhh: We don't need a lock for this module right?I have added a singleton class without any locks in the support_utils module. And also the signature's singleton uses a lock generated in that file, I don't have much idea about sending a lock to the singleton class. what do you think about looking at it in a different PR.

JoranHonig: Oh, I hadn't payed attention to the lock. `what do you think about looking at it in a different PR.` Yeah that's fine

PR #901, opened 2019-01-27
Implement BitVecFunc for sha3

nbanmp:
First step in storage refactor.

Lots of edge cases for interactions between BitVecs and BitVecFuncs were introduced. :(

Immediately, this means that that contracts with stuff like this work correctly:
```
if (keccak256(abi.encode(input)) == keccak256(abi.encode(known_storage_var)))
if (keccak256(abi.encode(input)) == keccak256(abi.encode(0x01001337)))
if (keccak256(abi.encode(0x0)) == 0x290decd9548b62a8d60345a988386fc84ba6bc95484008f6362f93160ef3e563)
```


mythril/laser/smt/bitvec.py
```diff
@@ -160,14 +181,282 @@ def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
...
+            return Bool(z3.BoolVal(True), annotations=union)
+
+        # MYPY: fix complaints due to z3 overriding __eq__
+        return And(
+            Bool(cast(z3.BoolRef, self.raw == other.raw), annotations=union),
+            self.input == other.input,
+        )
+
+    # MYPY: fix complains about overriding __ne__
+    def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
```
norhh: Adding other operators like >=, <= might also be good, mypy caught some problems with the Bitvec class regarding usage of >= and <= in some part of code.It might also happen for this class

mythril/laser/smt/bitvec.py
```diff
@@ -233,7 +572,6 @@ def Concat(*args: Union[BitVec, List[BitVec]]) -> BitVec:
     :param args:
     :return:
     """
-
     # The following statement is used if a list is provided as an argument to concat
     if len(args) == 1 and isinstance(args[0], list):
         bvs = args[0] # type: List[BitVec]
```
norhh: This variable isn't just a List of BitVecs right? It also contains bitvecfunc

mythril/laser/smt/bitvec.py
```diff
@@ -242,8 +580,16 @@ def Concat(*args: Union[BitVec, List[BitVec]]) -> BitVec:
...
     annotations = [] # type: Annotations
+    bitvecfunc = False
     for bv in bvs:
         annotations += bv.annotations
+        if isinstance(bv, BitVecFunc):
+            bitvecfunc = True
+
+    if bitvecfunc:
+        # Is there a better value to set name and input to in this case?
+        return BitVecFunc(raw=nraw, name=None, input=None, annotations=bv.annotations)
```
norhh: I don't have a clear idea of this part but `annotations=bv.annotations` doesn't make when bv is being used after the termination of the for loop in line 584.

mythril/laser/smt/bitvec.py
nbanmp: A BitVecFunc is a subclass of a BitVec.

mythril/laser/smt/bitvec.py
nbanmp: I do not believe `functools.total_ordering()` will work because we are not returning a `bool`, but a custom class. I'll implement `<=` and `>=` though.

mythril/laser/smt/bitvec.py
nbanmp: You're right, good catch.

mythril/laser/smt/bitvec.py
```diff
@@ -176,6 +522,12 @@ def If(a: Union[Bool, bool], b: Union[BitVec, int], c: Union[BitVec, int]) -> Bi
     :param c:
     :return:
     """
+    # TODO: Handle BitVecFunc
+    if isinstance(b, BitVecFunc):
+        die()
```
norhh: What is this function? I didn't see it's definition anywhere

mythril/laser/smt/bitvec.py
```diff
@@ -242,8 +641,16 @@ def Concat(*args: Union[BitVec, List[BitVec]]) -> BitVec:
...
     annotations = [] # type: Annotations
+    bitvecfunc = False
     for bv in bvs:
         annotations += bv.annotations
+        if isinstance(bv, BitVecFunc):
+            bitvecfunc = True
+
+    if bitvecfunc:
+        # Is there a better value to set name and input to in this case?
+        return BitVecFunc(raw=nraw, name=None, input=None, annotations=annotations)
```
norhh: if name=None is possible, in that case type of name should be Optional[str](in the \_\_init\_\_() )

mythril/laser/smt/bitvec.py
```diff
@@ -160,14 +181,339 @@ def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
...
+    def __init__(
+        self,
+        raw: z3.BitVecRef,
+        name: str,
+        input: Union[int, "BitVec"] = None,
+        annotations: Optional[Annotations] = None,
+    ):
+        """
+
+        :param raw:
```
norhh: Param: name documentation is missing here

norhh: And also having it named func_name might be better, WDYT?

mythril/laser/smt/bitvec.py
nbanmp: whoops. XD

norhh: Using an assert might be good in such circumstances

mythril/laser/smt/bitvec.py
```diff
@@ -160,14 +181,340 @@ def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
...
+            return Bool(
+                cast(z3.BoolRef, self.raw != other), annotations=self.annotations
+            )
 
         union = self.annotations + other.annotations
         # MYPY: fix complaints due to z3 overriding __eq__
         return Bool(cast(z3.BoolRef, self.raw != other.raw), annotations=union)
 
 
+class BitVecFunc(BitVec):
```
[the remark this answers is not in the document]
JoranHonig: Looking at the size of this file I second this :smile:

mythril/laser/smt/bitvec.py
```diff
@@ -98,8 +105,10 @@ def __and__(self, other: Union[int, "BitVec"]) -> "BitVec":
         :param other:
         :return:
         """
+        if isinstance(other, BitVecFunc):
+            return other & self
         if not isinstance(other, BitVec):
-            other = BitVec(z3.BitVecVal(other, 256))
+            other = BitVec(z3.BitVecVal(other, self.size()))
```
JoranHonig: good catch

mythril/laser/smt/bitvec.py
```diff
@@ -160,14 +181,340 @@ def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
...
+        """
+
+        :param raw: The raw bit vector symbol
+        :param func_name: The function name. e.g. sha3
+        :param input: The input to the functions
+        :param annotations: The annotations the BitVecFunc should start with
+        """
+
+        self.func_name = func_name
+        self.input = input
```
JoranHonig: input seems to be a taken keyword, maybe we should change it to `input_`

mythril/laser/smt/bitvec.py
```diff
@@ -160,14 +181,340 @@ def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
...
+        super().__init__(raw, annotations)
+
+    def __add__(self, other: Union[int, "BitVec"]) -> "BitVecFunc":
+        """Create an addition expression.
+
+        :param other:
+        :return:
+        """
+        if not isinstance(other, BitVec):
```
JoranHonig:
```suggestion
        if isinstance(other, int):
```

JoranHonig: I'd do the same for all the others functions. Although they should be functionally equivalent, this way it is more easily visible that we're always dealing with an int if the condition is true

mythril/laser/smt/bitvec.py
```diff
@@ -160,14 +181,340 @@ def __ne__(self, other: Union[int, "BitVec"]) -> Bool:  # type: ignore
...
+        raw = self.raw + other.raw
+        union = self.annotations + other.annotations
+
+        if isinstance(other, BitVecFunc):
+            # TODO: Find better value to set input and name to in this case
+            return BitVecFunc(raw=raw, func_name=None, input=None, annotations=union)
+
+        return BitVecFunc(raw=raw, func_name=self.func_name, input=self.input, annotations=union)
+
+    def __sub__(self, other: Union[int, "BitVec"]) -> "BitVecFunc":
```
JoranHonig: It seems to me that a lot of the logic for each of these implementations is equivalent, putting up a helper function that deals with everything and only passing it the operator could make this pr a bit shorter :+1:

mythril/laser/smt/bitvec.py
```diff
@@ -213,17 +240,49 @@ def UGT(a: BitVec, b: BitVec) -> Bool:
     :return:
     """
     annotations = a.annotations + b.annotations
+    if isinstance(a, BitVecFunc):
```
norhh: A similar thing can be done here, like the BitVecFun right? like having an comparator and operator helper

PR #902, opened 2019-01-28
Independence optimization

### LINKED ISSUE #232: Independence optimization
## Description
Add independence-optimization for constraint solving. 
When a state with constraint set C needs to add a new constraint c’, instead of passing C ∧ c’ to the constraint solver to test satisfiability, pass R(C, c’) ∧ c’ where the expression R(C, c’)  returns only the constraints in C that have symbolic variables also in c’.

### Benefits
This optimization reduces the difficulty of the query and also increases constraint cache hits.

JoranHonig:
closes #232


tests/laser/smt/independece_solver_test.py
```diff
@@ -0,0 +1,143 @@
...
+    # Act
+    for condition in conditions:
+        dependence_map.add_condition(condition)
+
+    # Assert
+    assert 2 == len(dependence_map.buckets)
+
+    assert x in dependence_map.buckets[0].variables
+    assert y in dependence_map.buckets[0].variables
+    assert z in dependence_map.buckets[0].variables
```
nbanmp: Should this test also assert that x, y, and z are not in `dependence_map.buckets[1].variables`, and other similar negatives?

mythril/laser/smt/independence_solver.py
```diff
@@ -0,0 +1,147 @@
...
+
+    def set_timeout(self, timeout: int) -> None:
+        """Sets the timeout that will be used by this solver, timeout is in
+        milliseconds.
+
+        :param timeout:
+        """
+        self.raw.set(timeout=timeout)
+
+    def add(self, constraints: list) -> None:
```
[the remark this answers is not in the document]
nbanmp: Actually, it looks like type hints for this are wrong in `smt.solver.Solver`, and a single constraint is supported there.

mythril/laser/smt/independence_solver.py
```diff
@@ -0,0 +1,147 @@
...
+
+        if relevant_buckets:
+            # Merge buckets, and rewrite variable map accordingly
+            relevant_buckets.add(new_bucket)
+            new_bucket = self.merge_buckets(relevant_buckets)
+
+        for variable in variables:
+            self.variable_map[str(variable)] = new_bucket
+
+    def merge_buckets(self, bucket_list: Set[DependenceBucket]):
```
norhh: It might be good for this function to be hidden right?

norhh: And also a set named as bucket_list might be a bit misleading

mythril/laser/smt/model.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class Model:
+    """ The model class wraps a z3 model
+
+    This implementation allows for multiple internal models, this is required for the use of an independence solver
+    which has models for multiple queries which need an uniform output.
+    """
+
+    def __init__(self, models=None):
```
norhh: Can you also typehint the models variable here?

mythril/laser/smt/model.py
```diff
@@ -0,0 +1,54 @@
...
+                result = internal_model[item]
+                if result is not None:
+                    return result
+            except IndexError:
+                if is_last_model:
+                    raise
+                continue
+        return None
+
+    def eval(self, expression: z3.ExprRef, model_completion: bool = False):
```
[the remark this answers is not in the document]
JoranHonig: Yes, but I don't want to do too much in one pr. The goal for Model was to provide the same interface that z3.Model does

mythril/laser/smt/independence_solver.py
```diff
@@ -0,0 +1,147 @@
...
+        dependence_map = DependenceMap()
+        for constraint in self.constraints:
+            dependence_map.add_condition(constraint)
+
+        self.models = []
+        for bucket in dependence_map.buckets:
+            self.raw.reset()
+            self.raw.append(*bucket.conditions)
+            check_result = self.raw.check()
+            if check_result == z3.unsat:
```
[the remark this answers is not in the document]
JoranHonig: good catch

mythril/laser/smt/independence_solver.py
JoranHonig: Well it was a list at first, and I liked the name so much I didn't change it

JoranHonig: Is that argument convincing enough to resolve this conversation :innocent: ?

norhh: and about making the `merge_buckets()` hidden? Would it be intended to be called from outside at any time?It might more readble if it's hidden in that case right?

JoranHonig: I think I already did that

PR #907, opened 2019-02-04
Effect check

### LINKED ISSUE #782: Don't report issues when transaction always reverts
## Description

Currently, a number of false positives exist due to the fact that we may report conditions that always result in a revert. For example `SafeMath` uses a "Effects -> Check pattern" where state is reverted after an integer overflow occurs. An example of this can be for [here](https://gist.github.com/mickayz/4d58ed15edc60d301a4f60fe193e1212).

## Requested Feature

With the callback pattern one solution could be to initialize newly discovered issues in an "unconfirmed" issues and change status to "confirmed" once a `STOP` or `RETURN` is reached in a subsequent state. This would require maintaining a list of unconfirmed issues and an additional callback module on `STOP` and `RETURN`.

Exact implementation unknown, I better leave that up to you ;)

JoranHonig:
This pr will add a bit of false positive filtering to the integer over&underflow module.
It will only report a vulnerability if:
- there is an overflow somewhere
- used by an sstore or STOP
- for which there is a trace that succesfully finishes
fixes #782

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -8,6 +8,7 @@ from mythril.analysis.report import Issue
 from mythril.analysis.swc_data import INTEGER_OVERFLOW_AND_UNDERFLOW
 from mythril.exceptions import UnsatError
 from mythril.laser.ethereum.state.global_state import GlobalState
+from mythril.laser.ethereum.state.annotation import StateAnnotation
 from mythril.analysis.modules.base import DetectionModule
 
 from mythril.laser.smt import (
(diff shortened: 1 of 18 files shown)


mythril/analysis/modules/integer.py
```diff
@@ -27,6 +28,19 @@
...
+    def __init__(
+        self, overflowing_state: GlobalState, operator: str, constraint
+    ) -> None:
+        self.overflowing_state = overflowing_state
+        self.operator = operator
+        self.constraint = constraint
+
+
+class OverUnderflowStateAnnotation(StateAnnotation):
+    """ State Annotation used if an overflow both possible and used in the annotated path"""
```
norhh: an `is` is missing here right? "State Annotation used if an overflow is both possible and used in the annotated path"

```suggestion
    """ State Annotation used if an overflow is both possible and used in the annotated path"""
```

PR #910, opened 2019-02-06
Solver statistics

JoranHonig:
This pr implements the logic to measure query count and solving time.

Additionally two minor bugs were fixed


solidity_examples/BECToken.sol
```diff
@@ -296,4 +295,4 @@ contract BecToken is PausableToken {
         //if ether is sent to this address, send it back.
         revert();
     }
-}
\ No newline at end of file
+}
```
nbanmp: Good fix. :smile:

mythril/laser/smt/solver/solver_statistics.py
```diff
@@ -0,0 +1,37 @@
+from time import time
+
+from mythril.support.support_utils import Singleton
+
+from typing import Callable
+
+
+def stat_smt_query(func: Callable):
```
nbanmp: Docstrings for functions / classes in this file? :smiley:

nbanmp:
It might be nice to have even more fine grained reporting of solver time than the total. Perhaps we can have a list of solver times, and to append to that list each time the solver is called. WDYT?

JoranHonig:
@nbanmp definately, depending on what we want we can extend the solver statistics & decorator to keep track of anything.

PR #918, opened 2019-02-10
Enable analysis at specific blocknumber for dynamic loader

Joool:
When running an archive node, one might be interested in analysing contracts at a specific block when using the dynamic loader.

This pull request provides such functionality, adding a command line option and additionally padding requests to even length conforming to [paritys API](https://wiki.parity.io/JSONRPC#types-in-the-jsonrpc) (go ethereum accepts both).


Joool:
I will revisit and afterwards reopen this.

PR #919, opened 2019-02-11
Detect State changes after external calls

### LINKED ISSUE #886: Re-introduce check for state changes after external calls
## Description

Report state changes after external calls. 

## Implementation details

Prior versions of Mythril used to report a warning if a state change after an external call occurred:

```
"The contract account state is changed after an external call. Consider that the called contract could re-enter the function before this state change takes place. This can lead to business logic vulnerabilities."
```

Re-implement this check using state annotations.

## Implementation

Off the top of my head: In the `external_calls` module, Set an annotation on calls and check for the annotation on SSTORE and balance changes.

It could make sense to change severity based on whether the target address can be influenced by the caller (this is already implemented in `external calls.py`). E.g.:

- State change after call to fixed address: Low
- State change after call to user-determined address: Medium

## Link

https://smartcontractsecurity.github.io/SWC-registry/docs/SWC-107#simple-daosol

https://medium.com/chainsecurity/constantinople-enables-new-reentrancy-attack-ace4088297d9

norhh:
fixes #886


JoranHonig:
Could you move the state change after external call to an external module? 
It shouldn't add much overhead, and is better in terms of understandability and maintainability IMO

nbanmp:
Does this mean that we will have duplicate issues, since for every report that the "state change external call" module makes, the "external call" module will also report something?

nbanmp:
They don't report the exact same issues. A "Low" severity issue will be raised for every call, and a "Medium" severity issue is raised for a call to an arbitrary address in the external call module. They should both be made "Low" severity if we are going to keep the modules separate. There is no possible reentrancy vulnerability without a state change, right?

norhh:
`They should both be made "Low" severity if we are going to keep the modules separate` if the user can make a call to an aribtrary address then it is a bit of a problem right?(probably thats the reason for "Medium" priority). `There is no possible reentrancy vulnerability without a state change, right?`, yeah they should not be considered as reentrancy issues.(I can't find any other relavent swc id for them)

nbanmp:
The "CREATE" opcode could be added, as reentrancy is possible in the constructor of a newly created contract.

norhh:
`The "CREATE" opcode could be added, as reentrancy is possible in the constructor of a newly created contract.` Can you mention how CREATE can cause a reentrancy. I don't have a clear idea of the problem.

nbanmp:
If a contract deploys bytecode with `create`, the constructor is executed. The constructor can call contracts that reenter or reenter itself. 

https://www.ndss-symposium.org/wp-content/uploads/2019/02/ndss2019_09-3_Rodler_paper.pdf mentions it a bit.

mythril/analysis/modules/external_calls.py
```diff
@@ -22,6 +27,22 @@
 """
 
 
+class CallIssue:
```
JoranHonig: Doc strings 😇

mythril/analysis/modules/external_calls.py
```diff
@@ -22,6 +27,22 @@
...
+    def __init__(self, call: Call, user_defined_address: bool) -> None:
+        self.call = call
+        self.user_defined_address = user_defined_address
+
+
+class ExternalCallsAnnotation(StateAnnotation):
```
JoranHonig: Not necessary for this module

mythril/analysis/modules/state_change_external_calls.py
```diff
@@ -0,0 +1,125 @@
...
+    forward all gas to the callee."""
+
+    def __init__(self):
+        """"""
+        super().__init__(
+            name="External calls",
+            swc_id=REENTRANCY,
+            description=DESCRIPTION,
+            entrypoint="callback",
+            pre_hooks=["CALL", "SSTORE"],
```
JoranHonig: Like @nbanmp mentioned we also need to support create. Additionally, the other call kinds should be added here I think (delegate call, etc.)

mythril/analysis/modules/state_change_external_calls.py
```diff
@@ -0,0 +1,125 @@
...
+
+log = logging.getLogger(__name__)
+
+DESCRIPTION = """
+
+Check whether there is a state change of the contract after the execution of an external call
+"""
+
+
+def _get_state_change_issues(
```
JoranHonig: What do you think about having these functions as static class functions

mythril/analysis/modules/state_change_external_calls.py
```diff
@@ -0,0 +1,125 @@
+from mythril.analysis.ops import Variable, VarType
+from mythril.analysis.swc_data import REENTRANCY
+from mythril.analysis.modules.base import DetectionModule
+from mythril.analysis.report import Issue
+from mythril.analysis.call_helpers import get_call_from_state
+from mythril.analysis.modules.external_calls import CallIssue, ExternalCallsAnnotation
```
JoranHonig: I think different modules shouldn't depend on one another. Now it's still relatively simple, but later on this will become super hard to maintain

norhh: Yes, I was thinking we can place all the annotations in 1 file(probably that might solve the future problems of this sort). WDYT?

JoranHonig: I'd prefer having them in each of the modules. 
That keeps them self contained, and understandable. 

I don't think the limited amount of code reuse is worth sacrificing that bit of understandability

b-mueller:
> * an issue is reported for every external call ( with medium severity if to an arbitrary address, otherwise low)
> * a separate issue is reported for re-entrancy vulnerabilities

Yep, that's exactly how I would report it.

mythril/analysis/modules/external_calls.py
```diff
@@ -31,8 +34,24 @@ def _analyze_state(state):
     node = state.node
     gas = state.mstate.stack[-1]
     to = state.mstate.stack[-2]
-
+    issues = []
     address = state.get_current_instruction()["address"]
+    annotations = cast(
```
JoranHonig: Correct me if I'm wrong but it seems to me that this module does not really use/need any annotations.
If so we can remove this code.

Moreover, I don't think different modules should share annotations, they each have different behaviours modifying the annotation. Which can make for some nasty bugs.
If both modules need to use an annotation I think they need separate annotations

norhh: The data from the computations done in this module is used in state change module(the `user_defined_address`), so i think it might make sense combine these two modules. WDYT?

mythril/analysis/modules/external_calls.py
```diff
@@ -31,11 +33,15 @@ def _analyze_state(state):
     node = state.node
     gas = state.mstate.stack[-1]
     to = state.mstate.stack[-2]
-
+    issues = []
```
JoranHonig: this is just a local variable rite?

mythril/analysis/modules/external_calls.py
```diff
@@ -99,8 +105,8 @@ def _analyze_state(state):
     except UnsatError:
         log.debug("[EXTERNAL_CALLS] No model found.")
         return []
-
-    return [issue]
+    issues.append(issue)
```
JoranHonig: same here, why not just return [issue]?

mythril/analysis/modules/state_change_external_calls.py
```diff
@@ -0,0 +1,197 @@
...
+        annotations = cast(
+            List[StateChangeCallsAnnotation],
+            list(state.get_annotations(StateChangeCallsAnnotation)),
+        )
+        if len(annotations) == 0:
+            log.debug("Creating annotation for state")
+            state.annotate(StateChangeCallsAnnotation())
+            annotations = cast(
+                List[StateChangeCallsAnnotation],
+                list(state.get_annotations(StateChangeCallsAnnotation)),
```
JoranHonig: this annotation is also added for each sstore now

mythril/analysis/modules/state_change_external_calls.py
```diff
@@ -0,0 +1,197 @@
...
+
+        if call is None:
+            return []
+
+        if StateChange._balance_change(call.value):
+            return StateChange._handle_state_change(
+                state, address=address, annotation=annotations[0]
+            )
+
+        if opcode == "CALL":
```
JoranHonig: why are only calls considered here?

mythril/analysis/modules/external_calls.py
norhh: yes, issues are retuned in this function.

mythril/analysis/modules/state_change_external_calls.py
norhh: This is to detect state change after sending gas with call.value() right?

mythril/analysis/modules/state_change_external_calls.py
norhh: sorry, i can't get it, can you elaborate?

JoranHonig: This function is executed for each callback, which includes callbacks for sstore insns. 
If there is a callback for a sstore insn no new annotation should be created

mythril/analysis/modules/state_change_external_calls.py
JoranHonig: This isn't to detect state changes, this is the check for a call. So I'm wondering if other types of calls should be considered here as well.

mythril/analysis/modules/external_calls.py
```diff
@@ -5,6 +5,7 @@
 from mythril.analysis.swc_data import REENTRANCY
 from mythril.analysis.modules.base import DetectionModule
 from mythril.analysis.report import Issue
+from mythril.analysis.call_helpers import get_call_from_state
```
JoranHonig: could you just revert this module to the previous version? The changes added to this module only add a dependency to the helpers, without adding any features.

Call should never be None, as the function is only called for call instructions.

norhh: I think Call returns None for internal calls. maybe i should change the name of that function later.

mythril/analysis/modules/state_change_external_calls.py
norhh: Argh! nice catch, some residual if condition.

norhh: Oh! I am not a security expert, do reentrancy sort of problems happen with other calls?

norhh:
I did the following changes:-
- Add possibility of address == 0 as external call
- Changing report issues from calls to state change locations, as in most cases it is possible to have multiple state changes after a call rather than the other way(this is temporary)
- Remove the annotation removal as we may not be able to detect the other state changes after a call

PR #921, opened 2019-02-12
Find issues for EXP operation

### LINKED ISSUE #350: Consider EVM instruction EXP in Integer Overflow
**Describe the bug**
Mythril doesn't detect overflow by exponentiation.

**To Reproduce**

Create solidity contract: 

```solidity

pragma solidity ^0.4.23;

contract IntegerOverflow {
    uint public count = 2;

    function run(uint256 input) public {
        count = count ** input;
    }
}
```

compile it, and then run `myth` on it.

**Expected behavior**

```
==== Integer Overflow ====
Type: Warning
Contract: IntegerOverflow
Function name: **ambiguous** r
PC address: 174
A possible integer overflow exists in the function `**ambiguous** r`.
The addition, multiplication or exponentiation may result in a value higher than the maximum representable integer.
--------------------
In file: test.sol:1

count = count ** input
```

**Desktop:**
 - OS: Ubuntu
 - Version Bionic

norhh:
fixes #350


mythril/analysis/modules/integer.py
```diff
@@ -88,66 +90,83 @@ def execute(self, state: GlobalState):
...
+            )
+            constraints = [c1, c2]
+        elif op1.symbolic:
+            c1 = op1 >= symbol_factory.BitVecVal(ceil(256 / log2(op0.value)), 256)
+            constraints = [c1]
+        elif op0.symbolic:
+            c1 = op0 >= symbol_factory.BitVecVal(2 ** ceil(256 / op1.value), 256)
+            constraints = [c1]
+        else:
+            constraints = [op0.value ** op1.value > 2 ** 256]
```
```suggestion
            constraints = [op0.value ** op1.value >= 2 ** 256]
```

mythril/analysis/modules/integer.py
```diff
@@ -88,66 +90,83 @@ def execute(self, state: GlobalState):
...
             return
 
-        annotation = OverUnderflowAnnotation(state, "subtraction", c)
+        annotation = OverUnderflowAnnotation(state, "subtraction", [c])
+        op0.annotate(annotation)
+
+    def _handle_exp(self, state):
+        op0, op1 = self._get_args(state)
+        if op0.symbolic and op1.symbolic:
+            c1, c2 = (
```
```suggestion
            constraint = And(
```
If we do that, then we don't need all the other changes which require a list of constraints vs just the one constraint

mythril/laser/ethereum/instructions.py
```diff
@@ -519,16 +519,22 @@ def exp_(self, global_state: GlobalState) -> List[GlobalState]:
...
                     "(" + str(simplify(base)) + ")**(" + str(simplify(exponent)) + ")",
                     256,
+                    base.annotations + exponent.annotations,
                 )
             )
         else:
 
             state.stack.append(
-                symbol_factory.BitVecVal(pow(base.value, exponent.value, 2 ** 256), 256)
+                symbol_factory.BitVecVal(
```
JoranHonig: good catch

mythril/analysis/modules/integer.py
```diff
@@ -31,22 +33,22 @@ class OverUnderflowAnnotation:
     """ Symbol Annotation used if a BitVector can overflow"""
 
     def __init__(
-        self, overflowing_state: GlobalState, operator: str, constraint
+        self, overflowing_state: GlobalState, operator: str, constraints
```
nbanmp: Cam we get a type hint on `constraints`? Also in the constructor below.

mythril/analysis/modules/integer.py
```diff
@@ -150,6 +148,29 @@ def _handle_sub(self, state):
...
 
+    def _handle_exp(self, state):
+        op0, op1 = self._get_args(state)
+        if op0.symbolic and op1.symbolic:
+            constraint = And(
+                op1 > symbol_factory.BitVecVal(256, 256),
+                op0 > symbol_factory.BitVecVal(1, 256),
+            )
+        elif op1.symbolic:
+            constraint = op1 >= symbol_factory.BitVecVal(
```
JoranHonig: There is an edge case where op0.value is either 0 or 1, which I think we would have problems with

mythril/analysis/modules/integer.py
```diff
@@ -150,6 +148,29 @@ def _handle_sub(self, state):
...
+                ceil(256 / log2(op0.value)), 256
+            )
+        elif op0.symbolic:
+            constraint = op0 >= symbol_factory.BitVecVal(
+                2 ** ceil(256 / op1.value), 256
```
JoranHonig: we'll get an exception if op1.value == 0

JoranHonig: We can do an easy optimisation I think `if op1.value == 1: constraint = False`

PR #925, opened 2019-02-14
Increase Documentation

nbanmp:
Some of the wiki was updated and brought into the read the docs, and the link to the documentation was added to the readme.


README.md
```diff
@@ -40,7 +41,7 @@ Instructions for using Mythril Classic are found on the [Wiki](https://github.co
 For support or general discussions please join the Mythril community on [Discord](https://discord.gg/E3YrVtG).
 
 ## Bulding the Documentation
-Mythril Classic's documentation is contained in the `docs` folder. It is based on Sphinx and can be built using the Makefile contained in the subdirectory:
+Mythril Classic's documentation is contained in the `docs` folder and published to [Read the Docs](https://mythril-classic.readthedocs.io/en/master/). It is based on Sphinx and can be built using the Makefile contained in the subdirectory:
```
norhh: shouldn't it be `and is published to` ?I am not an english grammer guy so not totally sure

nbanmp: Hmm... I think they both work, at least they both sound fine to me.

An object is something and is something else.
An object is something and something else.

Adding an is might read a bit nicer though, so I'll go ahead and add it.

docs/source/about.rst
```diff
@@ -0,0 +1,6 @@
+What is Mythril Classic?
+========================
+
+Mythril Classic is a security analysis tool for Ethereum smart contracts. It was `introduced at HITBSecConf 2018 <https://github.com/b-mueller/smashing-smart-contracts/blob/master/smashing-smart-contracts-1of1.pdf>`_.
+
+Mythril Classic detects a range of security issues, including integer underflows, owner-overwrite-to-Ether-withdrawal, and others. However, the analysis will not detect business logic issues and is not equivalent to formal verification.
```
JoranHonig: `and is not equivalent to formal verification`
I think this is a bit ambiguous. Symbolic execution is a formal verification technique, but it isn't sound in most practical cases (which is what is relevant here)
I'd like to rephrase it as follows:
```suggestion
Mythril Classic detects a range of security issues, including integer underflows, owner-overwrite-to-Ether-withdrawal, and others. 
Note that Mythril is targeted at finding common vulnerabilities, and is not able to discover issues in the business logic of an application.
Furthermore, Mythril and symbolic executors, are generally unsound, as they are often unable to explore all possible states of a program.
```

docs/source/security-analysis.rst
```diff
@@ -0,0 +1,92 @@
+Security Analysis
+=================
+
+Run :code:`myth -x` with one of the input options described below will run the analysis modules in the `/analysis/modules <https://github.com/ConsenSys/mythril-classic/tree/master/mythril/analysis/modules>`_ directory.
+
+Mythril detects a range of security issues, including integer underflows, owner-overwrite-to-ether-withdrawals, unprotected selfdestructs, and others. However, the analysis will not detect business logic issues and is not equivalent to formal verification.
```
JoranHonig: duplicate withabout.rst

is this intentional?

JoranHonig:
Is there an easy way to preview changes to the documentation. I'm not really familiar with restructured text, so I don't trust my ability to verify any syntax problems

nbanmp:
`sphinx-build source html` from the `docs/` directory should build the html iirc.

Issue #935, opened 2019-02-20
Integer overflow caught by Harvey but not Mythril - why? 


rocky:
## Description

Here is an example of something that the fuzzer Harvey catches that Mythril doesn't. Why? And should Mythril also detect it? 

## How to Reproduce

```console
$ myth -x -c "60d761002f600b82828239805160001a6073146000811461001f57610021565bfe5b5030600052607381538281f3fe73000000000000000000000000000000000000000030146080604052600436106050576000357c01000000000000000000000000000000000000000000000000000000009004806396e4ee3d146055575b600080fd5b608860048036036040811015606957600080fd5b810190808035906020019092919080359060200190929190505050609e565b6040518082815260200191505060405180910390f35b600081830290509291505056fea165627a7a7230582094ca199c801839337331343f733fa7283c9d7f430a20e7ec95933a083287a6d80029"
$
```

## Expected behavior

Here is what harvey-cli reports:

```console
$ cat ConvertLib-test.json | harvey-cli
{
    "issues": [
        {
            "swc-id": "SWC-101",
            "title": "integer overflow",
            "details": "Integer overflows are often not intended. Make sure your contracts take appropriate precautions when performing arithmetic operations.",
            "code-hash": "0x8faf42982cb593efd96aa4dcc7ba90e8d264168b04fd21bdcae48332bd7d533d",
            "function-signature": "96e4ee3d",
            "bytecode-offset": 163,
            "stack-trace": [],
            "time-secs": 4.1114236
        }
    ],
    "covered-paths": 5,
    "covered-instructions": 99,
    "error": "",
    "total-time-secs": 36.00175275
}
```

## Environment

- Mythril version v0.20.0
-  Python 3.6.6
- Ubuntu bionic

norhh:
@rocky , is the solidity source code available for the test case?

rocky:
Sorry - here it is: 

```
pragma solidity >=0.4.22 <0.6.0;

library ConvertLib {
	function convert(uint amount,uint conversionRate) public pure returns (uint convertedAmount)
	{
		return amount * conversionRate;
	}
}
```

and yes the failure may be because this is a library rather than a contract. Is there an issue already for handling libraries? (Note though that the #936 is not from a library.

PR #938, opened 2019-02-24
handle return for integer module

norhh:
will fix a part of the bug in #935 .

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -231,7 +231,8 @@ class IntegerOverflowUnderflowModule(DetectionModule):
         except TypeError:
             return
         for element in state.mstate.memory[offset : offset + length]:
-            if isinstance(element, Expression):
+            if not isinstance(element, Expression):
+                return
                 for annotation in element.annotations:
                     if isinstance(annotation, OverUnderflowAnnotation):
                         state.annotate(
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -220,6 +220,11 @@ class IntegerOverflowUnderflowModule(DetectionModule):
 
     @staticmethod
     def _handle_return(state: GlobalState) -> None:
+        """
+        Adds all the annotations into the state which correspond to the
+        locations in the memory returned by RETURN opcode.
+        :param state: The Global State
+        """
         stack = state.mstate.stack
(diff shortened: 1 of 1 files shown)


mythril/analysis/modules/integer.py
```diff
@@ -215,6 +218,25 @@ def _handle_jumpi(state):
                 )
             )
 
+    @staticmethod
+    def _handle_return(state: GlobalState) -> None:
```
JoranHonig: Can you add a comment detailing what this function adds.

mythril/analysis/modules/integer.py
```diff
@@ -215,6 +218,30 @@ def _handle_jumpi(state):
...
+        locations in the memory returned by RETURN opcode.
+        :param state: The Global State
+        """
+        stack = state.mstate.stack
+        try:
+            offset, length = get_concrete_int(stack[-1]), get_concrete_int(stack[-2])
+        except TypeError:
+            return
+        for element in state.mstate.memory[offset : offset + length]:
+            if isinstance(element, Expression):
```
```suggestion
            if not isinstance(element, Expression):
                return
```

JoranHonig: To reduce nesting

mythril/analysis/modules/integer.py
```diff
@@ -215,6 +218,30 @@ def _handle_jumpi(state):
...
+                for annotation in element.annotations:
+                    if isinstance(annotation, OverUnderflowAnnotation):
```
JoranHonig: There is a specific getter function, which you can pass the type of annotation you want too receive, could you use that here?

norhh: Isn't that for state annotations?

JoranHonig: Right, nvm. We should add that for these annotations too later 😅

mythril/analysis/modules/integer.py
```diff
@@ -215,6 +218,31 @@ def _handle_jumpi(state):
...
+        :param state: The Global State
+        """
+        stack = state.mstate.stack
+        try:
+            offset, length = get_concrete_int(stack[-1]), get_concrete_int(stack[-2])
+        except TypeError:
+            return
+        for element in state.mstate.memory[offset : offset + length]:
+            if not isinstance(element, Expression):
+                return
```
```suggestion
                continue
```

PR #943, opened 2019-02-27
Add the discovery time to Issue class

norhh:


mythril/start_time.py
```diff
@@ -0,0 +1,7 @@
+from time import time
+from mythril.support.support_utils import Singleton
+
+
+class StartTime(metaclass=Singleton):
```
JoranHonig: documentation 😇

JoranHonig: Also I think we should move this to support

mythril/mythril.py
```diff
@@ -36,8 +36,11 @@
 from mythril.support.truffle import analyze_truffle_project
 from mythril.ethereum.interface.leveldb.client import EthLevelDB
 from mythril.laser.smt import SolverStatistics
+from mythril.start_time import StartTime
+
 
 log = logging.getLogger(__name__)
+start_time = StartTime()
```
JoranHonig: I'm thinking that we should maybe re-initialize the start time for each contract that is being analyzed, wdyt?

norhh: what was I thinking when I put it there :man_facepalming: , will move it there :+1:

mythril/start_time.py
norhh: Moved

PR #948, opened 2019-03-05
Combine two functions in integer module

norhh:
the two functions have nearly the same code


JoranHonig:
I don't feel super strongly about this, but the separation was intentional. My idea was that if you ever wanted to extend the behaviour of either jumpi or sstore, the current implementation would be easier

PR #949, opened 2019-03-05
Handle -1 case in solidity file mapping

norhh:


mythril/analysis/report.py
```diff
@@ -103,6 +103,18 @@ def as_dict(self):
 
         return issue
 
+    def _add_false_positive_tags(self):
```
JoranHonig: Add internal compiler file information

mythril/analysis/report.py
```diff
@@ -103,6 +103,18 @@ def as_dict(self):
...
+        """
+        Adds the false positive to description and changes severity to low
+        """
+        self.severity = "Low"
+        self.description_tail += (
+            " This happened in the internal compiler generated code, "
```
JoranHonig: I don't want to say this is a false positive, rather just something like: "This issue is reported for internal compiler generated code."
wdyt?

mythril/analysis/report.py
```diff
@@ -115,6 +126,8 @@ def add_code_info(self, contract):
             self.filename = codeinfo.filename
             self.code = codeinfo.code
             self.lineno = codeinfo.lineno
+            if self.lineno is None:
```
JoranHonig:
```suggestion
            if codeinfo.solidity_file_idx == -1:
```
Should be equivalent, but I think this is a bit more self-explanatory

JoranHonig: what do you think?

norhh: I think codeinfo and codeinfo.filename attribute doesn't have solidity_file_idx.

JoranHonig: ah I had mistaken the type, what about `codeinfo.solc_mapping.solidity_file_idx`

PR #951, opened 2019-03-06
Add deprecation warning for truffle

norhh:
shows deprecation warning when using truffle.


mythril/support/truffle.py
```diff
@@ -181,7 +194,10 @@ def get_mappings(source, deployed_source_map):
         if len(mapping) > 2 and len(mapping[2]) > 0:
             idx = int(mapping[2])
 
-        lineno = source.encode("utf-8")[0:offset].count("\n".encode("utf-8")) + 1
+        if idx == -1:
```
JoranHonig: this is not part of this pr right?

norhh: Yes, removed it :+1:

PR #958, opened 2019-03-14
Change execution-timeout for solver

norhh:
Chnage execution-timeout for solver to `execution-timeout/tx_count`


JoranHonig:
This doesn't solve the problem for me, running
`myth -x rubixi.sol --execution-timeout 20 -t 1 -v4`
Doesn't return after 20 seconds

I'll try to triage this a bit more

JoranHonig:
I'm thinking, can we make solver use the timeout:

```python
timehandler.time_remaining() % (timehandler.execution_time // timehandler.transaction_count)
```

or create a separate method which returns this called `transaction_time_remaining()`
as I can see us making mistakes if the execution time is set like it is in this pr

norhh:
We are not going to calibarate the total time anyway right, probably renaming variables and function names to self._transaction_execution_timeout and transaction_time_remaining() might do the work now. What do you think?

JoranHonig:
@norhh sounds good

mythril/laser/ethereum/time_handler.py
```diff
@@ -5,14 +5,16 @@
...
+    def start_execution(self, execution_time: int, transaction_count: int):
         self._start_time = int(time.time() * 1000)
-        self._execution_time = execution_time * 1000
+        self._transaction_execution_timeout = execution_time * 1000 // transaction_count
 
-    def time_remaining(self):
-        return self._execution_time - (int(time.time() * 1000) - self._start_time)
+    def transaction_time_remaining(self):
+        return self._transaction_execution_timeout - (
+            int(time.time() * 1000) - self._start_time
```
JoranHonig: this will break for the second transaction right? or is the time reset for each transaction?

norhh: Nice catch, I removed that to catch something and forgot to add it back :man_facepalming:

mythril/laser/ethereum/time_handler.py
```diff
@@ -5,14 +5,19 @@
 class TimeHandler(object, metaclass=Singleton):
     def __init__(self):
         self._start_time = None
-        self._execution_time = None
+        self._transaction_execution_timeout = None
 
-    def start_execution(self, execution_time):
+    def reset_start_time(self):
```
JoranHonig: With this the creation transaction isn't counted

PR #969, opened 2019-03-20
Display the discoveryTime

### LINKED ISSUE #941: Record discovery time for reported issues
## Description
If we discover an issue we should store the discovery time, as this allows us to evaluate the performance of different optimisation techniques.

norhh:
fixes #941

### DIFF
--- a/mythril/analysis/report.py
+++ b/mythril/analysis/report.py
@@ -231,11 +231,7 @@ class Report:
                     },
                     "severity": issue.severity,
                     "locations": [{"sourceMap": "%d:1:%d" % (issue.address, idx)}],
-                    "extra": {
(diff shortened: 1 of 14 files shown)


mythril/analysis/report.py
```diff
@@ -231,7 +231,11 @@ def as_swc_standard_format(self):
                     },
                     "severity": issue.severity,
                     "locations": [{"sourceMap": "%d:1:%d" % (issue.address, idx)}],
-                    "extra": {},
+                    "extra": {
+                        "discoveryTime": "{}".format(
```
JoranHonig: This is the wrong type right? A string, while it should be an integer

norhh: Changed :+1:

Issue #954, opened 2019-03-11
-xo jsonv2 should return bytecode in sourcelist not some hash of it. 


rocky:
When option `-xo jsonv2` is used the `sourceList` field contains some unintelligable hexadecimal number. It should instead be the bytecode associated with the bytecode offset it reports. 

I was made aware of the following in work that shashank-srikant was doing in his machine learning project using Mythril Classic.  

## How to Reproduce

Consider this simple Solidity program:
```solidity
/*
 * @source: https://github.com/ConsenSys/evm-analyzer-benchmark-suite
 * @author: Suhabe Bugrara
 */

pragma solidity ^0.5.0;

contract AssertMinimal {
    function run() public {
        assert(false);
    }
}
```
when I run Mythril classic on it with `myth -xo jsonv2 contracts/assert_minimal.sol  | jq`

I get: 

```
[
  {
    "issues": [
      {
        "description": {
          "head": "A reachable exception has been detected.",
          "tail": "It is possible to trigger an exception (opcode 0xfe). Exceptions can be caused by type errors, division by zero, out-of-bounds array access, or assert violations. Note that explicit `assert()` should only be used to check invariants. Use `require()` for regular input checking."
        },
        "extra": {},
        "locations": [
          {
            "sourceMap": "90:1:0"
          }
        ],
        "severity": "Low",
        "swcID": "SWC-110",
        "swcTitle": "Assert Violation"
      }
    ],
    "meta": {},
    "sourceFormat": "evm-byzantium-bytecode",
    "sourceList": [
      "0x75f42bcc665c4533bd2ab93eb909fb500e730ed71ccefd8d493eba1aec2944f6"
    ],
    "sourceType": "raw-bytecode"
  }
]
```
The entry `sourceList` entry `"0x75f42bcc665c4533bd2ab93eb909fb500e730ed71ccefd8d493eba1aec2944f6"` is useless here. 

Instead the corresponding bytecode should appear. Note that in my invocation, I didn't specify any bytecode but instead gave Solidity source, so I can't really know what bytecode MythX was using. 

Lastly, I should note that this kind of problem appears in MythX as well and that should be addressed as well. But since Mythril Classic is its own thing it also needs to be addressed in this situation as well.

****
<!-- This section sometimes is optional but helpful to us.

Please modify for your setup

- Mythril version: `Mythril version v0.20.0`
- Solidity compiler and version: `Version: 0.5.2+commit.1df8f40c.Linux.g++`
[...]

PR #971, opened 2019-03-23
Fix the jsonv2 output for the type of the contract

norhh:
related to #954


mythril/support/source_support.py
```diff
@@ -17,7 +15,7 @@ def __init__(
         self.source_type = source_type
         self.source_format = source_format
         self.source_list = source_list or []
-        self.meta = meta
+        self._source_hash = []
```
JoranHonig: What is the _source_hash variable used for. It seems like its overwritten later

norhh: It can be used to query file index from hash.

PR #973, opened 2019-03-26
Use the current state constraints over the constraint at the annotati…

### LINKED ISSUE #972: Mythril reports a False Positive on swc-101
## Description
The problem happens for the [overflow_simple_add_fixed](https://github.com/SmartContractSecurity/SWC-registry/blob/master/test_cases/solidity/integer_overflow_and_underflow/overflow_simple_add_fixed/overflow_simple_add_fixed.sol) .

## How to Reproduce
```console
$ ./myth -x thefile.sol
==== Integer Overflow ====
SWC ID: 101
Severity: High
Contract: Overflow_Add
Function name: add(uint256)
PC address: 204
Estimated Gas Usage: 668 - 763
The binary addition can overflow.
The operands of the addition operation are not sufficiently constrained. The addition could therefore result in an integer overflow. Prevent the overflow by checking inputs or ensure sure that the overflow is caught by an assertion.
--------------------
In file: ../../over.sol:11

a + b
...
$
```

## Expected behavior
The above issue shouldn't be reported

## Environment
Please modify for your setup

- Mythril version: output from  `develop`
- Python version: `3.6.3`
- OS and Version: Ubuntu 17.10

norhh:
fixes #972

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -290,7 +290,7 @@ class IntegerOverflowUnderflowModule(DetectionModule):
             try:
 
                 transaction_sequence = solver.get_transaction_sequence(
-                    state, node.constraints + [annotation.constraint]
+                    state, state.node.constraints + [annotation.constraint]
                 )
             except UnsatError:
                 continue
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -290,7 +290,7 @@ class IntegerOverflowUnderflowModule(DetectionModule):
             try:
 
                 transaction_sequence = solver.get_transaction_sequence(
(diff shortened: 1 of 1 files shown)


mythril/analysis/modules/integer.py
```diff
@@ -290,7 +290,7 @@ def _handle_transaction_end(self, state: GlobalState) -> None:
             try:
 
                 transaction_sequence = solver.get_transaction_sequence(
-                    state, node.constraints + [annotation.constraint]
+                    state, state.node.constraints + [annotation.constraint]
```
```suggestion
                    state, state.mstate.constraints + [annotation.constraint]
```

Great catch! Probably got in there bc copy pasta.

The suggested change is bc I think we should start to move away from using the constraints stored in the nodes (this is one of those artifacts from the old laser)

PR #975, opened 2019-03-27
Set low priority to other internal compiler generated code

norhh:
There is some code which is used by all the code and which points to all the code in the src map, These are internal generated errors, these should be ignored. one example of it happens in this code https://github.com/SmartContractSecurity/SWC-registry/blob/master/test_cases/solidity/write_to_arbitrary_storage_location/arbitrary_location_write_simple/arbitrary_location_write_simple.sol


mythril/solidity/soliditycontract.py
```diff
@@ -140,6 +140,17 @@ def get_source_info(self, address, constructor=False):
         lineno = mappings[index].lineno
         return SourceCodeInfo(filename, lineno, code, mappings[index].solc_mapping)
 
+    def _is_autogenerated_code(self, offset: int, length: int, file_index: int) -> bool:
```
JoranHonig: offset isn't used

JoranHonig:
@norhh Do you have some documentation for this? I agree that src locations referring to the entire contract are not very usefull but why wouldn't they use -1 in this case?

norhh:
According to chriseth, -1 is used when the file itself isn't know

JoranHonig:
But that an entire contract is referenced isn't specified as compiler gen'd code right?

norhh:
As i couldn't find any documentation regarding this, so I pinged chriseth(lead solc developer) at gitter, his reply `such code is usually generated by the compiler itself, i.e. those are utility functions that are shared and where it does not make much sense to map them to a specific point in the code`.

JoranHonig:
@norhh Oh well, makes me wonder what the difference is between -1 and this then

JoranHonig:
In any case the implemented approach probably fails if there are multiple contracts in one file (right?)

norhh:
Yes, I am currently thinking about dealing with it.

norhh:
How about this :- 
- left strip and right strip the white spaces
- remove the initial contract/library prefix
- Do a simple iteration and till the end and check if the parenthesis were matched properly

JoranHonig:
@norhh How about the following, when we create the solidity contract object we call into solc to get the sourcemap. In addition to this we can let it generate the ast for the contract, using this ast we'll be able to find the sourcemap for the entire contract and compare it to the value we find.

JoranHonig:
@norhh How about the following, when we create the solidity contract object we call into solc to get the sourcemap. In addition to this we can let it generate the ast for the contract, using this ast we'll be able to find the sourcemap for the entire contract and compare it to the value we find.

mythril/support/truffle.py
```diff
@@ -170,6 +171,18 @@ def get_sigs_from_truffle(sigs, contract_data):
         sigs.add("0x" + sha3(signature)[:4].hex(), signature)
 
 
+def _is_autogenerated_code(length: int, file_index: int, code: str) -> bool:
```
JoranHonig: truffle support is depricated in mythril classic right 😉

mythril/solidity/soliditycontract.py
```diff
@@ -20,9 +22,10 @@ def __init__(self, solidity_file_idx, offset, length, lineno, mapping):
 class SolidityFile:
     """Representation of a file containing Solidity code."""
 
-    def __init__(self, filename, data):
+    def __init__(self, filename: str, data: str, full_contract_source: Set[str]):
```
JoranHonig: can you add a docstring that describes the parameters used for this function?

mythril/solidity/soliditycontract.py
```diff
@@ -117,6 +125,14 @@ def __init__(self, input_file, name=None, solc_args=None, solc_binary="solc"):
 
         super().__init__(code, creation_code, name=name)
 
+    @staticmethod
+    def get_full_contract_sources(ast: Dict) -> Set[str]:
```
JoranHonig: Docstring?

mythril/solidity/soliditycontract.py
```diff
@@ -117,6 +131,19 @@ def __init__(self, input_file, name=None, solc_args=None, solc_binary="solc"):
...
+        """
+        Takes AST and returns the source map of the contract
```
JoranHonig:
```suggestion
        Takes a solc AST and gets the src mappings for all the contracts defined in the top level of the ast
```

Issue #964, opened 2019-03-18
Move symbolic graph to plugin


JoranHonig:

JoranHonig:
This includes the features implemented in:
mythril/analysis/traceexplore.py

PR #977, opened 2019-03-28
Removes dependency on cfg elements from detection modules

JoranHonig:
This is part of a larger effort to move the cfg logic to a separate module 
related to #964


mythril/analysis/modules/external_calls.py
```diff
@@ -56,8 +56,8 @@ def _analyze_state(state):
             )
 
             issue = Issue(
-                contract=node.contract_name,
-                function_name=node.function_name,
+                contract=state.node.contract_name,
```
norhh: Node is used here.

mythril/analysis/modules/external_calls.py
```diff
@@ -83,7 +83,7 @@ def _analyze_state(state):
             )
 
             issue = Issue(
-                contract=node.contract_name,
+                contract=state.node.contract_name,
```
norhh: Node is used here

PR #979, opened 2019-03-28
Support tx sequences

norhh:


JoranHonig:
What is the usecase for this?

norhh:
tx_sequence

Maestro can use it to prune of lot's of transaction sequences if it has some hints from other tools. And it can also be used for debugging.

JoranHonig:
Okay, could you try to fit this in a plugin? 
I want to start moving as many non-core code to plugins as possible

PR #983, opened 2019-03-29
Use plugin loader and factory

### LINKED ISSUE #914: Implement Plugin Loader for Laser
- [x] implement loader
 - [ ] integrate loader

JoranHonig:
fixes #914

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -27,3 +27,4 @@ rlp>=1.0.1
 transaction>=2.2.1
 z3-solver-mythril>=4.8.4.1
 pysha3
+matplotlib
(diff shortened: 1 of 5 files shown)


requirements.txt
```diff
@@ -27,3 +27,4 @@ rlp>=1.0.1
 transaction>=2.2.1
 z3-solver-mythril>=4.8.4.1
 pysha3
+matplotlib
```
norhh: Are we going to draw graphs from mythril? Also this should be included in setup.py right?

JoranHonig: Only developers of mythril will probably have intrest in this feature, there is also not really a way to enable this module atm without changing the code.

But we could add it to the setup.py as well

PR #986, opened 2019-04-01
Use instruction coverage plugin

### LINKED ISSUE #978: move coverage calculation to plugin

JoranHonig:
Removes coverage related logic from core svm

Extends the instruction coverage plugin with functionality to report coverage increase between transactions

closes #978

### DIFF
--- a/mythril/laser/ethereum/plugins/implementations/coverage.py
+++ b/mythril/laser/ethereum/plugins/implementations/coverage.py
@@ -21,6 +21,11 @@ class InstructionCoveragePlugin(LaserPlugin):
 
     """
 
+    def __init__(self):
+        self.coverage = {}  # type: Dict[str, Tuple[int, List[bool]]]
+        self.initial_coverage = 0
+        self.tx_id = 0
+
     def initialize(self, symbolic_vm: LaserEVM):
         """Initializes the instruction coverage plugin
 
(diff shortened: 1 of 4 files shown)


mythril/laser/ethereum/plugins/implementations/coverage.py
```diff
@@ -47,13 +54,36 @@ def execute_state_hook(global_state: GlobalState):
...
+
+        @symbolic_vm.laser_hook("stop_sym_trans")
+        def execute_stop_sym_trans_hook():
+            end_coverage = self._get_covered_instructions()
+            log.info(
+                "Number of new instructions covered in tx %d: %d"
+                % (self.tx_id, end_coverage - self.initial_coverage)
+            )
+            self.tx_id += 1
+
```
norhh: When will this function get executed?

mythril/laser/ethereum/plugins/implementations/coverage.py
```diff
@@ -47,13 +54,36 @@ def execute_state_hook(global_state: GlobalState):
...
+    def _get_covered_instructions(self) -> int:
```
norhh: When will this function get executed?

JoranHonig: In the two previous hooks, whom are executed at each transaction end & start

mythril/laser/ethereum/svm.py
```diff
@@ -527,6 +490,10 @@ def register_laser_hooks(self, hook_type: str, hook: Callable):
             self._start_sym_exec_hooks.append(hook)
         elif hook_type == "stop_sym_exec":
             self._stop_sym_exec_hooks.append(hook)
+        elif hook_type == "start_sym_trans":
+            self._start_sym_trans_hooks.append(hook)
+        elif hook_type == "stop_sym_trans":
```
norhh: As this part might  increase in the future, we can make a static dictionary of strings and functions and directly use it, so each hook access would be in O(1) if we move the static dictionary into the svm initialization. but it is a negligible optimization, anyway using a dictionary in these cases is a bit more clean, WDYT?

JoranHonig: I agree, was planning on doing something similar. But not in this pr 😅

PR #989, opened 2019-04-03
Coverage based search strategy

JoranHonig:
Tries to direct symbolic execution towards yet uncovered instructions

@norhh do you think this could be used to counteract the issues with big contracts that the function entry of some functions might not be reached? (At least in the first transaction) The instruction coverage plugin should be extensible to keep track of coverages in different transactions, which would mitigate the problem over multiple transactions.


JoranHonig:
Possible future additions:
- branch coverage based search strategy
- BB run through (basically, if there is a global_state in the middle of a basic block, execute that transaction first) In itself not interesting, but other strategies might benefit from this ( specifically the previous one )

Issue #968, opened 2019-03-19
Implement the pending opcodes for EVM


norhh:
## Description
The following opcodes and precompiles should be implemented
- [x] Create
- [x] Create2
- [x] Staticcall
- [x] log
- [x] extcodehash
- [x] expmod
- [x] ecadd
- [x] ecmul
- [x] ecpairing

JoranHonig:
This seems to be a done issue -> closing 🙌

PR #991, opened 2019-04-04
Implement static call

norhh:


mythril/laser/ethereum/instructions.py
```diff
@@ -68,14 +69,15 @@ class StateTransition(object):
     if `increment_pc=True`.
     """
 
-    def __init__(self, increment_pc=True, enable_gas=True):
+    def __init__(self, increment_pc=True, enable_gas=True, static_check=False):
```
JoranHonig: Can you change the variable name to something that is more clear about the purpose.

mythril/laser/ethereum/instructions.py
```diff
@@ -1854,6 +1864,21 @@ def call_(self, global_state: GlobalState) -> List[GlobalState]:
             )
             return [global_state]
 
+        if environment.static:
+            if isinstance(value, int) and value > 0:
```
JoranHonig: I don't think this is possible at a transaction creation level

mythril/laser/ethereum/state/environment.py
```diff
@@ -49,6 +50,7 @@ def __init__(
         self.gasprice = gasprice
         self.origin = origin
         self.callvalue = callvalue
+        self.static = static
```
JoranHonig: Variable name not clear.
How about `write_permission`

mythril/laser/ethereum/instructions.py
norhh: meaning?

mythril/laser/ethereum/state/environment.py
norhh: Variable name is good right? Like it says something like `environment is static` if static is true,

mythril/laser/ethereum/instructions.py
```diff
@@ -68,14 +69,18 @@ class StateTransition(object):
...
+    ):
         """
 
         :param increment_pc:
         :param enable_gas:
+        :param state_mutation_func: The function mutates state
         """
         self.increment_pc = increment_pc
         self.enable_gas = enable_gas
+        self.state_mutation_instruction = state_mutation_instruction
```
JoranHonig:
```suggestion
        self.is_state_mutation_instruction = is_state_mutation_instruction
```

mythril/laser/ethereum/instructions.py
JoranHonig: Ah nvm, github code folding had me confused

mythril/laser/ethereum/instructions.py
```diff
@@ -1871,20 +1899,70 @@ def call_(self, global_state: GlobalState) -> List[GlobalState]:
...
     def call_post(self, global_state: GlobalState) -> List[GlobalState]:
         """
 
+        :param global_state:
+        :return:
+        """
+
+        return self.post_handler(global_state, function_name="call")
+
+    @StateTransition()
```
JoranHonig: This is also a state mutating instruction according to the yellowpaper

norhh: But the EIP calls it some exception https://github.com/ethereum/EIPs/blob/master/EIPS/eip-214.md .

mythril/laser/ethereum/state/environment.py
JoranHonig: Its just that in the yellow paper this variable is called $$I_w$$ and the description is "the permission to make modifications to the state". 
The way to get this value to be true is "STATICCALL", which makes the name `static` a reasonable one. 
However, I feel like write_permission or has_write_permission directly confers what it is that the variable describes. 
Anyway, this is not really a dealbreaker 😅

JoranHonig: Lol nvm, documentation as always is not consistent. In other places they call the flag a static flag, so we can reasonably assume that anybody should be able to understand the variable name

mythril/laser/ethereum/instructions.py
JoranHonig: I've checked the geth and parity implementations an the EIP is indeed what is implemented. 
So the status quo is good 👍

palkeo:
Is there anything preventing this PR to be merged?

(bug #968 )

mythril/laser/ethereum/instructions.py
```diff
@@ -1894,17 +1971,16 @@ def callcode_post(self, global_state: GlobalState) -> List[GlobalState]:
...
 
         :param global_state:
         :return:
         """
         instr = global_state.get_current_instruction()
         environment = global_state.environment
-
         try:
-            callee_address, callee_account, call_data, value, gas, _, _ = get_call_parameters(
+            callee_address, callee_account, call_data, value, gas, memory_out_offset, memory_out_size = get_call_parameters(
```
nbanmp: The EIP https://github.com/ethereum/EIPs/pull/214/files#diff-29d8cb8d19a769c141160ce5c1e65caeR31 claims that there are only 6 parameters for STATICCALL and value is always 0.

norhh: In this case it returns 0. there is the parameter `with_value` which is False in this case.

PR #1000, opened 2019-04-10
Fix jsonv2 log format

### LINKED ISSUE #999: Log format isn't fully up to spec

dmuhs:
Fixes #999.
Note: #1000 Yay! :tada:


norhh:
@dmuhs , something similar is also present in `mythril/interafaces/cli.py` in the `exit_with_error()` function, can you also change that?

dmuhs:
@norhh Fixed, thanks for the catch!

PR #1004, opened 2019-04-16
Support transaction sequence display

### LINKED ISSUE #993: Add transaction trace output to the json output format

norhh:
fixes #993 
- This PR doesn't display the transaction sequence in jsonv2 yet(will show in text and other output forms)


mythril/laser/ethereum/svm.py
```diff
@@ -482,18 +513,8 @@ def register_hooks(self, hook_type: str, hook_dict: Dict[str, List[Callable]]):
...
-            self._execute_state_hooks.append(hook)
-        elif hook_type == "start_sym_exec":
-            self._start_sym_exec_hooks.append(hook)
-        elif hook_type == "stop_sym_exec":
-            self._stop_sym_exec_hooks.append(hook)
-        elif hook_type == "start_sym_trans":
-            self._start_sym_trans_hooks.append(hook)
-        elif hook_type == "stop_sym_trans":
-            self._stop_sym_trans_hooks.append(hook)
+        if hook_type in self.laser_hooks_dict:
```
JoranHonig: why is this in this pr?

mythril/laser/ethereum/state/world_state.py
```diff
@@ -26,6 +26,7 @@ def __init__(
         self.node = None  # type: Optional['Node']
         self.transaction_sequence = transaction_sequence or []
         self._annotations = annotations or []
+        self.initial_state_account = {}  # type: Dict
```
JoranHonig: It isn't clear what `initial_state_account` is

mythril/laser/ethereum/instructions.py
```diff
@@ -1842,6 +1842,14 @@ def call_(self, global_state: GlobalState) -> List[GlobalState]:
             callee_address, callee_account, call_data, value, gas, memory_out_offset, memory_out_size = get_call_parameters(
                 global_state, self.dynamic_loader, True
             )
+
+            if callee_account is not None and callee_account.code.bytecode == "":
```
JoranHonig: I don't think this edgecase should be implemented in the call instruction, I'd say that this belongs in the transaction handling

mythril/laser/ethereum/svm.py
norhh: I first added a hook for recording the initial account, it looked a bit of an overkill, so reverted that.(anyway this might help)

PR #1013, opened 2019-04-28
Set default tx count to 1 in graph generator

b-mueller:
Fixes crash when tx count is not explicitly provided.


norhh:
I think this was recently fixed, can you try it with the latest develop branch?

b-mueller:
Yep, it's fixed in develop :)

PR #1014, opened 2019-04-30
Supports tx sequence lists

### LINKED ISSUE #993: Add transaction trace output to the json output format

norhh:


mythril/laser/ethereum/instructions.py
```diff
@@ -1842,6 +1842,14 @@ def call_(self, global_state: GlobalState) -> List[GlobalState]:
             callee_address, callee_account, call_data, value, gas, memory_out_offset, memory_out_size = get_call_parameters(
                 global_state, self.dynamic_loader, True
             )
+
+            if callee_account is not None and callee_account.code.bytecode == "":
+                log.debug("The call is related to ether transfer between accounts")
```
norhh: without this, it seems to cause some weird index errors(due to the way cfg is handled) if this isn't placed here. It's a bit hard to place it on the transaction handling area so this might be a quicker way to fix that.

mythril/laser/ethereum/svm.py
```diff
@@ -149,6 +168,11 @@ def sym_exec(
                     "No contract was created during the execution of contract creation "
                     "Increase the resources for creation execution (--max-depth or --create-timeout)"
                 )
+            else:
+                # Replace previous initial state with new ones after contract creation
+                self.set_standard_initial_state(
```
JoranHonig: Why is this necessary. In this case the initial state would just be the one with the creator and attacker account right? The creation transaction is included in the transaction sequence

mythril/laser/ethereum/svm.py
```diff
@@ -132,10 +148,13 @@ def sym_exec(
 
         if main_address:
             log.info("Starting message call transaction to {}".format(main_address))
+            self.set_standard_initial_state(self.world_state.accounts)
```
JoranHonig: this can be moved before the if statement, so we don't need the statement on ln 157 as well

mythril/laser/ethereum/svm.py
```diff
@@ -114,6 +115,21 @@ def accounts(self) -> Dict[str, Account]:
         """
         return self.world_state.accounts
 
+    def set_standard_initial_state(
+        self, accounts: Dict[str, Account], ignore_addr=True
+    ):
+        initial_state = self.world_state.initial_state_account
```
JoranHonig: What's the reason for introducing a new satastructure? Both initial state, and initial_state_account in the world state class

mythril/analysis/report.py
```diff
@@ -64,6 +63,24 @@ def __init__(
         self.source_mapping = None
         self.discovery_time = time() - StartTime().global_start_time
         self.bytecode_hash = get_code_hash(bytecode)
+        if transaction_sequence is None:
+            self.tx_sequence_users = None
```
JoranHonig: why are these variables?

mythril/analysis/report.py
```diff
@@ -64,6 +63,24 @@ def __init__(
...
+            self.transaction_sequence_jsonv2 = None
+        else:
+            self.tx_sequence_users = json.dumps(transaction_sequence, indent=4)
+            self.transaction_sequence_jsonv2 = self.add_block_data(transaction_sequence)
+
+    @staticmethod
+    def add_block_data(transaction_sequence: Dict):
```
JoranHonig: Shouldn't this be done by the logic that creates the concrete transaction sequence?

mythril/analysis/solver.py
```diff
@@ -109,29 +116,37 @@ def get_transaction_sequence(global_state, constraints):
...
-        concrete_transactions[tx_id]["caller"] = "0x" + (
+        value = model.eval(transaction.call_value.raw, model_completion=True).as_long()
+        concrete_transaction["value"] = "0x%x" % value
+        origin = "0x" + (
             "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
         ).zfill(40)
+        concrete_transaction["origin"] = origin
+        concrete_transaction["address"] = "%s" % transaction.callee_account.address
+        concrete_transactions.append(concrete_transaction)
+        min_price_dict[origin] = min_price_dict.get(origin, 0) + value
```
JoranHonig: I don't think origin is necessarily the one who pays the price right? The difference between origin and sender

mythril/laser/ethereum/svm.py
norhh: there is 1 var: initial_state_account, which will just have an initial account state and is propagated everywhere through reference everywhere.

mythril/analysis/report.py
norhh: But we need to use this from symbolic tx sequence stuff right?

mythril/analysis/report.py
norhh: tx_sequence_users is used for displaying for users with "<blockTime>" sort of field than some concrete field. (tx_sequence_jsonv2 is for maestro)

mythril/laser/ethereum/svm.py
norhh: Here the initial state would have the created account.

mythril/analysis/solver.py
norhh: It's acutally caller which is being used, according to mythril origin can literally be anything.Will rename them.

mythril/laser/ethereum/svm.py
JoranHonig: That doesn't make much sense. Then the creation transaction will create another new account, if you try to replay the transaction sequence?

mythril/analysis/report.py
JoranHonig: But what's the reason for storing them as class variables versus just calculating when they are displayed

mythril/analysis/report.py
JoranHonig: What do you mean?

mythril/laser/ethereum/svm.py
norhh: The creation transaction isn't included in the steps I think.

mythril/analysis/report.py
norhh: We don't have the concrete data of these gasLimit, blockNumber etc right?

mythril/analysis/report.py
norhh: the function self.add_block_data, and also json.dumps() should be written in multiple places if that's the case, in this case it's just easily reusing them.

mythril/analysis/report.py
JoranHonig: The blocknumber could be constrained right?

JoranHonig: Same for some of the other variables, they are part of the execution environment and can be queried through some of the opcodes

mythril/analysis/report.py
JoranHonig: I can find only one use of the variables. 
My main point is that the actual data is in transaction_sequence. 
Creating two variables which maintain some representation of the same data doesn't seem like a good approach

I'd either store the initial data, and do the conversion when you need it in some function. 
Or create some functions which do something like
```python
@property
def tx_sequence_users(self):
    return json.dumps(self.transaction_sequence, indent=4) if self.transaction_sequence else None
```

mythril/analysis/report.py
norhh: yes, but I wanted to do that after this gets merged, there are other things too which I skipped at the moment, I wanted to create a new issue for them to be tackled after the merge. (Like if an issue pops up  in creation tx etc)

JoranHonig: Sounds reasonable :+1:

mythril/laser/ethereum/instructions.py
JoranHonig: With my pr #1025 we might not have this problem anymore

mythril/laser/ethereum/svm.py
JoranHonig: Ok, then this is okay

mythril/laser/ethereum/svm.py
JoranHonig: But depending on the trace through the constructor this account can be different

JoranHonig: Additionally this introduces some record of accounts using dicts while we already have a datamodel for account's

JoranHonig: How about we keep track of each transactions' initial and final world state. Then, when serializing you just take the state of the world before the first transaction that you are interested in. 
This is rather simple to add, whereas the current approach adds more complexity to the svm

wdyt

mythril/analysis/solver.py
```diff
@@ -83,19 +90,19 @@ def get_transaction_sequence(global_state, constraints):
 
     transaction_sequence = global_state.world_state.transaction_sequence
 
-    # gaslimit & gasprice don't exist yet
     tx_template = {
```
JoranHonig: can we remove this initialization. There is only one use, so it does not make much sense to initialize this here

mythril/laser/ethereum/svm.py
JoranHonig: it also seems that the transaction serialization is dependent on the structure of the account dicts, this dependency is not apparent and could easily break in the future if we would change or add variables

mythril/laser/ethereum/svm.py
```diff
@@ -114,6 +115,21 @@ def accounts(self) -> Dict[str, Account]:
...
+        initial_state["accounts"] = {}  # This variable persists for all world states.
+        for address, account in accounts.items():
+            if ignore_addr and address == "0x" + "0" * 40:
+                continue
+            initial_state["accounts"][address] = {
+                "nounce": account.nonce,
+                "balance": "<ARBITRARY_BALANCE>",
+                "code": account.code.bytecode,
+                "storage": {},
```
JoranHonig: An initial state doesn't necessarily not have any storage

JoranHonig: This would result in incomplete transaction traces right?

mythril/laser/ethereum/svm.py
norhh: `How about we keep track of each transactions' initial and final world state. Then, when serializing you just take the state of the world before the first transaction that you are interested in` But most of the data might redundant right?
`But depending on the trace through the constructor this account can be different`
It's a dict of all accounts(it seems to be a bad variable name). 
`this dependency is not apparent and could easily break in the future if we would change or add variables` If we change variables then yes, else no, I think that should be fine as the members are allowed access there, some tests might help here WDYT?

norhh: It seems we already store those :man_facepalming:.

mythril/laser/ethereum/svm.py
```diff
@@ -1,9 +1,9 @@
 """This module implements the main symbolic execution engine."""
 import logging
 from collections import defaultdict
-from copy import copy
+from copy import copy, deepcopy
```
JoranHonig: These are now unused imports right?

mythril/analysis/solver.py
```diff
@@ -109,29 +108,49 @@ def get_transaction_sequence(global_state, constraints):
...
-        concrete_transactions[tx_id]["caller"] = "0x" + (
+        value = model.eval(transaction.call_value.raw, model_completion=True).as_long()
+        concrete_transaction["value"] = "0x%x" % value
+        caller = "0x" + (
             "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
         ).zfill(40)
 
-    return concrete_transactions
+        concrete_transaction["origin"] = caller
+        concrete_transaction["address"] = "%s" % transaction.callee_account.address
```
JoranHonig: why do the caller concretization and not

mythril/analysis/solver.py
```diff
@@ -109,29 +108,49 @@ def get_transaction_sequence(global_state, constraints):
...
-        )
-        concrete_transactions[tx_id]["caller"] = "0x" + (
+        value = model.eval(transaction.call_value.raw, model_completion=True).as_long()
+        concrete_transaction["value"] = "0x%x" % value
+        caller = "0x" + (
             "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
         ).zfill(40)
 
-    return concrete_transactions
+        concrete_transaction["origin"] = caller
```
JoranHonig: The caller and origin are not the same

mythril/analysis/solver.py
```diff
@@ -109,29 +108,49 @@ def get_transaction_sequence(global_state, constraints):
...
+                initial_state = transaction.world_state.accounts
+                break
+
+    states = {}
+    for address, account in initial_state.items():
+        if address == "0x" + 40 * "0" and len(account.code.bytecode) > 0:
+            continue
+        data = dict()  # type: Dict[str, Union[int, str]]
+        data["nonce"] = account.nonce
+        data["balance"] = account.balance
```
JoranHonig:
```suggestion
        data["balance"] = account.balance()
```

JoranHonig: balance is written twice here

mythril/analysis/solver.py
```diff
@@ -109,29 +108,49 @@ def get_transaction_sequence(global_state, constraints):
...
+    if len(transactions) > 0:
+        initial_state = transactions[0].world_state.accounts
+    else:
+        for transaction in transaction_sequence:
+            if transaction.id == "1":
+                initial_state = transaction.world_state.accounts
+                break
+
+    states = {}
+    for address, account in initial_state.items():
```
JoranHonig: initial_state can be uninitialized

norhh: There will always be a tx_id=“1” right?

mythril/analysis/solver.py
norhh: I still didn’t merge it with develop

mythril/analysis/solver.py
norhh: We don’t have an origin, so I put in caller there

mythril/analysis/solver.py
JoranHonig: Right now, maybe. But we never document/test that this is always the case, so any refactor can change this by accident

norhh: Now it should be fine

mythril/analysis/templates/report_as_text.jinja2
```diff
@@ -18,11 +18,11 @@ In file: {{ issue.filename }}:{{ issue.lineno }}
 
 --------------------
 {% endif %}
-{% if verbose and issue.debug %}
+{% if verbose and issue.tx_sequence %}
 --------------------
 Transaction Sequence:
 
-{{ issue.debug }}
+{{ issue.tx_sequence }}
```
JoranHonig: Shouldn't this be something like:
```suggestion
{{ issue.tx_sequence_users }}
```

norhh: jinja takesin the issue.as_dict(), where this is tx_sequence

mythril/analysis/solver.py
```diff
@@ -109,29 +108,42 @@ def get_transaction_sequence(global_state, constraints):
...
+        caller = "0x" + (
             "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
         ).zfill(40)
 
-    return concrete_transactions
+        concrete_transaction["origin"] = caller
+        concrete_transaction["address"] = "%s" % transaction.callee_account.address
+        concrete_transactions.append(concrete_transaction)
+        min_price_dict[caller] = min_price_dict.get(caller, 0) + value
+    initial_state = transaction_sequence[0].world_state.accounts
```
JoranHonig: If the creation transactions are not included, shouldn't we take the world state after the creation and not before?

norhh: It ran perfectly in my tests, there seems to be a copy stuff not happening somewhere in worldstate.

b-mueller:
Looking good?

mythril/analysis/solver.py
```diff
@@ -83,55 +90,103 @@ def get_transaction_sequence(global_state, constraints):
...
+    # Get concrete values from transaction
+    address = transaction.callee_account.address
+    value = model.eval(transaction.call_value.raw, model_completion=True).as_long()
+    caller = "0x" + (
+        "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
+    ).zfill(40)
+
+    # TODO: Do this at Laser
+    if isinstance(transaction, ContractCreationTransaction):
+        calldata = transaction.code.bytecode
```
JoranHonig:
```suggestion
        input = transaction.code.bytecode
```

mythril/analysis/solver.py
```diff
@@ -83,55 +90,103 @@ def get_transaction_sequence(global_state, constraints):
...
+    else:
+        calldata = "".join(
```
JoranHonig:
```suggestion
        input = "".join(
```

mythril/analysis/solver.py
```diff
@@ -83,55 +90,103 @@ def get_transaction_sequence(global_state, constraints):
...
+                hex(b)[2:] if len(hex(b)) % 2 == 0 else "0" + hex(b)[2:]
+                for b in transaction.call_data.concrete(model)
+            ]
+        )
 
-            minimize.append(transaction.call_data.calldatasize)
-            minimize.append(transaction.call_value)
+    # Create concrete transaction dict
+    concrete_transaction = dict()  # type: Dict[str, str]
+    concrete_transaction["input"] = "0x" + calldata
```
JoranHonig:
```suggestion
    concrete_transaction["input"] = "0x" + input_
```

mythril/analysis/solver.py
```diff
@@ -83,55 +90,103 @@ def get_transaction_sequence(global_state, constraints):
...
+    """ Gets a concrete transaction from a transaction and z3 model"""
+    # Get concrete values from transaction
+    address = transaction.callee_account.address
+    value = model.eval(transaction.call_value.raw, model_completion=True).as_long()
+    caller = "0x" + (
+        "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
+    ).zfill(40)
+
+    # TODO: Do this at Laser
+    if isinstance(transaction, ContractCreationTransaction):
```
JoranHonig:
```suggestion
    if isinstance(transaction, ContractCreationTransaction):
        address = ""
```

mythril/analysis/solver.py
```diff
@@ -83,55 +90,103 @@ def get_transaction_sequence(global_state, constraints):
...
+def _get_concrete_transaction(model: z3.Model, transaction: BaseTransaction):
+    """ Gets a concrete transaction from a transaction and z3 model"""
+    # Get concrete values from transaction
+    address = transaction.callee_account.address
+    value = model.eval(transaction.call_value.raw, model_completion=True).as_long()
+    caller = "0x" + (
+        "%x" % model.eval(transaction.caller.raw, model_completion=True).as_long()
+    ).zfill(40)
+
+    # TODO: Do this at Laser
```
JoranHonig:
```suggestion
```

PR #1017, opened 2019-05-01
Dockerfile improvement

yxliang01:
1. Run Mythril under normal user `mythril`, not need at all to run as `root`
2. Support preinstalling custom versions of `solc` done by setting build time argument `SOLC`.


JoranHonig:
@birdofpreyru Could you check this? You probably have a bit more experience with docker configurations

birdofpreyru:
@JoranHonig purely from pov of Dockerfile syntax, it looks fine.
Whether the actual commands achieve the goals without any side effects, I don't know. I guess, you just try to build and operate resulting image?

yxliang01:
@birdofpreyru @JoranHonig For the reference of the new command added, you can see https://github.com/ethereum/py-solc . In terms of adding new user and switch to that user I think is straightforward.

norhh:
@yxliang01 I think this change should be reflected in https://github.com/ConsenSys/mythril-classic/blob/develop/docker_build_and_deploy.sh . Due to the ARG SOLC part

yxliang01:
@norhh Originally, the Mythril image on Dockerhub comes with the latest solc from Ethereum PPA. This PR keeps that as the current version of [py-solc](https://github.com/ethereum/py-solc) doesn't support solc 0.5 . As a result, I don't think https://github.com/ConsenSys/mythril-classic/blob/develop/docker_build_and_deploy.sh needs to be changed. I can change if you want to include some solc 0.4.x into the Dockerhub image.

norhh:
Oh! I probably thought that the argument was necessary(which doesn't seem to be the case), as without any argument It seems to fail in my pc.
dockerfile

Dockerfile
```diff
@@ -36,5 +39,11 @@ COPY . /opt/mythril
...
   && python setup.py install
 
-COPY ./mythril/support/assets/signatures.db /root/.mythril/signatures.db
+RUN useradd -m mythril
+USER mythril
+WORKDIR /home/mythril
+
+RUN ( [ ! -z "${SOLC}" ] && set -e && for ver in $SOLC; do python -m solc.install v${ver}; done ) || true
+
+COPY --chown=mythril:mythril ./mythril/support/assets/signatures.db /root/.mythril/signatures.db
```
norhh: --chown doesn't work for docker 17.05 which is being used, so can you use a different syntax there?

yxliang01: Hmm, I would say this option has been available quite a while already. Maybe it's a good idea to update docker? This can be replaced with either 3 more lines of code (switch to `root`, `RUN chown`, switch back to `mythril`) or move this line to right after line 42 and `RUN chown`. Both solution makes dockerfile longer though and the second solution requires `solc` to be reinstalled after changing `signatures.db`. @norhh which one you prefer?

norhh: I think it is possible to do it in one shot with `RUN sudo chown`

yxliang01: @norhh To do so, the container requires `sudo` to be installed and user `mythril` is added to `sudo` group. I would argue that it's not really better than the suggested solutions as the user is given unnecessary privilege.

norhh: yes, then we can go with the other option of RUN chown.

yxliang01: @norhh which `RUN chown` solution? There are two I proposed.

norhh: yes

yxliang01: No @norhh , there are two options. You need to pick one.

"This can be replaced with either 3 more lines of code (switch to root, RUN chown, switch back to mythril) or move this line to right after line 42 and RUN chown"

norhh: The 2nd one.

yxliang01: Oops, just now I review this line again. I realized that the `signatures.db` is supposed to be copied to `mythril`'s directory and `chown` wasn't really necessary.

Issue #734, opened 2018-11-20
Incomplete analysis on contact sample


b-mueller:
## Description

Mythril returns a false negative for a contract sample that contains an Ether withdrawal flaw. The reason appears to be that symbolic execution does not complete. With max-transaction-count set to 5, execution finishes very quickly achieves only 72.52% coverage.

Code:

```solidity
contract Mapping {
    struct Participant {
        uint total_paid;
        uint payout;
    }

    mapping(address => Participant) participants;


    function deposit(address beneficiary) public payable {
        var b = participants[beneficiary];
        b.total_paid += msg.value;
        b.payout += msg.value / 2;
    }

    function remove(uint amount, address beneficiary) public {
        var b = participants[beneficiary];
        require(b.payout > 0);
        require(b.total_paid > 0);
        b.payout -= amount;
    }

    function withdraw() public {
        var p = participants[msg.sender];
        require(p.payout > 1 ether);
        msg.sender.transfer(p.payout);
        p.payout = 0;
    }

}
```

## How to Reproduce

```
$ myth -mether_thief --strategy bfs -t4 -x mapping.sol -v1 
root[18309] INFO Using INFURA for RPC queries
root[18309] INFO LASER EVM initialized with dynamic loader: <mythril.support.loader.DynLoader object at 0x10ee8dd30>
root[18309] INFO Found 0 detection modules
root[18309] INFO Starting contract creation transaction
root[18309] INFO Finished contract creation, found 1 open states
root[18309] INFO Starting message call transaction, iteration: 0
root[18309] INFO Number of new instructions covered in tx 0: 256
root[18309] INFO Starting message call transaction, iteration: 1
root[18309] INFO Number of new instructions covered in tx 1: 0
root[18309] INFO Starting message call transaction, iteration: 2
root[18309] INFO Number of new instructions covered in tx 2: 0
root[18309] INFO Starting message call transaction, iteration: 3
root[18309] INFO Number of new instructions covered in tx 3: 0
root[18309] INFO Finished symbolic execution
root[18309] INFO 83 nodes, 82 edges, 1055 total states
[...]

PR #1024, opened 2019-05-06
Support keccak func for concrete data

norhh:
related to #734, one more bug needs to be fixed in it.


JoranHonig:
Why didn't you replace the keccak function manager with BitVecFuncs, and solve the problem that way?

norhh:
I happen to remove 2 if conditions and reindent, and thats the majority of it :smile:(which made it this big) . I am currently adding that change of replacing BitVecFuncs to handle keccak + replacing storage with hash of expr than a string.(which had some serious problems)

norhh:
closing this as #1030 is more complete version of this.

PR #1025, opened 2019-05-06
Balance modelling and symbolic sender variables

JoranHonig:
This pr implements three elements
 - Balance modelling
 - LaserEVM account interface refactor
   - This is necessary as it allows us to control the accounts that are added to a worldstate
   - Also, the new interface should be more intuitive
 - Symbolic senders

This is a more correct model than what we currently use. As a consequence of this pr we should refactor the ethersend and thief modules to work with this precise balance model.


mythril/laser/ethereum/state/account.py
```diff
@@ -91,36 +92,52 @@ def __init__(
...
         :param balance:
         """
-        self.balance = balance
+        balance = (
+            symbol_factory.BitVecVal(balance, 256)
+            if isinstance(balance, int)
+            else balance
+        )
+        if self._balances is None:
+            assert False
```
norhh: something like `assert self._balances is not None` might be better here.

mythril/laser/ethereum/state/account.py
```diff
@@ -91,36 +92,52 @@ def __init__(
...
-        self.balance += balance
+        balance = (
+            symbol_factory.BitVecVal(balance, 256)
+            if isinstance(balance, int)
+            else balance
+        )
+        self._balances[self.address] += balance
 
     @property
     def as_dict(self) -> Dict:
```
norhh: the parentheses of self.balance are missing here.
it should be `"balance": self.balance()`

mythril/laser/ethereum/state/world_state.py
```diff
@@ -22,18 +24,29 @@ def __init__(
         :param transaction_sequence:
         :param annotations:
         """
-        self.accounts = {}  # type: Dict[str, Account]
+        self._accounts = {}  # type: Dict[int, Account]
+        self.balance = Array("balance", 256, 256)
```
norhh: Any reason why this is singular? (self.balance over self.balances)

mythril/laser/ethereum/state/world_state.py
```diff
@@ -22,18 +24,29 @@ def __init__(
         :param transaction_sequence:
```
norhh: The docstrings for the new variables.

mythril/laser/ethereum/transaction/symbolic.py
```diff
@@ -69,12 +78,15 @@ def execute_contract_creation(
     :return:
     """
     # TODO: Resolve circular import between .transaction and ..svm to import LaserEVM here
-    open_states = laser_evm.open_states[:]
+    # open_states = laser_evm.open_states[:]
```
norhh: Can this be removed?

mythril/laser/smt/bool.py
```diff
@@ -90,15 +90,18 @@ def And(*args: Union[Bool, bool]) -> Bool:
...
 
     :param a:
     :param b:
     :return:
     """
-    union = a.annotations + b.annotations
-    return Bool(z3.Or(a.raw, b.raw), annotations=union)
+    union = []
+    args_list = [arg if isinstance(arg, Bool) else Bool(arg) for arg in args]
+    for arg in args_list:
```
norhh: We can also combine all the 3 lines to `union = [arg.annotations for arg in args_list]` which might be a 1 liner

mythril/laser/ethereum/svm.py
```diff
@@ -106,35 +98,44 @@ def __init__(
...
-            log.info("Starting message call transaction to {}".format(main_address))
-            self._execute_transactions(main_address)
+        if pre_configuration_mode:
+            self.open_states = [world_state]
+            log.info("Starting message call transaction to {}".format(target_address))
+            self._execute_transactions(symbol_factory.BitVecVal(target_address, 256))
 
-        elif creation_code:
+        elif scratch_mode:
+            # self.open_states = [WorldState()]
```
norhh: any reason why this is commented?

mythril/laser/ethereum/state/account.py
JoranHonig: This was probably here like this to debug

JoranHonig: Fixed now 👍

mythril/laser/ethereum/state/account.py
JoranHonig: Good catch

mythril/laser/ethereum/state/world_state.py
JoranHonig: not really, I changed it to be plural 👍

mythril/laser/ethereum/svm.py
JoranHonig: Yeah, execute contract creation set's up the global state

mythril/laser/ethereum/transaction/symbolic.py
JoranHonig: yes

mythril/laser/ethereum/svm.py
norhh: I mean this can be removed right?

JoranHonig: it is 😅

JoranHonig:
@norhh is this good to go like this?

PR #1027, opened 2019-05-06
Change timeout logic

b-mueller:
Changes introduced:

- Increase default execution timeout to 24h (i.e. usually fully execute two tx-count transactions)
- When execution-timeout is set, instead of dividing time between transactions, attempt to complete execution of each tx until the timeout.


norhh:
This shouldn't cause any difference in tyro right? Also a print message to stdout saying `analysis is in progress`(and warn them about the execution timeout) might be some help, as some users raised issues saying that mythril froze, if the default is switched to 24hrs then the users would definitely assume thats the case.

JoranHonig:
We'd need to adapt tyro configurations

b-mueller:
> This shouldn't cause any difference in tyro right?

The API behavior shouldn't change because we always provide timeouts. @JoranHonig I don't think we need to change anything in Tyro?

b-mueller:
> as some users raised issues saying that mythril froze

If users are concerned about analysis progress they can add the -v4 flag. Analysis is expected to take long for anything >-t1.

JoranHonig:
@b-mueller timeouts would behave differently right? each transaction would be allowed to run for the entire execution-timeout, whereas now the execution-timeout sets a bound on the global time taken for the analysis.
So the analysis can take N times as long, where N is the amount of transactions we tell mythril to do

JoranHonig:
I'm not sure how this got in there but there seem to be both `self.time_handler` and `self.time` being used for timout purposes. The time handler probably also needs to be changed to allow this pr to work.

However I think we should go with either using the time handler or the time variable, not both

b-mueller:
> the execution-timeout sets a bound on the global time taken for the analysis.

That's how it works in this PR, the timeout applies to the entire analysis. At least that's what I intended.

b-mueller:
> I'm not sure how this got in there but there seem to be both self.time_handler and self.time being used for timout purposes. The time handler probably also needs to be changed to allow this pr to work

On a quick glance it seems that time handler doesn't need changing, even though I'm not sure if the way it's used makes sense. Currently we use it to dynamically set the solver timeout to `time_remaining() - 500`. I'd rather not have the solver use 52 minutes of a 1 hour analysis if it gets stuck on a difficult task.

Why not just use a fixed timeout?

norhh:
@b-mueller , It's mostly max(100s, time_remaining()-500) for solving, that was done for the sake of making execution-timeout flag proper. We couldn't kill mythril when it is solving, as it results in some segfaults, so had to use this way.

b-mueller:
Is there anything that prevents merging this? Also, does it conflict with #958?

Issue #866, opened 2019-01-04
Refactor storage


JoranHonig:
## Description 
Recent additions (state annotations, variable annotations) permit an overall better approach to how storage should be dealt with. This issue is meant to track that improvement.

JoranHonig:
This seems to be a done issue -> closing 🙌

PR #1030, opened 2019-05-11
Storage Refactor

### LINKED ISSUE #734: Incomplete analysis on contact sample

norhh:
fixes #734 , #866 
TODO for latter: Support for other functions like ECRECOVER, ....


mythril/laser/ethereum/transaction/symbolic.py
```diff
@@ -73,7 +73,7 @@ def execute_contract_creation(
     del laser_evm.open_states[:]
 
     new_account = laser_evm.world_state.create_account(
-        0, concrete_storage=True, dynamic_loader=None, creator=CREATOR_ADDRESS
+        0, concrete_storage=False, dynamic_loader=None, creator=CREATOR_ADDRESS
```
JoranHonig: Why make the storage symbolic?

mythril/laser/ethereum/keccak.py
```diff
@@ -16,7 +17,9 @@ def is_keccak(self, expression: Expression) -> bool:
         :param expression:
         :return:
         """
-        return str(expression) in self.keccak_expression_mapping.keys()
+        if not isinstance(expression, BitVecFunc):
+            return False
+        return expression.func_name == "keccak256"
```
JoranHonig: why keep the keccak function manager when we start using bitvecfuncs?

mythril/laser/ethereum/transaction/symbolic.py
norhh: That flag doesn't seem to make storage symbolic, it just assigns the storage slot to 0 before returning, if the slot doesn't exist. We should do something about it.

mythril/laser/ethereum/keccak.py
norhh: It seemed to be a nice, handy and extensible class to other functions.

mythril/laser/ethereum/transaction/symbolic.py
JoranHonig: It does. If concrete_storage is true and the key is not found, then it returns a symbol with value 0. Otherwise it creates a new unconstrained variable which is returned.
This is the difference between concretely initialized storage (all at 0) and symbolic storage right?

mythril/laser/ethereum/keccak.py
JoranHonig: What functionality does it add that BitVecFuncs do not already give?

norhh: save some lines of code + we can extend this to ECDSA.

mythril/laser/ethereum/transaction/symbolic.py
norhh: `Otherwise it creates a new unconstrained variable which is returned` it seems to assign the storage slot to 0 and then return the value.

mythril/laser/ethereum/keccak.py
JoranHonig: If these are useful helper functions for bitvecfuns then why not add them as such instead of maintaining this class?

mythril/laser/ethereum/transaction/symbolic.py
JoranHonig: Ah, that's a bug. It should put a new storage symbol there

JoranHonig: Still, there is no need for this change right?
Once the behaviour would be fixed, then this would break

mythril/laser/ethereum/keccak.py
norhh: My plan was to place few static classes in this file, so I left it off as it is.Will clean it up a bit.

b-mueller:
Is this still necessary and/or compatible after merging #1076?

norhh:
Yep, it is compatible, but i need to make a few changes to fix the merge conflicts now

PR #1032, opened 2019-05-13
ignore external calls to precompile contracts

### LINKED ISSUE #1012: External call
## Description
If there is an external call to a precompile then we probably don't want to report an issue at all.

Example contract for which we report a vulnerability: https://github.com/TokenMarketNet/smart-contracts/blob/master/contracts/test/TestSolidityAddressHash.sol

JoranHonig:
fixes #1012


mythril/analysis/modules/external_calls.py
```diff
@@ -104,6 +106,22 @@ def _analyze_state(state):
...
+    to = global_state.mstate.stack[-2]  # type: BitVec
+    try:
+        constraints = copy(global_state.mstate.constraints)
+        constraints += [
+            Or(
+                to < symbol_factory.BitVecVal(1, 256),
+                to > symbol_factory.BitVecVal(16, 256),
+            )
+        ]
+        solver.get_model(constraints)
```
norhh: we can restrict the try to here right?

PR #1034, opened 2019-05-14
Storage return symbol

JoranHonig:


norhh:
I think the commits have gotten messed up in here.

PR #1041, opened 2019-05-21
Add DOS detection module

b-mueller:
This module returns a low severity warning when a state change or call is executed in a loop. It correctly detects the following SWC test cases:

https://smartcontractsecurity.github.io/SWC-registry/docs/SWC-128#dos-addresssol
https://smartcontractsecurity.github.io/SWC-registry/docs/SWC-128#dos-numbersol
https://smartcontractsecurity.github.io/SWC-registry/docs/SWC-128#dos-simplesol

### DIFF
--- a/mythril/analysis/modules/dos.py
+++ b/mythril/analysis/modules/dos.py
@@ -1,0 +1,110 @@
+"""This module contains the detection code for insecure delegate call usage."""
+
+import logging
+from typing import List
+
+from mythril.analysis.swc_data import DOS_WITH_BLOCK_GAS_LIMIT
+from mythril.analysis.report import Issue
+from mythril.analysis.modules.base import DetectionModule
+from mythril.laser.ethereum.state.global_state import GlobalState
+from mythril.laser.ethereum.state.annotation import StateAnnotation
+from mythril.laser.ethereum import util
+
+log = logging.getLogger(__name__)
+
+
+class LoopAnnotation(StateAnnotation):
+    def __init__(self, jump_address: int) -> None:
+        self.jump_address = jump_address
+
+
+class DOS(DetectionModule):
+    """This module detects denial of service by exhausting gas."""
+
+    def __init__(self) -> None:
+        """"""
+        super().__init__(
+            name="DOS",
+            swc_id=DOS_WITH_BLOCK_GAS_LIMIT,
+            description="Check for DOS",
+            entrypoint="callback",
+            pre_hooks=["JUMPI", "CALL", "SSTORE"],
+        )
+
+        self._jumpdest_count = {}
+
+    def execute(self, state: GlobalState) -> None:
+        """
+
+        :param state:
+        :return:
+        """
+        log.debug("Executing module: DOS")
+
+        self._issues.extend(self._analyze_states(state))
+
+    def _analyze_states(self, state: GlobalState) -> List[Issue]:
+        """
+        :param state: the current state
+        :return: returns the issues for that corresponding state
+        """
+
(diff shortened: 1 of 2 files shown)


mythril/analysis/modules/dos.py
```diff
@@ -0,0 +1,111 @@
...
+
+        opcode = state.get_current_instruction()["opcode"]
+
+        if opcode == "JUMPI":
+
+            target = util.get_concrete_int(state.mstate.stack[-1])
+
+            try:
+                transaction = state.current_transaction
+                self._jumpdest_count[transaction]
```
norhh: we can use ` if transaction in self._jumpdest_count` in this case.

mythril/analysis/modules/dos.py
```diff
@@ -0,0 +1,111 @@
...
+
+                try:
+                    if self._jumpdest_count[transaction][target] > 3:
+                        state.annotate(
```
norhh: I think we might be able to reduce the number of annotate()'s here, by checking if it's 4 then annotate and ++, >4 then ignore, might save us some memory.

mythril/analysis/modules/dos.py
```diff
@@ -0,0 +1,110 @@
...
+
+            transaction = state.current_transaction
+            if state.current_transaction in self._jumpdest_count:
+
+                try:
+                    if self._jumpdest_count[transaction][target] == 4:
+                        state.annotate(
+                            LoopAnnotation(state.get_current_instruction()["address"])
+                        )
+                    else:
```
norhh: shouldn't an addition to jumpdest_count happen above?

mythril/analysis/modules/dos.py
```diff
@@ -0,0 +1,110 @@
...
+            if len(annotations) > 0:
+
+                operation = (
+                    "A storage modification" if opcode == "CALL" else "An external call"
+                )
+
+                description_head = (
+                    "Potential Denial-of-Service if block gas limit is reached."
+                )
+                description_tail = operation + " is executed in a loop."
```
norhh: string formatting might be cool here.

mythril/analysis/modules/dos.py
```diff
@@ -0,0 +1,109 @@
...
+        else:
+
+            annotations = cast(
+                List[LoopAnnotation], list(state.get_annotations(LoopAnnotation))
+            )
+
+            if len(annotations) > 0:
+
+                operation = (
+                    "A storage modification" if opcode == "CALL" else "An external call"
```
norhh: shouldn't it be reverse?

norhh:
This also detects the SSTORES after the loop right? as the annotations aren't being flushed, one way to deal with it is to check the count of those operation addresses in the loop too.

norhh:
I mean it also shows that in cases like these.
```


contract DosGas {

    address[] creditorAddresses;
    bool win = false;

    function emptyCreditors() public {
        if(creditorAddresses.length>1500) {
            creditorAddresses = new address[](0);
            win = true;
        }
    }

    function addCreditors() public returns (bool) {
        uint x=0;
        for(uint i=0;i<4;i++) {
        }
        creditorAddresses.push(msg.sender);
        return true;
    }

    function iWin() public view returns (bool) {
        return win;
    }

    function numberCreditors() public view returns (uint) {
        return creditorAddresses.length;
    }
}
```

mythril/analysis/modules/dos.py
b-mueller: Ammm actually yes :)

b-mueller:
> I mean it also shows that in cases like these.

Should work now I think...

PR #1042, opened 2019-05-22
Enhance delegate call

### LINKED ISSUE #974: Enhance delegatecall module detection capabilities
## Description
Currently delegatecall module just detects delegatecall(msg.data) in fallbacks, it should also be able to detect the other cases where a user defined argument is passed to call to a user defined contract.(and also other cases)

## Tests
https://github.com/SmartContractSecurity/SWC-registry/blob/master/test_cases/solidity/delegate_call_to_untrusted_callee/proxy/proxy.sol

norhh:
fixes #974 
Helps in fixing for these below contracts.
https://github.com/SmartContractSecurity/SWC-registry/tree/master/test_cases/solidity/delegate_call_to_untrusted_callee
@b-mueller should we also consider the case of forwarding a part of calldata to the delegatecall?, it wasn't mentioned in the swc registry.


b-mueller:
> @b-mueller should we also consider the case of forwarding a part of calldata to the delegatecall?

I think it's worth noting when calldata is forwarded (i.e. delegate call proxy pattern).

norhh:
would that come under SWC-112?

b-mueller:
Hmm, not sure if calldata being passed on needs to be an issue by itself. For the current implementation, it would be neat that if when creating issue is created, we would also add a note to the description if the calldata is passed on (just a thought tough, we can also leave that out for now).

norhh:
Gave it an attempt, due to the way memory is being modeled, we can just approximately detect the passing of calldata to delegatecall, this requires some changes in the way memory is handled, we can look into that later.

b-mueller:
Ok, it's not a priority so don't spend too much time on it :)

PR #1043, opened 2019-05-23
Refactor multiple_sends module

b-mueller:
Changes to multiple_send module:

- Store call offsets instead of call objects in the state annotation (to remove overhead)
- Instead of reporting a single issue listing all calls, create a separate issue for each call that follows a previous call. This results in more usable location information and passes the SWC test case [send_loop.sol](https://github.com/SmartContractSecurity/SWC-registry/blob/master/test_cases/solidity/call_best_practices/send_loop/send_loop.sol).

### DIFF
--- a/mythril/analysis/modules/multiple_sends.py
+++ b/mythril/analysis/modules/multiple_sends.py
@@ -17,11 +17,11 @@ log = logging.getLogger(__name__)
 
 class MultipleSendsAnnotation(StateAnnotation):
     def __init__(self) -> None:
-        self.calls = []  # type: List[Optional[Call]]
+        self.call_offsets = []  # type: List[int]
 
     def __copy__(self):
         result = MultipleSendsAnnotation()
-        result.calls = copy(self.calls)
+        result.call_offsets = copy(self.call_offsets)
         return result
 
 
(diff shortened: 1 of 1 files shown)


mythril/analysis/modules/multiple_sends.py
```diff
@@ -62,51 +60,41 @@ def _analyze_state(state: GlobalState):
...
 
     if instruction["opcode"] in ["CALL", "DELEGATECALL", "STATICCALL", "CALLCODE"]:
-        call = get_call_from_state(state)
-        if call:
-            calls += [call]
+        call_offsets.append(state.get_current_instruction()["address"])
 
     else:  # RETURN or STOP
-        if len(calls) > 1:
+        if len(call_offsets) > 1:
```
norhh: I think we can skip this line in python. In python if `a = []` then `a[1:]` is also `[]`

b-mueller: Wow, you're right :)

mythril/analysis/modules/multiple_sends.py
```diff
@@ -62,46 +60,35 @@ def _analyze_state(state: GlobalState):
...
-            for call in calls:
-                description_tail += "Offset: {}\n".format(
-                    call.state.get_current_instruction()["address"]
-                )
-
-            description_tail += (
-                "Try to isolate each external call into its own transaction,"
-                " as external calls can fail accidentally or deliberately.\n"
+            description_tail = (
+                "This call is executed after a previous call in the same transaction. "
```
norhh: Is the extra space at the end intended?

Issue #931, opened 2019-02-18
Integer overflow isn't being detected in BECToken.sol


norhh:
## Description
Bectoken's integer overflow issue isn't being detected

## How to Reproduce
```console
$ myth -x BECToken.sol
No Issues being detected
...
$
```

## Expected behavior
Integer overflow should be detected

## Environment
- Mythril version: develop
- Solidity compiler and version: `0.5.0`
- Python version: `3.6.3`
- OS and Version: Ubuntu 17.10

norhh:
It seems that Stop is reached at the depth of 50 where the issues are flushed

b-mueller:
@norhh I'm re-opening this as it appears that the overflow is no longer detected in 0.20.4. This is what I get:

```
$ myth -t1 -x BECToken.sol:BecToken -v4
mythril.support.signatures [INFO]: Using signature database at /Users/bernhardmueller/.mythril/signatures.db
mythril.analysis.security [INFO]: Found 0 detection modules
mythril.laser.ethereum.svm [INFO]: LASER EVM initialized with dynamic loader: <mythril.support.loader.DynLoader object at 0x110e3a518>
mythril.laser.ethereum.plugins.plugin_loader [INFO]: Loading plugin: <mythril.laser.ethereum.plugins.implementations.mutation_pruner.MutationPruner object at 0x111f86a58>
mythril.laser.ethereum.plugins.plugin_loader [INFO]: Loading plugin: <mythril.laser.ethereum.plugins.implementations.coverage.coverage_plugin.InstructionCoveragePlugin object at 0x111f86b70>
mythril.analysis.security [INFO]: Found 10 detection modules
mythril.analysis.security [INFO]: Found 10 detection modules
mythril.laser.ethereum.svm [INFO]: Starting contract creation transaction
mythril.laser.ethereum.svm [INFO]: Finished contract creation, found 1 open states
mythril.laser.ethereum.svm [INFO]: Starting message call transaction, iteration: 0, 1 initial states
mythril.analysis.modules.exceptions [INFO]: Exceptions module: found ASSERT_FAIL instruction 
mythril.analysis.modules.exceptions [INFO]: Exceptions module: found ASSERT_FAIL instruction 
mythril.analysis.modules.exceptions [INFO]: Exceptions module: found ASSERT_FAIL instruction 
mythril.laser.ethereum.plugins.implementations.coverage.coverage_plugin [INFO]: Number of new instructions covered in tx 0: 2144 
[...]

PR #1045, opened 2019-05-24
Use newer z3 version

b-mueller:


b-mueller:
Solves the issue where Z3 ignores execution timeout. Further helps saving #931.

norhh:
I think our release of z3 doesn't have a problem with BECToken.sol, Thats happens to be my regular testing contract these days and I haven't had a problem with it.

b-mueller:
> I think our release of z3 doesn't have a problem with BECToken.sol, That happens to be my regular testing contract these days and I haven't had a problem with it for that span

Really? That's weird, whenever I ran Mythril on BECToken.sol with any Solver timeout longer than a few seconds Z3 would get stuck on some task for several minutes. In never got it to get proper coverage or detect the MUL overflow within a reasonable time. With the new version, I detect it reliably with `-t1 --execution-timeout 90`.

Could be a Mac-related problem...

norhh:
The older z3 should be uninstalled, else that version would be used. That might be the reason.

Issue #1048, opened 2019-05-27
Mythril is slow while executing proxy.sol


norhh:
## Description
As @b-mueller mentioned in #1042 mythril is slow while executing proxy.sol which is a very small contract.

b-mueller:
It seems like we're entering a loop where the calldata is copied byte-per-byte until max-depth is reached. For reference here is the function:

```
  function forward(address callee, bytes _data) public {
    require(callee.delegatecall(_data));
  }
```

norhh:
@b-mueller we can close this right? Now it takes around 13s to finish

b-mueller:
Yep, can close long ago!

PR #1060, opened 2019-05-30
Bounded loops

b-mueller:
Adds a new search strategy called "bfs-bounded". When an address is repeatedly targeted by a JUMPI in path (`>JUMPDEST_LIMIT` times which is currently set to 4), this strategy ignores subsequent JUMPI instructions that target the same destination.

Solves #1048 for the most part (it's still quite slow on byte array copy because memory operations are slow, but with the number of iteration reduced the time Mythril spends on this should become reasonable). Should also help with BECToken and other samples that contain loops.


mythril/analysis/modules/dos.py
```diff
@@ -70,7 +70,7 @@ def _analyze_states(self, state: GlobalState) -> List[Issue]:
 
                 try:
                     self._jumpdest_count[transaction][target] += 1
-                    if self._jumpdest_count[transaction][target] == 4:
+                    if self._jumpdest_count[transaction][target] == 3:
```
norhh: Using JUMPDEST_LIMIT - 1 might be good here, what do you think?

b-mueller: Yeah, great idea!

mythril/laser/ethereum/strategy/custom.py
```diff
@@ -0,0 +1,55 @@
...
+    def get_strategic_global_state(self) -> GlobalState:
+        """
+        :return:
+        """
+
+        state = self.work_list.pop(0)
+
+        opcode = state.get_current_instruction()["opcode"]
+
+        if opcode == "JUMPI":
```
norhh: We can reduce the nesting here by returning if the `opcode != "JUMPI"`, less nesting looks cool and easy to read.

norhh:
This can actually be placed in a plugin right?

b-mueller:
I guess? Joran's suggestion was to put it into a strategy.

norhh:
Maintaining a global counter for all the states may not be a good idea right? Like if there are 3 states in worklist, all hit the loop, then self._jumpcount[tx][address] ++ happens globally right? are the transaction stacks deepcopied each time?

b-mueller:
Yes, the strategy counts jumpdests globally (per transaction) for the whole symbolic execution run, `_jumpcount` is a member of the `BFSBoundedLoopsStrategy` class instance, how is that related to transaction stacks? Sorry if I'm not following correctly :)

norhh:
What if there are 10 states in worklist which enter the same JUMPI condition(which is an if condition rather than a loop), In that case there might be a problem right?

b-mueller:
In what cases would that happen? The only case I can think of is complex if-statements if they compile in certain ways? 

```
if (a == 0 | a == 1 | a == 2 | a == 3 | a == 4) {

} else {

}
```

Would be interesting to have some concrete examples where it causes issues.

norhh:
```
...
if()
{}
else {}
// there will be 2 states here

for(i=0; i<num; i++){} 
// there will be 2*JUMPI_COUNT states here

if(condition) //some states above might get pruned here.
{} 
 ...
```
It might be uncommon but still might be possible :cry: .(uncommon but likely to be encountered given that we analyze lot's of contracts per day)

b-mueller:
Ok, I confirmed that this is exactly the case. With this strategy we can detect this:

```
pragma solidity ^0.5.0;


contract Lol {

	function lol(uint256 a) public {

		if ((a == 0) || (a == 1) || (a == 2) || (a == 3)) {
		   if (a == 3) {
		      selfdestruct(msg.sender);
		  	}
		 }
	} 
}
```

But not this:

```
pragma solidity ^0.5.0;

contract Lol {

	function lol(uint256 a) public {

		if ((a == 0) || (a == 1) || (a == 2) || (a == 3) || (a == 4)) {
		   if (a == 4) {
		      selfdestruct(msg.sender);
		  	}
		 }
	} 
}
```

Although, in this case, you could say that it is very similar to a loop iterating over `a`. I guess with this strategy we'll miss cases whenever the code searches for a specific value in a variable or some complex combination of conditions.

The most straightforward way to deal with this IMO is to select a proper bound (perhaps a bit higher than 4?) and accept that we can miss some paths of there's an unbounded amount of values, in which case fuzzing would be more appropriate (and this is something we could hand over to Harvey)?

JoranHonig:
I'd add a state annotation which tracks which instructions have been executed over a given path.
Additionally, I'd say this should be an extending strategy (similar to how the coverage strategy is implemented). 
That way we don't have to implement the same code all the time

mythril/analysis/modules/dos.py
```diff
@@ -4,6 +4,7 @@
 from typing import Dict, cast, List
 
 from mythril.analysis.swc_data import DOS_WITH_BLOCK_GAS_LIMIT
+from mythril.laser.ethereum.strategy.custom import JUMPDEST_LIMIT
```
JoranHonig: Analysis code should not be dependent on search strategies, keeping them decoupled makes maintenance & further development easier.

JoranHonig:
@norhh I think a search strategy is perfect for this use case

I'd use a plugin if we extend the behaviour of laser in some way ( for example with the coverage ).
A search strategy interface is limited, making it easier to implement and more simple ( I hope )

norhh:
something like this might fail
```


contract Lame {

    uint bata;

    function addCreditors(uint a) public returns (bool) {
        if(a == 0)
        {
            bata += 1;
        }
        else {
            bata -= 1;
        }
        // there will be 2 states here
        uint i;
        for(i=0; i<2; i++){
            a+=1;
            bata += a;
        } 

        if(a == 3) // some states might get pruned here
        {
            selfdestruct(msg.sender);
        } 
    }

}

```
`I'd add a state annotation which tracks which instructions have been executed over a given path`
That should do the work.

norhh:
@JoranHonig .I was saying something like the coverage strategy plugin ,We can either place it as a plugin or an extensible class, as long as we can have a cool flag for it.

b-mueller:
> something like this might fail
>   contract Lame {

Yeah, basically another variant where we'd get state explosion because `a`\`bata` are incremented in a loop and each value of a could potentially trigger a different path. My intuition is that Harvey could deal with this better? @wuestholz

mythril/analysis/modules/dos.py
norhh: Then let's have a constants file for the laser module.That might be cool.

JoranHonig:
symbolic execution shouldn't have a problem with this contract

wuestholz:
@b-mueller Yeah, Harvey deals with path/state explosion a bit more gracefully. The main reason is that it explores the different paths and states less systematically. This avoids getting stuck in one part of the state space by exploring only its surrounding states.

b-mueller:
> I'd add a state annotation which tracks which instructions have been executed over a given path

That sounds reasonable. Not sure if refactoring this to a plugin is better, I'll finish up the class fist :)

JoranHonig:
The difference between a plugin and a search strategy is that a plugin has an initialization phase where it is allowed to introduce hooks in the laser vm.
A search strategy doesn't introduce hooks, but is initialized with the work list & called each time the executor wants to know a state to work on.

I don't think a plugin is necessary for this case. The coverage based search strategy is based in a plugin because it is dependent on that plugin. I'm not sure whether this is the best way to go about indicating the dependency, but this is what I decided to add at that point.

b-mueller:
I rewrote this to use a StateAnnotation. Not sure how extending strategies are meant to be used / invoked?

mythril/analysis/modules/dos.py
```diff
@@ -70,7 +71,7 @@ def _analyze_states(self, state: GlobalState) -> List[Issue]:
 
                 try:
                     self._jumpdest_count[transaction][target] += 1
-                    if self._jumpdest_count[transaction][target] == 4:
+                    if self._jumpdest_count[transaction][target] == JUMPDEST_LIMIT - 1:
```
norhh: this would always go in for if conditions right? if JUMPDEST_LIMIT is 2

mythril/analysis/modules/dos.py
JoranHonig: There doesn't need to be a dependency here

norhh: The JUMPDEST_LIMIT is dependent in both areas right? so this should have a dependece with laser.

mythril/analysis/modules/dos.py
b-mueller: no always but if there's more than 1 condition. So probably 4 is a more reasonable limit...

mythril/analysis/modules/dos.py
b-mueller: Yeah, the idea here is to make sure that the limit doesn't matches the loop bound because otherwise the it will never be triggered. But we can remove the dependency.

b-mueller: Actually, now that the search strategy sets a `JumpdestCountAnnotation` this module could be simplified further by accessing that annotation instead of maintaining its own counter. However this would mean that the dependence remains.

b-mueller: No, that's actually BS, because a different search strategy could be used.

b-mueller:
@JoranHonig I don't see `CoverageStrategy` used anywhere in the code. How are extended strategies supposed to be selected? Do we still need to add CLI support for those?

b-mueller:
I refactored this into an extension similar to the coverage strategy and added a method to laser for loading it "on top of" its existing strategy. Also added a `--loop-bound` CLI argument with a default value of 4 which is passed through to SymExecWrapper.

Should be ready for review now...

mythril/laser/ethereum/strategy/extensions/bounded_loops.py
```diff
@@ -0,0 +1,81 @@
...
+        self.super_strategy = super_strategy
+        self.jumpdest_limit = args[0][0]
+
+        log.info(
+            "Loaded search strategy extension: Loop bounds (limit = {})".format(
+                self.jumpdest_limit
+            )
+        )
+
+        BasicSearchStrategy.__init__(
```
norhh: We can use `super()` here. It would be more readable as super means parent class

mythril/laser/ethereum/strategy/extensions/bounded_loops.py
```diff
@@ -0,0 +1,81 @@
...
+            self, super_strategy.work_list, super_strategy.max_depth
+        )
+
+    def get_strategic_global_state(self) -> GlobalState:
+        """
+        :return:
+        """
+
+        while 1:
```
norhh:
```suggestion
        while True:
```

mythril/laser/ethereum/strategy/extensions/bounded_loops.py
```diff
@@ -0,0 +1,81 @@
...
+                list(state.get_annotations(JumpdestCountAnnotation)),
+            )
+
+            if len(annotations) == 0:
+                annotation = JumpdestCountAnnotation()
+                state.annotate(annotation)
+            else:
+                annotation = annotations[0]
+
+            target = int(util.get_concrete_int(state.mstate.stack[-1]))
```
norhh: wouldn't there be a chance of a ValueError? sometimes jumpi can be arbitrary.

mythril/interfaces/cli.py
```diff
@@ -197,13 +197,20 @@ def create_parser(parser: argparse.ArgumentParser) -> None:
...
         default="bfs",
         help="Symbolic execution strategy",
     )
+    options.add_argument(
+        "-b",
+        "--loop-bound",
+        type=int,
+        default=4,
+        help="Bound loops at n iterations",
+        metavar="N",
```
norhh: Should we have metavar as N here? as that would be used for display of the arg.We can just skip it.

mythril/analysis/modules/dos.py
norhh: We should use that constant here right? Propogating that to here might be a problem.

b-mueller: Unfortunately there's currently no way to pass arguments to modules or access svm variables from it. I'd leave this as a hardcoded value for now. Later, we could have Mythril set a ModuleConfigAnnotation as a new feature?

mythril/interfaces/cli.py
b-mueller: I thought it would be cool to say "n iterations" and call the metavar N 😀 Can we leave it like that?

mythril/laser/ethereum/strategy/extensions/bounded_loops.py
b-mueller: I think it's TypeError right?

norhh: Oh! yeah.

mythril/laser/ethereum/strategy/extensions/bounded_loops.py
b-mueller: super() does not work here because of the way search strategies are passed to laser and initialized. See also [CoverageStrategy](https://github.com/ConsenSys/mythril/blob/develop/mythril/laser/ethereum/plugins/implementations/coverage/coverage_strategy.py).

PR #1062, opened 2019-05-31
Refactor of setup.py

JoranHonig:
This pr updates project names & other metadata throughout the repo
It also cleans up setup.py ( following some best practices )

I'd like to move the test logic to setup.py based commands as well. Anyone thoughts on that?
Additionally we might want to consider using some format like Pipfile or setup.cfg to store the metadata, this seems to be a cleaner way to go about this.


JoranHonig:
Not sure why the precompile check is in here also, but it should be in the external calls module anyway

setup.py
```diff
@@ -9,21 +9,82 @@
...
+    "persistent>=4.2.0",
+    "ethereum-input-decoder>=0.2.2",
+    "matplotlib",
+]
+
+TESTS_REQUIRE = ["mypy", "pytest>=3.6.0", "pytest_mock", "pytest-cov"]
+
+# What packages are optional?
+EXTRAS = {
+    # 'fancy feature': ['django'],
```
norhh: wait! what :scream_cat:

Issue #1036, opened 2019-05-15
Rewrite integer overflow module to use solver selectively (for discussion)


b-mueller:
## Description

Revert the logic of the integer module: Instead of solving ahead of time and tracking the overflowed variables to STOPs, use the solver selectively when a STOP instruction is reached. This could be done in several ways, e.g.:

1. Keep track of arithmetic operations, e.g. by building a list in a state annotation (save type, offset & constraints)
2. Taint list entries that get used in SSTORE or JUMPI (mark for overflow check)
3. When STOP is reached, use the solver to check overflows on any tainted arithmetic operations that were tainted (perhaps use some additional logic to reduce the list).

## Rationale

Hopefully this should significantly reduce the time spent on solving.

JoranHonig:
Yes that could definately work, and shouldn't be too big of a change to the module. Great idea!

PR #1067, opened 2019-06-06
Refactor integer module

b-mueller:
Introduces several changes to the integer module:

1. Use a single `OverflowUnderflowStateAnnotation` to track `OverflowUnderflowAnnotations`
2. Only add a new `OverflowUnderflowAnnotations` to the state annotations if the (potentially) overflowing state hasn't yet been added
3. Solve constraints only on transaction end (as proposed in #1036)

This speeds up analysis performance of BECToken by ~4x. Combined with a loop bound limit of 2 a full transaction runs for 40s on my Mac, and 1m15s with normal bfs.

However, while I haven't seen any regressions with `DISABLE_EFFECT_CHECK = True`, the MUL overflow is not not reported when `DISABLE_EFFECT_CHECK` is set to `False` (still need to investigate why).

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -332,9 +332,6 @@ class IntegerOverflowUnderflowModule(DetectionModule):
                 else:
                     constraints = state.mstate.constraints + [annotation.constraint]
 
-                if ostate.get_current_instruction()['address'] == 1587:
-                    logging.info(constraints)
-
                 log.info("Potential overflow: {} at {}".format(annotation.operator, ostate.get_current_instruction()['address']))
 
                 transaction_sequence = solver.get_transaction_sequence(
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -48,7 +48,7 @@ class OverUnderflowStateAnnotation(StateAnnotation):
 
     def __init__(self) -> None:
         self.overflowing_state_annotations = []  # type: List[OverUnderflowAnnotation]
-        self.ostates_seen = []  # type: List[GlobalState]
+        self.ostates_seen = set()  # type: List[GlobalState]
 
     def __copy__(self):
         new_annotation = OverUnderflowStateAnnotation()
@@ -88,8 +88,8 @@ class IntegerOverflowUnderflowModule(DetectionModule):
         Cache satisfiability of overflow constraints
         """
 
-        self._ostates_satisfiable = []  # type: List[GlobalState]
-        self._ostates_unsatisfiable = []  # type: List[GlobalState]
+        self._ostates_satisfiable = set()
+        self._ostates_unsatisfiable = set()
 
     def reset_module(self):
         """
@@ -98,8 +98,8 @@ class IntegerOverflowUnderflowModule(DetectionModule):
         """
         super().reset_module()
         self._overflow_cache = {}
-        self._ostates_satisfiable = []
-        self._ostates_unsatisfiable = []
+        self._ostates_satisfiable = set()
+        self._ostates_unsatisfiable = set()
 
     def _execute(self, state: GlobalState) -> None:
         """Executes analysis module for integer underflow and integer overflow.
@@ -232,7 +232,7 @@ class IntegerOverflowUnderflowModule(DetectionModule):
                 continue
 
             state_annotation.overflowing_state_annotations.append(annotation)
-            state_annotation.ostates_seen.append(annotation.overflowing_state)
+            state_annotation.ostates_seen.add(annotation.overflowing_state)
 
     @staticmethod
     def _handle_jumpi(state):
@@ -250,7 +250,7 @@ class IntegerOverflowUnderflowModule(DetectionModule):
                 continue
 
             state_annotation.overflowing_state_annotations.append(annotation)
-            state_annotation.ostates_seen.append(annotation.overflowing_state)
+            state_annotation.ostates_seen.add(annotation.overflowing_state)
 
     @staticmethod
     def _handle_return(state: GlobalState) -> None:
@@ -258,7 +258,7 @@ class IntegerOverflowUnderflowModule(DetectionModule):
         Adds all the annotations into the state which correspond to the
         locations in the memory returned by RETURN opcode.
         :param state: The Global State
-
+        """
 
         stack = state.mstate.stack
         try:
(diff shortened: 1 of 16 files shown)


b-mueller:
Going by the debug outputs, with `EFFECT_CHECKS` the potential `MUL1` overflow at offset 1587 is checked during multiple STOP hooks but `state.mstate.constraints + [annotation.constraint]` can never be satisfied, which I find suprising. There must be a thinking error somewhere.
 
```
mythril.analysis.modules.integer [INFO]: Potential overflow: multiplication at 1587 
mythril.analysis.modules.integer [INFO]: UNSAT 
ythril.analysis.modules.integer [INFO]: Potential overflow: multiplication at 1587 
mythril.analysis.modules.integer [INFO]: UNSAT 
```

norhh:
I looked at it a while back too.How long does the constraint solving take place?

b-mueller:
The solver timeout is set to 10 seconds which should be plenty. I tried upping it to 50 second just to be sure but it didn't make a difference.

b-mueller:
Another things to note is that the small samples from SWC all work fine even with `DISABLE_EFFECT_CHECK = False`.

norhh:
`The solver timeout is set to 10 seconds which should be plenty. I tried upping it to 50 second just to be sure but it didn't make a difference.` Now I remember adding extra zeros to timeout and got no result then.
Previously with EFFECT_CHECK it took 17min in my pc. Currently how much time did it take in your PC. If it's lower then it might be easier for us to fix this time.

b-mueller:
I don't think that insufficient solver time is the reason, I think `state.constraint` at the STOP or RETURN ends up being unsolvable for other reasons.

b-mueller:
Analysis time on my Mac is 1m28s for:

`$MYTH --strategy dfs -t1  -xc `6080604052600060036.....`

b-mueller:
With unboundsed loops and `DISABLE_EFFECT_CHECK = True`:

```
$ time $MYTH --strategy dfs -t1 -xc [bytecode]
 ==== Integer Overflow ====
SWC ID: 101
Severity: High
Contract: MAIN
Function name: batchTransfer(address[],uint256)
PC address: 659
Estimated Gas Usage: 29996 - 118377
The binary addition can overflow.
The operands of the addition operation are not sufficiently constrained. The addition could therefore result in an integer overflow. Prevent the overflow by checking inputs or ensure sure that the overflow is caught by an assertion.
--------------------

==== Integer Overflow ====
SWC ID: 101
Severity: High
Contract: MAIN
Function name: batchTransfer(address[],uint256)
PC address: 1587
Estimated Gas Usage: 29996 - 118377
The binary multiplication can overflow.
The operands of the multiplication operation are not sufficiently constrained. The multiplication could therefore result in an integer overflow. Prevent the overflow by checking inputs or ensure sure that the overflow is caught by an assertion.
--------------------

==== Integer Overflow ====
SWC ID: 101
Severity: High
Contract: MAIN
Function name: batchTransfer(address[],uint256)
PC address: 4219
Estimated Gas Usage: 29996 - 118377
The binary addition can overflow.
The operands of the addition operation are not sufficiently constrained. The addition could therefore result in an integer overflow. Prevent the overflow by checking inputs or ensure sure that the overflow is caught by an assertion.
--------------------


real	1m43.602s
user	1m41.038s
sys	0m1.052s
```


With unboundsed loops and `DISABLE_EFFECT_CHECK = False`:

```
$ time $MYTH --strategy dfs -t1 -xc [bytecode]

real	1m17.024s
user	1m15.015s
sys	0m0.979s
```

UUUargh 😥

b-mueller:
The reason for the `MUL` not being reported is that apparently non of the transaction end states is reachable. I created an [experimental branch](https://github.com/ConsenSys/mythril/tree/integer_experimental) that solves the constraints separately with additional logging, the output for all transaction ends is similar to:

```
mythril.analysis.modules.integer [INFO]: Transaction end with RETURN 
root [INFO]: Number of potentially overflowing states: 5
mythril.analysis.modules.integer [INFO]: Potential overflow: multiplication at 1587 
root [INFO]: get_model(state.mstate.constraints) failed (end state is unreachable)
root [INFO]: get_model([annotation.constraint]) SUCCEEDED (integer can overflow)
mythril.analysis.modules.integer [INFO]: Solving combined constraints failed 
mythril.analysis.modules.integer [INFO]: Transaction end with RETURN 
root [INFO]: Number of potentially overflowing states: 5
mythril.analysis.modules.integer [INFO]: Potential overflow: multiplication at 1587 
root [INFO]: get_model(state.mstate.constraints) failed (end state is unreachable)
root [INFO]: get_model([annotation.constraint]) SUCCEEDED (integer can overflow)
mythril.analysis.modules.integer [INFO]: Solving combined constraints failed 
mythril.analysis.modules.integer [INFO]: Transaction end with RETURN 
```

I think there might be some fundamental problem that may not be related specifically to the module?

b-mueller:
Btw I'm not sure I understand the code in `_handle_return`. Why do those state annotations need to be created? If a value in memory is tainted with an `OverUnderflowAnnotation`, shouldn't everything work fine without any extra logic?

mythril/analysis/modules/integer.py
```diff
@@ -72,8 +77,19 @@ def __init__(self) -> None:
...
+        """
+
         self._overflow_cache = {}  # type: Dict[int, bool]
-        self._underflow_cache = {}  # type: Dict[int, bool]
+
+        """
+        Cache satisfiability of overflow constraints
+        """
+
+        self._ostates_satisfiable = []  # type: List[GlobalState]
```
norhh: A set might be perfect here. what do you think?

mythril/analysis/modules/integer.py
```diff
@@ -256,45 +282,62 @@ def _handle_return(state: GlobalState) -> None:
             offset, length = get_concrete_int(stack[-1]), get_concrete_int(stack[-2])
         except TypeError:
             return
+
+        """ TODO: Rewrite this
```
norhh: We can't comment this at the moment, as this would cause false negatives.

norhh: commenting this might cause false negatives like in
```
function func(uint a, uint b) {
    return a*b;
} 
```
And i think there is 1 swc sample which has this.

mythril/analysis/modules/integer.py
```diff
@@ -256,45 +282,62 @@ def _handle_return(state: GlobalState) -> None:
...
+
+        for annotation in annotations:
 
             ostate = annotation.overflowing_state
-            address = _get_address_from_state(ostate)
 
-            if annotation.operator == "subtraction" and self._underflow_cache.get(
-                address, False
-            ):
+            if ostate in self._ostates_unsatisfiable:
```
norhh: cache is being used, but there isn't any check in here. We can add that to check.

mythril/analysis/modules/integer.py
```diff
@@ -72,8 +77,19 @@ def __init__(self) -> None:
             entrypoint="callback",
             pre_hooks=["ADD", "MUL", "EXP", "SUB", "SSTORE", "JUMPI", "STOP", "RETURN"],
         )
+
+        """
+        Cache addresses for which overflows already have been detected.
```
norhh: As we merged overflow cache and underflow cache, we can underflow too in here.

PR #1078, opened 2019-06-14
Define ENV variable to suspend regressions reporting for some testcases

s0b0lev:


.circleci/config.yml
```diff
@@ -111,7 +111,8 @@ jobs:
                 --output-dir /opt/edelweiss \
                 --plugin-dir /opt/mythril \
                 --s3 \
-                --circle-ci CircleCI/mythril.csv
+                --circle-ci CircleCI/mythril.csv\
+                --ignore-regressions $IGNORE_REGRESSIONS
```
b-mueller: Change to $IGNORE_FALSE_POSITIVES?

.circleci/config.yml
```diff
@@ -111,7 +111,8 @@ jobs:
                 --output-dir /opt/edelweiss \
                 --plugin-dir /opt/mythril \
                 --s3 \
-                --circle-ci CircleCI/mythril.csv
+                --circle-ci CircleCI/mythril.csv\
```
b-mueller: This file is hosted on S3 right?

s0b0lev: Yep, right

.circleci/config.yml
s0b0lev: Changed

PR #1079, opened 2019-06-14
Add caller constraints on all transactions in tx sequence

b-mueller:
In the delegate call module, a false positive was caused by the fact that a delegate call to `ATTACKER_ADDRESS` can be executed in tx2 the owner first sets the callee to `ATTACKER_ADDRESS` in tx1.

The same issue affected the calls, ether_thief and suicide modules. For example, Mythril would report an issue for the following:

```
pragma solidity ^0.5.0;

contract Missing{
    address payable private owner;

    modifier onlyowner {
        require(msg.sender==owner);
        _;
    }

    // The name of the constructor should be Missing
    // Anyone can call the IamMissing once the contract is deployed
    constructor()
        public
    {
        owner = msg.sender;
    }

    function setowner(address payable newOwner)
       public onlyowner
    {
       owner = newOwner; 
    }


    function withdraw()
        public
        onlyowner
    {
    }
}
```

This PR adds a check on the modules that verifies that all transactions in the sequence were executed by the attacker:

```
        for tx in state.world_state.transaction_sequence:
            if not isinstance(tx, ContractCreationTransaction):
                constraints.append(tx.caller == ATTACKER_ADDRESS)
```


b-mueller:
@JoranHonig I noticed that you also introduced `ACTOR_ADDRESSES` but I'm not sure how those are supposed to be used. What's the idea behind it?

b-mueller:
Second question: Is the return value stuff in the delegate call module really necessary? Why not keep it simple like the external_calls module?

norhh:
`I noticed that you also introduced ACTOR_ADDRESSES but I'm not sure how those are supposed to be used. What's the idea behind it?`
@b-mueller In rubixi.sol's case To steal ether we might need to do that from an actor rather than the contract as contract's ether is got from an actor only.And the actor shouldn't be the thief.

b-mueller:
@norhh currently the module still checks if Ether call value > total Ether sent to contract, which should report a bug without explicitly sending Ether to the account first assuming that we don't start with a concrete zero balance. Amirite?

norhh:
`currently the module still checks if Ether call value > total Ether sent to contract` .Yes
but unfortunately Total_Ether_send is only the sum of all the call values of prev transactions, if there is only the attacker then he would just steal his total investment.
`that we don't start with a concrete zero balance. Amirite?` but we currently don't use balances in the ether_thief module

b-mueller:
> Total_Ether_send is only the sum of all the call values of prev transactions

Exactly, so if you can solve `UGT(call_value, eth_sent_total)` is means that the attacker can withdraw *more* than his total investment.

norhh:
If there is no actor then `call_value == eth_send_total`, thats why we might need one.

b-mueller:
> If there is no actor then call_value == eth_send_total

Only if the initial balance is 0, if its symbolic then it's `balance + eth_send_total`.

b-mueller:
Ah sorry, I see what you mean. In the Rubixi examples you're right of course becuase `collectedFees` is paid out, not `this.balance`.

Issue #1085, opened 2019-06-19
Error in on-chain analysis: TypeError: Object of type 'BitVec' is not JSON serializable


b-mueller:
## Description

Running Mythril against smart contracts on-chain results in storage-related type errors. 

## How to Reproduce

Mythril develop:

```console
$ myth -xa 0xc1ca4e6fa2d888168a34a3429bf186bcd9b475a0

mythril.mythril.mythril_analyzer [CRITICAL]: Exception occurred, aborting analysis. Please report this issue to the Mythril GitHub page.
Traceback (most recent call last):
  File "/Users/bernhardmueller/Projects/mythril/mythril/laser/ethereum/state/account.py", line 30, in __getitem__
    return self._storage[item]
KeyError: 3

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bernhardmueller/Projects/mythril/mythril/support/loader.py", line 38, in read_storage
    contract_ref = self.storage_cache[contract_address]
KeyError: 1106346802045208851207092997196424559151310730656

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/Users/bernhardmueller/Projects/mythril/mythril/laser/ethereum/instructions.py", line 1375, in sload_
    return self._sload_helper(global_state, index)
  File "/Users/bernhardmueller/Projects/mythril/mythril/laser/ethereum/instructions.py", line 1421, in _sload_helper
    data = global_state.environment.active_account.storage[index]
  File "/Users/bernhardmueller/Projects/mythril/mythril/laser/ethereum/state/account.py", line 41, in __getitem__
    contract_address=self.address, index=int(item)
  File "/Users/bernhardmueller/Projects/mythril/mythril/support/loader.py", line 46, in read_storage
    contract_address, position=index, block="latest"
  File "/Users/bernhardmueller/Projects/mythril/mythril/ethereum/interface/rpc/base_client.py", line 71, in eth_getStorageAt
    return self._call("eth_getStorageAt", [address, hex(position), block])
  File "/Users/bernhardmueller/Projects/mythril/mythril/ethereum/interface/rpc/client.py", line 61, in _call
    log.debug("rpc send: %s" % json.dumps(data))
  File "/Users/bernhardmueller/.pyenv/versions/3.6.3/lib/python3.6/json/__init__.py", line 231, in dumps
    return _default_encoder.encode(obj)
[...]

PR #1088, opened 2019-06-19
Fix lookup of on-chain storage

nbanmp:


b-mueller:
Looking' good

b-mueller:
Closes #1085.

PR #1033, opened 2019-05-13
Refactor cli

norhh:
fixes #895.

PR #1105, opened 2019-06-25
Analyze contracts with MythX

nbanmp:
Adds the `pro (p)` command to analyze contracts with MythX:
```
━━┫ ./myth pro suicide.sol:Suicide2              
==== Floating Pragma ====
SWC ID: 103
Severity: Low
Contract: Suicide2
PC address: 0
No pragma is set.
It is recommended to make a conscious choice on what version of Solidity is used for compilation. Currently no version is set in the Solidity file.
--------------------

==== Unprotected SELFDESTRUCT Instruction ====
SWC ID: 106
Severity: High
Contract: Suicide2
PC address: 121
The contract can be killed by anyone.
Anyone can kill this contract and withdraw its balance to an arbitrary address.
--------------------
In file: suicide.sol:25

selfdestruct(msg.sender)

--------------------
```

This involved a slight refactor of how we compile contracts, using `--standard-json` instead of `--combined-json`, so the `--solc-args` parameter will be renamed to `--solc-settings` and will be a json string for the settings input of combined-json: https://solidity.readthedocs.io/en/v0.4.24/using-the-compiler.html#input-description.


nbanmp:
Hmm... Not sure why a bunch of commits from the cli refactor (#1033) are displayed here... it should all be merged into develop... Whatever.

norhh:
We can use the username and password from `config.ini` right? It might be a bit more cleaner and a one time setup.

mythril/mythx/__init__.py
```diff
@@ -0,0 +1,97 @@
...
+            sources=sources or None,
+            main_source=main_source,
+            source_list=source_list or None,
+        )
+
+        while not c.analysis_ready(resp.uuid):
+            log.info(c.status(resp.uuid).analysis)
+            time.sleep(5)
+
+        for issue in c.report(resp.uuid):
```
norhh: one of the issue in the list isn't an issue but to upgrade to "pro" version if people are using trail version. So this might crash in such cases. we should handle it like printing to stdout saying your account isn't pro if we receive an issue with no swc id.

mythril/solidity/soliditycontract.py
```diff
@@ -44,23 +44,28 @@ def __init__(self, filename, lineno, code, mapping):
...
-    data = get_solc_json(input_file, solc_args=solc_args, solc_binary=solc_binary)
+    data = get_solc_json(
+        input_file, solc_settings_json=solc_settings_json, solc_binary=solc_binary
+    )
 
     try:
-        for key, contract in data["contracts"].items():
-            filename, name = key.split(":")
-            if filename == input_file and len(contract["bin-runtime"]):
+        for contractName in data["contracts"][input_file].keys():
```
norhh: python usually should have snake case.

mythril/mythx/__init__.py
nbanmp: Ah, good catch. I updated it to set the SWC id to `None` and address to `-1` in that case. While not the best solution, it's what truffle-security does, so I think we can stick with that approach for now.

mythril/solidity/soliditycontract.py
nbanmp: :sweat_smile: This is what I get for referencing javascript when writing this.

mythril/mythx/__init__.py
```diff
@@ -79,8 +79,10 @@ def analyze(contracts: List[SolidityContract], analysis_mode: str = "quick") ->
             issue = Issue(
                 contract=contract.name,
                 function_name=None,
-                address=issue.locations[0].source_map.components[0].offset,
-                swc_id=issue.swc_id[4:],  # remove 'SWC-' prefix
+                address=issue.locations[0].source_map.components[0].offset
```
norhh: can you also put in some text which says the user is using trail mode and to visit mythx.io for pro.

norhh: It can be done in the place where we use default address and password.

nbanmp: :+1:

mythril/mythx/__init__.py
```diff
@@ -30,6 +33,11 @@ def analyze(contracts: List[SolidityContract], analysis_mode: str = "quick") ->
         password=os.environ.get("MYTHX_PASSWORD", "trial"),
     )
 
+    if c.eth_address == "0x0000000000000000000000000000000000000000":
```
norhh: using a variable for that 0x000.. in here and above might be cleaner.

nbanmp: :+1:

PR #1133, opened 2019-07-05
Fixes memory

### LINKED ISSUE #1096: False negative with array args
## Description
The below contract doesn't report the overflow bug
```
contract PausableToken{
  function batchTransfer(address[] memory _receivers,uint cnt, uint256 _value) public  returns (uint) {
    uint256 amount = cnt * _value;
    return amount;
  }
}

```
## How to Reproduce
For example:

```console
$ myth -x <filename.sol>
...
$
```

## Expected behavior
An issue should be detected

## Environment
- Mythril version: `0.21.3`
- Solidity compiler and version: `0.5.0`
- Python version: `3.6`
- OS and Version: Ubuntu 17.10

norhh:
fixes #1096

### DIFF
--- a/mythril/analysis/modules/integer.py
+++ b/mythril/analysis/modules/integer.py
@@ -265,10 +265,7 @@ class IntegerOverflowUnderflowModule(DetectionModule):
         """
 
         stack = state.mstate.stack
-        try:
-            offset, length = get_concrete_int(stack[-1]), get_concrete_int(stack[-2])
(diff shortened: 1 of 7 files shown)


b-mueller:
With this this PR the example #1096 isn't detected on my system:

```
$ $MYTH a pausabletoken.sol 
```

It works after removing the array from call data.

b-mueller:
Works now.

PR #1144, opened 2019-07-09
Fix the issue with dependency pruning

### LINKED ISSUE #1141: With dependency pruning activated, we do not detect bugs that first require 2 or more functions to be executed that are independent of each other.
## Description
For the contract below mythril doesn't detect this with dependency pruning enabled.
```


contract Rubixi {
    uint private collectedFees = 0;

    address private creator;


    function() external payable {
      if (msg.value < 1 ether ) {
            collectedFees += msg.value;
            return;
      }

    }

    function dynamicPyramid() public {
        creator = msg.sender;
    }


    function collectAllFees() public{
        require(collectedFees > 0);
        creator.transfer(collectedFees);
        collectedFees = 0;
    }

```

## How to Reproduce

```console
$ myth a file.sol
nothing
...
$
```

## Expected behavior
It should be detected

## Environment
Please modify for your setup

- Mythril version: `develop`
- Python version: `3.6.7`
- OS and Version: Ubuntu 18.10

norhh:
Checks whether the block is an if condition or a loop before checking for a state skip. fixes #1141. This doesn't still fix rubixi.sol :cry:

### DIFF
--- a/mythril/laser/ethereum/plugins/implementations/dependency_pruner.py
+++ b/mythril/laser/ethereum/plugins/implementations/dependency_pruner.py
@@ -142,7 +142,7 @@ class DependencyPruner(LaserPlugin):
         self.iteration = 0
         self.dependency_map = {}  # type: Dict[int, List[object]]
         self.protected_addresses = set()  # type: Set[int]
-        self.loop_address = set()         # type: Set[int]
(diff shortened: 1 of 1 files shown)


norhh:
This causes unnecessary slowdown, we can probably try something else.

PR #1146, opened 2019-07-10
Refactor pruning plugin

b-mueller:
New algorithm:

- this block is encountered for the first time in the current 'thread' ? -> execute
- somewhere along this path are state modifications that have been observed to affect execution -> execute 
- somewhere along this paths are dependencies on state that has been modified in the previous transaction -> execute 
- otherwise -> skip

### DIFF
--- a/mythril/laser/ethereum/plugins/implementations/dependency_pruner.py
+++ b/mythril/laser/ethereum/plugins/implementations/dependency_pruner.py
@@ -45,9 +45,8 @@ class DependencyAnnotation(StateAnnotation):
     def extend_storage_write_cache(self, iteration: int, value: object):
         if iteration not in self.storage_written:
             self.storage_written[iteration] = [value]
-        else:
-            if value not in self.storage_written[iteration]:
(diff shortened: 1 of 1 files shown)


b-mueller:
No regressions on quick Edelweiss run.

mythril/laser/ethereum/plugins/implementations/dependency_pruner.py
```diff
@@ -44,15 +43,11 @@ def get_storage_write_cache(self, iteration: int):
...
-                self.storage_written[iteration] = [value]
-            else:
-                if value not in self.storage_written[iteration]:
-                    self.storage_written[iteration].append(value)
-        except Z3Exception as e:
-            # FIXME: This should not happen unless there's a bug in laser such as a BitVec512 being generated
-            log.debug("Error updating storage write cache: {}".format(e))
+        if iteration not in self.storage_written:
+            self.storage_written[iteration] = [value]
+        else:
```
norhh: We can merge this and the below statement to elif

PR #1163, opened 2019-07-15
Nested bitvec func 

JoranHonig:
This pr extends the current bvf model to work appropriately with nested bitvecfuncs.
It does so by adding additional implication rules to the comparative logic.


mythril/laser/smt/bitvecfunc.py
```diff
@@ -26,18 +26,19 @@ def _arithmetic_helper(
...
-        # TODO: Find better value to set input and name to in this case?
-        input_string = "MisleadingNotationop(invhash({}) {} invhash({})".format(
-            hash(a), operation, hash(b)
-        )
         return BitVecFunc(
             raw=raw,
             func_name="Hybrid",
-            input_=BitVec(z3.BitVec(input_string, 256), annotations=union),
+            input_=BitVec(z3.BitVec("", 256), annotations=union),
+            nested_functions=a.nested_functions + b.nested_functions + [a, b],
```
norhh: No inverse of functions exist for our current BitVecFunc supporting functions, So hash(a) + hash(b)'s inverse has no use(or no one can find it). We can return a BitVec in these cases right?

b-mueller:
Here are a few tests:

```
pragma solidity ^0.5.0;

contract Test1 {

    bytes32 a = keccak256(abi.encodePacked(keccak256(abi.encodePacked(bytes32("ABCD")))));
    bytes32 b = keccak256(abi.encodePacked(keccak256(abi.encodePacked(bytes32("ABCD")))));

    function a_and_b_equal() public {
    	assert(a == b);
	}

    function a_and_b_inequal() public {
    	assert(a != b);
	}	
}
```

Works ->

```
$MYTH a test1.sol 
==== Exception State ====
SWC ID: 110
Severity: Low
Contract: Test3
Function name: a_and_b_inequal()
PC address: 83
Estimated Gas Usage: 938 - 1033
A reachable exception has been detected.
It is possible to trigger an exception (opcode 0xfe). Exceptions can be caused by type errors, division by zero, out-of-bounds array access, or assert violations. Note that explicit `assert()` should only be used to check invariants. Use `require()` for regular input checking.
--------------------
In file: mini4.sol:13

assert(a != b)

--------------------
Transaction Sequence:

Caller: [CREATOR], data: [CONTRACT CREATION], value: 0x0
Caller: [SOMEGUY], data: 0x853c514b, value: 0x0
```

In these examples though Mythril can't come up with an input value:

```
pragma solidity ^0.5.0;

contract Test2 {

	bytes32 a = keccak256(abi.encodePacked(bytes32("ABCD")));

    function invariant(bytes32 input) public {
    	assert(keccak256(abi.encodePacked(input)) != a);

	}
}
```

```
pragma solidity ^0.5.0;

contract Test3 {

    bytes32 a = keccak256(abi.encodePacked(keccak256(abi.encodePacked(bytes32("ABCD")))));

    function invariant(bytes32 input) public {
    	assert(keccak256(abi.encodePacked(keccak256(abi.encodePacked(input)))) != a);
	}
}
```

Shouldn't we get assert violation with a = bytes32('ABCD')?

norhh:
Same things should also be done for `Concat()` and `Extract()`(Unfortunately they seem to be in BitVector file). And probably we should move them to a different file in future.

mythril/laser/smt/bitvecfunc.py
JoranHonig: But you want to be able to compare `hash(a) +hash(b) == hash(c) + hash(d)`?

norhh: For that we shouldn't directly add the two nested functions lists right? We might have problems with commutative stuff.

JoranHonig: Could you give an example?

norhh: Oh! you are using product() nevermind.

norhh:
For some reason hash(a) + hash(b) != a + b is returning unsat. this may lead to false positives.
```
def test_bitvecfunc_hash():
    s = Solver()
    input_1 = symbol_factory.BitVecVal(1, 256)
    input_2 = symbol_factory.BitVecVal(5, 256)

    a = symbol_factory.BitVecFuncSym("bvf", "sha3", 256, input_=input_1)
    b = symbol_factory.BitVecFuncSym("bvf10", "sha3", 256, input_=input_2)

    a1 = symbol_factory.BitVecFuncSym("bvf1", "sha3", 256, input_=a)
    b1 = symbol_factory.BitVecFuncSym("bvf11", "sha3", 256, input_=b)

    s.add(a1 + b1 != a + b)

    assert s.check() != z3.unsat
```
We can deal with this latter if needed.(as this won't cause nested map issues)

b-mueller:
If that's too difficult to fix right now we can merge this and create a new issue (it's weird, but shouldn't cause problems too often).

norhh:
Now, BecToken isn't running(fails while creating contract). That should be fixed.

PR #1167, opened 2019-07-16
Load contracts from the blockchain by default

palkeo:
Hi !

First PR, I may not have done everything correctly. (Should I push to develop instead of master?)
Also I was not able to pass the full test suite (some truffle problems finding the right solc compiler…), but most tests passed, and the one that failed should be unrelated.

This is an opinionated choice that would change the default behavior of Mythril. I'm happy to make it an additional parameter to control only contract loading, if you prefer.

What I want is to be able to enable contract loading from the CLI, which is impossible now.

I kept the old parameter name so nothing breaks. It could be deprecated/removed at some point.

### DIFF
--- a/mythril/interfaces/cli.py
+++ b/mythril/interfaces/cli.py
@@ -350,7 +350,7 @@ def create_analyzer_parser(analyzer_parser: ArgumentParser):
         "--create-timeout",
         type=int,
         default=10,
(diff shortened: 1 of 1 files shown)


b-mueller:
Hi @palkeo, thanks for the PR! Yep, please push to develop by default.

palkeo:
Switched to develop.

Issue #1168, opened 2019-07-16
Crash caused by dynamic loader


b-mueller:
## Description

The recent change to activate loading of contract code by default exposed a bug that causes Mythril to crash when when the loader is initialized with an `eth` instance.

## How to Reproduce

Run Mythril against the [send loop sample contract](https://github.com/SmartContractSecurity/SWC-registry/blob/master/test_cases/solidity/call_best_practices/send_loop/send_loop.sol):

```
$ myth a -c 608060405234801561001057600080fd5b5060007379b483371e87d664cd39491b5f06250165e4b18490806001815401808255809150509060018203906000526020600020016000909192909190916101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff1602179055505060007379b483371e87d664cd39491b5f06250165e4b18590806001815401808255809150509060018203906000526020600020016000909192909190916101000a81548173ffffffffffffffffffffffffffffffffffffffff021916908373ffffffffffffffffffffffffffffffffffffffff16021790555050610213806101146000396000f30060806040526004361061004c576000357c0100000000000000000000000000000000000000000000000000000000900463ffffffff16806338e771ab14610051578063bc3da53514610068575b600080fd5b34801561005d57600080fd5b506100666100bf565b005b34801561007457600080fd5b506100a9600480360381019080803573ffffffffffffffffffffffffffffffffffffffff1690602001909291905050506101cf565b6040518082815260200191505060405180910390f35b60005b6000805490508110156101cc576000818154811015156100de57fe5b9060005260206000200160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff166108fc60016000808581548110151561013457fe5b9060005260206000200160009054906101000a900473ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff1673ffffffffffffffffffffffffffffffffffffffff168152602001908152602001600020549081150290604051600060405180830381858888f1935050505015156101bf57600080fd5b80806001019150506100c2565b50565b600160205280600052604060002060009150905054815600a165627a7a723058204f612e911a1a01486cc43dc19e9ebc244fd8c0cdc4075b829a610912a2d273bc0029
mythril.mythril.mythril_analyzer [CRITICAL]: Exception occurred, aborting analysis. Please report this issue to the Mythril GitHub page.
Traceback (most recent call last):
[...]

PR #1169, opened 2019-07-16
 Fix bug #1168 by checking eth is not None in the loader. Refactor the loader.

palkeo:
Also refactor the loader quite a bit:
 - remove the cache dictionary, and use functools.lru_cache instead
 - also use lru_cache for caching contract code
 - add type annotations, now mypy would be complaining if we didn't
   check "self.eth" for None.
 - use log.debug() with %s, it's good pratice and allow advanced logging
   to group by log messages / parameters if needed.

Let me know if you prefer to split the refactoring part with the part where I check that "eth" is None.

### DIFF
--- a/mythril/support/loader.py
+++ b/mythril/support/loader.py
@@ -36,13 +36,12 @@ class DynLoader:
         :param index:
         :return:
         """
-        # TODO: Don't raise Exception. Generic and inconsistent with below (ValueError).
         if not self.storage_loading:
-            raise Exception(
+            raise ValueError(
                 "Cannot load from the storage when the storage_loading flag is false"
             )
         if not self.eth:
-            raise Exception("Cannot load from the storage when eth is None")
+            raise ValueError("Cannot load from the storage when eth is None")
 
         return self.eth.eth_getStorageAt(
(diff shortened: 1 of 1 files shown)


mythril/support/loader.py
```diff
@@ -3,70 +3,63 @@
...
         :param index:
         :return:
         """
+        # TODO: Don't raise Exception. Generic and inconsistent with below (ValueError).
         if not self.storage_loading:
             raise Exception(
                 "Cannot load from the storage when the storage_loading flag is false"
             )
+        if not self.eth:
+            raise Exception("Cannot load from the storage when eth is None")
```
norhh: We can switch to value error like in the below case.

mythril/support/loader.py
```diff
@@ -3,70 +3,62 @@
 from mythril.disassembler.disassembly import Disassembly
 import logging
 import re
+import functools
+from mythril.ethereum.interface.rpc.client import EthJsonRpc
+from typing import Optional, Dict
```
norhh: Dict is not being used right?

PR #1170, opened 2019-07-17
Feature/regioned storage

norhh:
Oops, commits from the BitVecFunc PR are here as they were squashed and merged. But those commits aren't reflected in the code, so it should be fine


JoranHonig:
The ArrayStorageRegion rn implements regioning for different keccaks. 

I think a better solution would be to have the Storage, take care of all this stuff and keep the ArrayStorage & IteStorage quite simple

JoranHonig:
Array Storage seems to leverage a Dict. Does a dict preserve order?

norhh:
Closing this as we aren't going with this storage at the moment.

PR #1171, opened 2019-07-23
Enable coverage based search strategy through CLI

e-ngo:
Enable coverage based search strategy through CLI with "--enable-code-coverage".

### DIFF
--- a/mythril/laser/ethereum/svm.py
+++ b/mythril/laser/ethereum/svm.py
@@ -56,6 +56,7 @@ class LaserEVM:
         requires_statespace=True,
         enable_iprof=False,
         enable_coverage_strategy=False,
+        instruction_laser_plugin=None,
     ) -> None:
         """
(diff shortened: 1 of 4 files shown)


mythril/laser/ethereum/svm.py
```diff
@@ -102,6 +103,18 @@ def __init__(
...
 
+        if enable_coverage_strategy:
+            from mythril.laser.ethereum.plugins.implementations.coverage.coverage_plugin import (
+                InstructionCoveragePlugin,
+            )
+            from mythril.laser.ethereum.plugins.implementations.coverage.coverage_strategy import (
+                CoverageStrategy,
+            )
+
+            instruction = InstructionCoveragePlugin()
```
norhh: InstructionCoveragePlugin is already being initialised by plugin loader, that should be used in here.

PR #1173, opened 2019-07-25
Print initial state in jinja

e-ngo:
Updated jinja templates to reflect initialStates of transaction sequences. Should 'code' also be included as part of the initial state?


norhh:
`Should 'code' also be included as part of the initial state?` Currently I don't think so. If we want to then we can add it later.

PR #1175, opened 2019-07-29
Refactored interfaces to allow custom modules

e-ngo:
Allows the running of custom modules in addition to built-in modules by way of specifying the custom modules directory with the --custom-modules-directory option.

### DIFF
--- a/mythril/analysis/security.py
+++ b/mythril/analysis/security.py
@@ -75,28 +75,20 @@ def get_detection_modules(entrypoint, include_modules=(), custom_modules_directo
             if module.detector.entrypoint == entrypoint:
                 _modules.append(module)
     if custom_modules_directory:
-        custom_modules = [os.path.abspath(custom_modules_directory)]
-        sys.path.append(custom_modules_directory)
+        custom_modules_path = os.path.abspath(custom_modules_directory)
+        if custom_modules_path not in sys.path:
+            sys.path.append(custom_modules_path)
 
-        for loader, module_name, _ in pkgutil.walk_packages(custom_modules):
+        for loader, module_name, _ in pkgutil.walk_packages([custom_modules_path]):
             if include_modules and module_name not in include_modules:
                 continue
 
             if module_name != "base":
-                module = importlib.import_module(module_name, custom_modules[0])
+                module = importlib.import_module(module_name, custom_modules_path)
                 module.log.setLevel(log.level)
                 if module.detector.entrypoint == entrypoint:
                     _modules.append(module)
 
-    """
-    for loader, module_name, _ in pkgutil.walk_packages([custom_modules_path]):
-
-    custom_modules_path = os.path.abspath("custom/")
-    sys.path.append(custom_modules_path);
-    module = importlib.import_module(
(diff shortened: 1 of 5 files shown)


mythril/analysis/security.py
```diff
@@ -45,42 +51,57 @@ def get_detection_module_hooks(modules, hook_type="pre"):
...
+            continue
+
+        if module_name != "base":
+            module = importlib.import_module("mythril.analysis.modules." + module_name)
+            module.log.setLevel(log.level)
+            if module.detector.entrypoint == entrypoint:
+                _modules.append(module)
+    if custom_modules_directory:
+        custom_modules = [os.path.abspath(custom_modules_directory)]
+        sys.path.append(custom_modules_directory)
```
nbanmp: It is probably worth checking if the custom_modules_directory has already been appended to the python path, as this method will be called multiple times.

mythril/analysis/security.py
```diff
@@ -45,42 +51,57 @@ def get_detection_module_hooks(modules, hook_type="pre"):
     return dict(hook_dict)
 
 
-def get_detection_modules(entrypoint, include_modules=()):
+def get_detection_modules(entrypoint, include_modules=(), custom_modules_directory=""):
```
nbanmp: This method could be more concise by first checking for custom modules, then creating a list of packages with something like this:
```python
custom_packages = list(pkgutil.walk_packages(custom_modules_directory)) if custom_modules_directory else []`
packages = list(pkgutil.walk_packages(modules.__path__)) + custom_packages

for loader, module_name, _ in packages:
    ...
```

mythril/analysis/security.py
```diff
@@ -45,42 +51,57 @@ def get_detection_module_hooks(modules, hook_type="pre"):
...
 
+    """
+    for loader, module_name, _ in pkgutil.walk_packages([custom_modules_path]):
+
+    custom_modules_path = os.path.abspath("custom/")
+    sys.path.append(custom_modules_path);
+    module = importlib.import_module(
+        module_name, custom_modules_path
+    )
+    """
```
nbanmp: This comment can be deleted now. :smile:

e-ngo: Debugging master 😆

mythril/analysis/security.py
```diff
@@ -45,42 +51,55 @@ def get_detection_module_hooks(modules, hook_type="pre"):
...
-                if module.detector.entrypoint == entrypoint:
-                    _modules.append(module)
-    else:
-        for module_name in include_modules:
-            module = importlib.import_module("mythril.analysis.modules." + module_name)
-            if module.__name__ != "base" and module.detector.entrypoint == entrypoint:
-                module.log.setLevel(log.level)
+            except ModuleNotFoundError:
+                try:
+                    module = importlib.import_module(module_name, custom_modules_path)
```
nbanmp: Hmm... I hadn't noticed that it's a different syntax to import a module from a custom path.

I'm wondering about the situation where a custom module has the same name as a non-custom module. Ehh, we can probably just pretend that it will never happen.

PR #1177, opened 2019-07-29
Various small fixes.

palkeo:
That I encountered when reading the code.

 - comments
 - good practices to increase readability
 - unused variables
 - TODOs (not fixing them, because I'm not sure how to yet, and I want to keep this PR purely minor fixes)

### DIFF
--- a/mythril/analysis/modules/ether_thief.py
+++ b/mythril/analysis/modules/ether_thief.py
@@ -120,8 +120,6 @@ class EtherThief(DetectionModule):
             state.current_transaction.caller == ATTACKER_ADDRESS,
         ]
 
-        log.debug("Constraints: %s", constraints)
-
         try:
 
             transaction_sequence = solver.get_transaction_sequence(state, constraints)
@@ -144,10 +142,6 @@ class EtherThief(DetectionModule):
         except UnsatError:
             log.debug("No model found")
             return []
-
-        log.debug("Found ether thief issue: %s", issue.as_dict)
-
-        log.debug("Model: %s", solver.pretty_print_model(solver.get_model(constraints)))
 
         return [issue]
 
--- a/mythril/analysis/modules/ether_thief.py
+++ b/mythril/analysis/modules/ether_thief.py
@@ -93,7 +93,9 @@ class EtherThief(DetectionModule):
             Also constrain the addition from overflowing (otherwise the solver produces solutions with
             ridiculously high call values).
             """
-            constraints += [BVAddNoOverflow(eth_sent_by_attacker, tx.call_value, signed=False)]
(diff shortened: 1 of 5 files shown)


mythril/analysis/modules/ether_thief.py
```diff
@@ -120,6 +120,8 @@ def _analyze_state(state):
             state.current_transaction.caller == ATTACKER_ADDRESS,
         ]
 
+        log.debug("Constraints: %s", constraints)
```
nbanmp: Printing the constraints in the debug information can get overly verbose, even in `debug`. Not only that, but even if the debug is not printed, this can sometimes cause a slowdown as well.

mythril/analysis/modules/ether_thief.py
```diff
@@ -140,9 +142,13 @@ def _analyze_state(state):
                 gas_used=(state.mstate.min_gas_used, state.mstate.max_gas_used),
             )
         except UnsatError:
-            log.debug("[ETHER_THIEF] no model found")
+            log.debug("No model found")
             return []
 
+        log.debug("Found ether thief issue: %s", issue.as_dict)
+
+        log.debug("Model: %s", solver.pretty_print_model(solver.get_model(constraints)))
```
nbanmp: I would probably choose not to print the model in the debug info here, especially if we aren't doing it for the other modules as well.

mythril/laser/ethereum/instructions.py
```diff
@@ -172,7 +172,7 @@ def evaluate(self, global_state: GlobalState, post=False) -> List[GlobalState]:
         :return:
         """
         # Generalize some ops
-        log.debug("Evaluating {}".format(self.op_code))
+        log.debug("Evaluating %s at %i", self.op_code, global_state.mstate.pc)
```
nbanmp: :+1: Thanks for this one :smile:

mythril/laser/ethereum/state/account.py
```diff
@@ -93,7 +96,8 @@ def __getitem__(self, item: BitVec) -> BitVec:
                 )
                 self.printable_storage[item] = storage[item]
                 return storage[item]
-            except ValueError:
+            except ValueError as e:
+                log.debug("Couldn't read storage at %s: %s", item, e)
                 pass
```
nbanmp: We can delete pass here now.

mythril/analysis/modules/ether_thief.py
palkeo: Slowdowns are mitigated by the fact that the string is only generated if printed (with passing constraints as a supplemental argument).
But ok, let me remove it.

Issue #1179, opened 2019-07-31
Potential dependency conflicts between mythril and eth-utils


NeolithEra:
Hi, as shown in the following full dependency graph of  **_mythril_**, mythril requires _**eth-utils**_ (the latest version), while the installed version of **_eth-account_**(0.4.0) requires _**eth-utils>=1.3.0,<2**_.

According to Pip's “first found wins” installation strategy, _**eth-utils 1.6.2**_ is the actually installed version.  

Although the first found package version _**eth-utils 1.6.2**_ just satisfies the later dependency constraint （_**eth-utils>=1.3.0,<2**_), it will lead to a build failure once developers release a newer version of **_eth-utils_** in the near future, which is greater than 2.

### Dependency tree--------
```
mythril(version range:)
| +-coloredlogs(version range:>=10.0)
| +-py_ecc(version range:==1.4.2)
| +-ethereum(version range:>=2.3.2)
| | +-coincurve(version range:>=7.0.0)
| | +-future(version range:)
| | +-pbkdf2(version range:)
| | +-py_ecc(version range:)
| | +-pyethash(version range:>=0.1.27,<1.0.0)
| | +-pycryptodome(version range:>=3.4.7)
| | +-pysha3(version range:>=1.0.1)
| | +-PyYAML(version range:)
| | +-repoze.lru(version range:)
| | +-rlp(version range:>=1.0.1,<2.0.0)
| | +-scrypt(version range:)
| +-z3-solver(version range:>=4.8.5.0)
| +-requests(version range:)
| | +-chardet(version range:>=3.0.2,<3.1.0)
| | +-idna(version range:>=2.5,<2.9)
| | +-urllib3(version range:>=1.21.1,<1.26)
| | +-certifi(version range:>=2017.4.17)
| +-py-solc(version range:)
| +-plyvel(version range:)
| +-eth_abi(version range:==1.3.0)
| +-eth-utils(version range:>=1.0.1)
| +-eth-account(version range:>=0.1.0a2,<=0.3.0)
| | +-attrdict(version range:>=2.0.0,<3)
| | +-eth-abi(version range:>=2.0.0b7,<3)
| | +-eth-keyfile(version range:>=0.5.0,<0.6.0)
| | +-eth-keys(version range:>=0.2.1,<0.4.0)
| | | +-eth-utils(version range:>=1.3.0,<2.0.0)
| | +-eth-rlp(version range:>=0.1.2,<1)
[...]

PR #1181, opened 2019-08-01
Fix dependency conflict for issue

NeolithEra:
#1179

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -12,7 +12,6 @@ eth-keys>=0.2.0b3,<0.3.0
 eth-rlp>=0.1.0
 eth-tester==0.1.0b32
 eth-typing>=2.0.0
-eth-utils>=1.0.1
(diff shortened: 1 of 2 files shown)


NeolithEra:
@norhh 
Could you please help me review this issue?    
Thank you very much!

BTW, dependency conflict problems are very common in practice in python ecosystem.
I have encountered the issues several times. What's your oppinion on the challenges of detecting and repairing such issues?
And how to avoid dependency conflict problems among python projects, especially from the ecosystem's perspective?

norhh:
Thanks for the PR @NeolithEra 
`I have encountered the issues several times. What's your opinion on the challenges of detecting and repairing such issues?` One terrible way would be to fix all the versions to a hard coded value(with ==) and update them when necessary, but it's not always feasible but on a positive note we only have to deal with indirect dependency problem.

PR #1186, opened 2019-08-07
Pending opcodes

e-ngo:
Adding extcodehash, create, create2, and log opcodes.


tests/instructions/extcodehash_test.py
```diff
@@ -0,0 +1,44 @@
...
+from mythril.laser.ethereum.state.world_state import WorldState
+from mythril.laser.ethereum.instructions import Instruction
+from mythril.laser.ethereum.transaction.transaction_models import MessageCallTransaction
+
+from mythril.support.support_utils import get_code_hash
+
+from mythril.laser.smt import symbol_factory
+
+
+def test_extcodehash_concrete():
```
norhh: I think we can make it into 3 separate tests, Because if we have it combined then we may not execute all the tests,(if the top one fails then the ones below aren't executed)

e-ngo: Ah I noticed that. Good point. Thanks!

mythril/laser/ethereum/instructions.py
```diff
@@ -1591,12 +1605,41 @@ def create_(self, global_state: GlobalState) -> List[GlobalState]:
...
+                global_state,
+                util.get_concrete_int(mem_offset),
+                util.get_concrete_int(mem_offset + mem_size),
+            )
+        except TypeError:
+            log.debug("Create with symbolic length or offset. Not supported")
+            mstate.stack.append(0, 256)
+            return [global_state]
+
+        caller = environment.sender
```
nbanmp: The caller should actually be the address of the active contract, not the original sender.

mythril/laser/ethereum/instructions.py
```diff
@@ -1184,18 +1186,30 @@ def extcodecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
-        global_state.mstate.stack.append(
-            global_state.new_bitvec("extcodehash_{}".format(str(address)), 256)
-        )
+        world_state = global_state.world_state
+        stack = global_state.mstate.stack
+        address = Extract(159, 0, stack.pop())
+
+        if address.symbolic:
+            log.debug("unsupported symbolic address for EXTCODEHASH")
+            stack.append(global_state.new_bitvec("extcodehash_" + str(address), 256))
```
nbanmp: Hmm... I don't like this, because it is completely symbolic. If, for example code had:
```
require(extcodehash(addr) == 1),
```
this would think it is possible.

I think returning the code hash of an address with no code would be better here, as in general, we've been treating symbolic addresses as non-contract (user) addresses, which still have a codehash, but its always the same.

mythril/laser/ethereum/instructions.py
```diff
@@ -1184,18 +1186,30 @@ def extcodecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
+            return [global_state]
+
+        if address.value not in world_state.accounts:
+            code_hash = symbol_factory.BitVecVal(0, 256)
```
nbanmp: This might not be needed. If I recall correctly, `world_state.accounts_exist_or_load` should add the account to the world_state if it is not already there.

mythril/laser/ethereum/instructions.py
```diff
@@ -1184,18 +1186,30 @@ def extcodecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
+        else:
+            code = world_state.accounts_exist_or_load(
+                hex(address.value), self.dynamic_loader
```
nbanmp: `hex(address.value)` can sometimes lead to weird, difficult to detect bugs - If the address starts with a 0, like `0x0ab4606496db368e3b0dbbf95d75dad1bb433cbd`, hex(address) will return `0xab4606496db368e3b0dbbf95d75dad1bb433cbd`, sometimes causing parsing errors. Just to be on the safe side, it would be good to pad the hexadecimal part to be exactly 40 characters long.

mythril/laser/ethereum/instructions.py
e-ngo: Hey Nathan, thanks for the feedback. For extcodehash, addresses that do not exist should return a 0, so I had to explicitly check it. I use the accounts_exist_or_load primarily to load the code of accounts that do exist.

mythril/laser/ethereum/instructions.py
e-ngo: Nice, good idea!

tests/instructions/create_test.py
```diff
@@ -0,0 +1,35 @@
...
+    og_state.transaction_stack.append(
+        (MessageCallTransaction(world_state=WorldState(), gas_limit=8000000), None)
+    )
+
+    laser_evm = LaserEVM()
+
+    new_states, op_code = laser_evm.execute_state(og_state)
+    # checks
+
+    # TODO: come up with sequence, then grab an address from stack and check that its code is created.
```
norhh: Make a sequence this way.
1) Few Pushs for Create arguments
2) Check if the new contract exists in the account

mythril/laser/ethereum/instructions.py
```diff
@@ -1578,18 +1626,92 @@ def log_(self, global_state: GlobalState) -> List[GlobalState]:
...
+        world_state = global_state.world_state
+
+        try:
+            call_data = get_call_data(
+                global_state,
+                util.get_concrete_int(mem_offset),
+                util.get_concrete_int(mem_offset + mem_size),
+            )
+
+        except TypeError:
```
norhh: I don't think we need this

mythril/laser/ethereum/instructions.py
```diff
@@ -1578,18 +1626,92 @@ def log_(self, global_state: GlobalState) -> List[GlobalState]:
...
+            # can be reached?
+            log.debug("Create with symbolic length or offset. Not supported")
+            mstate.stack.append(0, 256)
+            return [global_state]
+
+        call_data = call_data.concrete(None)
```
norhh: Why is this in here :thinking: ?

mythril/laser/ethereum/instructions.py
```diff
@@ -1578,18 +1626,92 @@ def log_(self, global_state: GlobalState) -> List[GlobalState]:
...
+
+        code_end = len(call_data)
+        for i in range(len(call_data)):
```
norhh: we can use enumerate, that might be cleaner.

mythril/laser/ethereum/instructions.py
```diff
@@ -1578,18 +1626,92 @@ def log_(self, global_state: GlobalState) -> List[GlobalState]:
...
+
+        contract_address = None
+        if create2_salt:
+            salt = hex(create2_salt)[2:]
+            salt = "0" * (64 - len(salt)) + salt
+
+            addr = hex(caller.value)[2:]
+            addr = "0" * (40 - len(addr)) + addr
+
+            Instruction._sha3_gas_helper(global_state, len(code_str[:2] // 2))
```
norhh:
```suggestion
            Instruction._sha3_gas_helper(global_state, len(code_str[2:] // 2))
```

mythril/laser/ethereum/transaction/transaction_models.py
```diff
@@ -191,12 +191,15 @@ def __init__(
         code=None,
         call_value=None,
         contract_name=None,
+        contract_address=None,
     ) -> None:
         self.prev_world_state = deepcopy(world_state)
         callee_account = world_state.create_account(
-            0, concrete_storage=True, creator=caller.value
+            0, concrete_storage=True, creator=caller.value, address=contract_address
```
norhh: I think this might break if we supply world_state.create_account() with a hex address.

mythril/laser/ethereum/instructions.py
```diff
@@ -1578,18 +1626,92 @@ def log_(self, global_state: GlobalState) -> List[GlobalState]:
...
+
+        code = Disassembly(code_str)
+
+        caller = environment.active_account.address
+        gas_price = environment.gasprice
+        origin = environment.origin
+
+        contract_address = None
+        if create2_salt:
+            salt = hex(create2_salt)[2:]
```
norhh: we can merge these two lines right?

norhh:
There are also commit diffs from the constructor args PR, If you merge develop into this branch then it would be a bit more easier to review.

mythril/laser/ethereum/instructions.py
```diff
@@ -1578,18 +1626,92 @@ def log_(self, global_state: GlobalState) -> List[GlobalState]:
...
+    ):
+
+        mstate = global_state.mstate
+        environment = global_state.environment
+        world_state = global_state.world_state
+
+        try:
+            call_data = get_call_data(
+                global_state,
+                util.get_concrete_int(mem_offset),
```
norhh: We should directly use the mem_offset and their size instead of using their concrete values. As this function handles symbolic offsets and sizes

mythril/laser/ethereum/gas.py
```diff
@@ -180,10 +180,7 @@ def calculate_sha3_gas(length: int):
     "LOG3": (4 * 375, 4 * 375 + 8 * 32),
     "LOG4": (5 * 375, 5 * 375 + 8 * 32),
     "CREATE": (32000, 32000),
-    "CREATE2": (
-        32000,
-        32000,
-    ),  # TODO: The gas value is dynamic, to be done while implementing create2
+    "CREATE2": (32000, 32000),  # TODO: Make the gas value is dynamic
```
norhh: I am no English expert but shouldn't it be this?
```suggestion
    "CREATE2": (32000, 32000),  # TODO: Make the gas values dynamic
```

mythril/laser/ethereum/instructions.py
e-ngo: Good catch!

mythril/laser/ethereum/instructions.py
```diff
@@ -1063,28 +1079,101 @@ def codecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
-            return self._code_copy_helper(
-                code=global_state.environment.code.bytecode,
-                memory_offset=memory_offset,
-                code_offset=code_offset,
-                size=size,
-                op="CODECOPY",
-                global_state=global_state,
-            )
+
+            if isinstance(global_state.environment.calldata, SymbolicCalldata):
```
nbanmp: With this, in theory, we should no longer need this logic in `_code_copy_helper`:
```
        if concrete_size == 0 and isinstance(
            global_state.current_transaction, ContractCreationTransaction
        ):
            if concrete_code_offset >= len(code) // 2:
                Instruction._handle_symbolic_args(global_state, concrete_memory_offset)
                return [global_state]
```
And `_handle_symbolic_args` can be deleted.

That was our previous, hacky solution that didn't work in a couple situations.

mythril/laser/ethereum/instructions.py
```diff
@@ -1063,28 +1079,101 @@ def codecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
+                if code_offset >= code_size:
+                    return self._calldata_copy_helper(
+                        global_state, global_state.mstate, memory_offset, offset, size
+                    )
+            else:
+                # Copy from both code and calldata appropriately.
+                state = global_state.mstate
```
nbanmp: Can we rename this variable to mstate for clarity? It's going to get confused with global_state as-is.

mythril/laser/ethereum/instructions.py
```diff
@@ -1063,28 +1079,101 @@ def codecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
+                environment = global_state.environment
+                concrete_memory_offset = helper.get_concrete_int(memory_offset)
+                concrete_code_offset = helper.get_concrete_int(code_offset)
+                concrete_size = helper.get_concrete_int(size)
```
nbanmp: Are you able to simplify this section of the codebase by using both `_code_copy_helper` and `_calldata_copy_helper`, just with the right offsets and sizes?

mythril/laser/ethereum/transaction/transaction_models.py
nbanmp: Yes, `world_state.create_account` casts it to a BitVecVal, so it will need to be an integer

mythril/laser/ethereum/instructions.py
e-ngo: Seems to work. My tests are passing. Can you try your tests?

nbanmp: Seems good to me.

PR #1189, opened 2019-08-08
Extract concat invariance

norhh:
ability to retain stuff after multiple Concats and Extracts.
Also the ability to assign to an Extracted part of stuff.
Removed the py35 tests as it seems to have some issue with copy()(bug in python3.5)


norhh:
This would miss a case which is hard to test but might occur, will shift to dict approach

Issue #674, opened 2018-11-07
Slightly wrong concrete transaction data in multi-tx & multi-contract (delegatecall) scenario


b-mueller:
## Description

Running Mythril on a relatively complex example appears to detect the correct bug, but returns slightly wrong concrete transaction data to reach it. Here is the code:

```solidity
pragma solidity ^0.4.24;

contract EvilDelegate {

  bool xyz;

  function abc() public {
    // do nothing
    xyz = false;
  }
}

contract Wallet {

  bool private initialized = false;
  address public owner;
  EvilDelegate private _delegate;

  constructor(address _delegateAddress) public {
    _delegate = EvilDelegate(_delegateAddress);
  }
  
  // The owner can only be set once during initialization.
  function initWallet() public {
    require(initialized == false);
    initialized = true;
    owner = msg.sender;
  }

  // Dangerous! Can only be called by the owner!
  function kill() public {
      require(msg.sender == owner);
      selfdestruct(msg.sender);
  }
  
  function() public payable {
    if (msg.data.length > 0) {
      _delegate.delegatecall(msg.data);
    }
  }
}
```

Wallet is deployed on Ropsten with address [0xc2e75b81d21831c3bf6a739c3b1495c33045d258](https://ropsten.etherscan.io/address/0xc2e75b81d21831c3bf6a739c3b1495c33045d258) and the owner has been initialized to 0x0583858adfe7d1d3DC7F8447301cf7B7b056d638.

Mythril output is as follows:

```
$ $MYTH --rpc infura-ropsten -msuicide -xla 0xc2e75b81d21831c3bf6a739c3b1495c33045d258 --transaction-count 3 --verbose-report
==== Unchecked SUICIDE ====
SWC ID: 106
Type: Warning
Contract: 0xc2e75b81d21831c3bf6a739c3b1495c33045d258
Function name: kill()
PC address: 572
A reachable SUICIDE instruction was detected. The remaining Ether is sent to the caller's address.

--------------------
--------------------
DEBUGGING INFORMATION:

Transaction Sequence: {'1': {'calldata': '0x00', 'call_value': None, 'caller': '0x0000000000000000000000000000000000000001'}, '6': {'calldata': '0x3e326048', 'call_value': '0x0', 'caller': '0x0000000000000000000000000000000000000001'}, '16': {'calldata': '0x41c0e1b5', 'call_value': '0x0', 'caller': '0x0000000000000000000000000000000000000001'}}
```
## Expected behavior

[...]

PR #1190, opened 2019-08-08
Fix a bug: the key for the accounts is an integer.

palkeo:


palkeo:
Found when debugging for #674, but should not be directly related. I'm not sure what the implications of that are, but it definitely sounds wrong.

PR #1191, opened 2019-08-08
Fix a bug: correctly copy printable storage over

palkeo:
Also fix another serialization bug that appears now that the printable
storage is not empty.


palkeo:
Found when debugging for #674, but should not be directly related.

Issue #1176, opened 2019-07-29
Dynamic loader does not retrieve map items


b-mueller:
## Description

Mythril on-chain analysis is unable to dynamically retrieve the content of map items from the blockchain, resulting in false positives.

## How to Reproduce

Run Mythril on an instance of [Ethernaut Fallback](https://ethernaut.openzeppelin.com/level/0x234094aac85628444a82dae0396c680974260be7).

```console
$ myth analyze --rpc infura-ropsten -a 0x2ffc88679871e212aacbd7222900023315299b54
==== Unprotected Ether Withdrawal ====
SWC ID: 105
Severity: High
Contract: 0x2ffc88679871e212aacbd7222900023315299b54
Function name: fallback
PC address: 475
Estimated Gas Usage: 1467 - 2408
Anyone can withdraw ETH from the contract account.
Arbitrary senders other than the contract creator can withdraw ETH from the contract account without previously having sent an equivalent amount of ETH to it. This is likely to be a vulnerability.
--------------------
Transaction Sequence:

Caller: [ATTACKER], function: contribute(), txdata: 0xd7bb99ba, value: 0x1
Caller: [ATTACKER], function: withdraw(), txdata: 0x3ccfd60b, value: 0x0
```

This should return an issue reachable in 3 transactions but not in 2. Apparently, Mythril does not consider that `contributions[msg.sender] = 1000 * (1 ether)`.

## Expected behavior

Should return no results with -t2 and the correct result with -t3 e.g.

```
Transaction Sequence:

Caller: [ATTACKER], function: contribute(), txdata: 0xd7bb99ba, value: 0x1
Caller: [ATTACKER], function: fallback, txdata: , value: 0x0
Caller: [ATTACKER], function: withdraw(), txdata: 0x3ccfd60b, value: 0x0
```

palkeo:
Haha, I actually stumbled independently on that same bug.

I have a "fix", but as my understanding of the storage is pretty limited, it likely will break other things.

Leaving it here just in case: in account.py, class Storage, method __getitem__, in the condition, replace "value.value == 0" by "not value.value".

palkeo:
Actually maybe my fix is correct. PR opened.

b-mueller:
The false positive was fixed with  #1178. Unfortunately we're not detecting the issues at all now:

```
$ $MYTH analyze -mether_thief --rpc localhost:8545 -a 0x169fa8d8a900f32d956f7599abbfde72b015dd46 -t3
[...]

PR #1192, opened 2019-08-08
Fix for #1176: track what storage has been initialized. Load uninitialized storage once.

palkeo:
This should be the right fix for #1176 that does not introduce a regression (see #674 for a regression example)

Please triple check this code, I still don't know how to run integrations tests properly, and I feel like the default test suite is pretty "light" (it didn't catch a few errors I made). I tested on a few contracts but that's all.

### DIFF
--- a/mythril/laser/ethereum/state/account.py
+++ b/mythril/laser/ethereum/state/account.py
@@ -4,7 +4,7 @@ This includes classes representing accounts and their storage.
 """
 import logging
 from copy import copy, deepcopy
-from typing import Any, Dict, Union, Tuple, cast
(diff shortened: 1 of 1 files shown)


mythril/laser/ethereum/state/account.py
```diff
@@ -60,6 +60,7 @@ def __init__(self, concrete=False, address=None, dynamic_loader=None) -> None:
         self.printable_storage = {}  # type: Dict[BitVec, BitVec]
 
         self.dynld = dynamic_loader
+        self.storage_keys_loaded = set()
```
norhh:
```suggestion
        self.storage_keys_loaded = set()    # type: Set[int] 
```

PR #1195, opened 2019-08-09
Fix support for calls forwarding the entire calldata.

palkeo:
This is the core fix for issue #674
Note that another PR is still needed to propagate annotations for that bug to be fully fixed.

### DIFF
--- a/mythril/laser/ethereum/call.py
+++ b/mythril/laser/ethereum/call.py
@@ -16,7 +16,7 @@ from mythril.laser.ethereum.state.calldata import (
     ConcreteCalldata,
 )
 from mythril.laser.ethereum.state.global_state import GlobalState
-from mythril.laser.smt import BitVec
+from mythril.laser.smt import BitVec, Bool, is_true
 from mythril.laser.smt import simplify, Expression, symbol_factory
 from mythril.support.loader import DynLoader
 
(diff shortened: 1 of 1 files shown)


mythril/laser/ethereum/call.py
```diff
@@ -197,10 +197,12 @@ def get_call_data(
     )
 
     uses_entire_calldata = simplify(
-        memory_size - global_state.environment.calldata.calldatasize == 0
+        memory_size == global_state.environment.calldata.calldatasize
     )
 
-    if uses_entire_calldata is True:
+    if (isinstance(uses_entire_calldata, bool) and uses_entire_calldata) or (
```
norhh: `uses_entire_calldata` is always `Bool` right?

palkeo:
Does is_true means it must always be true, or that it can be true ?
If it means it can be true, maybe I should change it to "not is_false()" (it cannot be false, i.e. it is always true) ?

mythril/laser/ethereum/call.py
palkeo: Probably so. I fixed it. However in the JUMPI implementation it's checked for bool as well. Can you double check?

palkeo: Reverted, it seems that it breaks a test, so I guess it's not always Bool. I cannot verify because I cannot reproduce locally (don't have the right solidity version…)

norhh:
I think is_true() is always true.

palkeo:
Ok thanks. Then it looks in a good shape for merging :)

PR #1197, opened 2019-08-09
Propagate annotations to the parent call, when a call succeeds.

palkeo:
This makes the MutationPruner behave correctly, and should fix bug #674 

It's more of a proposal than an actual PR, please make sure it looks good and tell me if there is a better way. It can totally break other things.


mythril/laser/ethereum/svm.py
```diff
@@ -337,6 +337,9 @@ def execute_state(
                     not isinstance(transaction, ContractCreationTransaction)
                     or transaction.return_data
                 ) and not end_signal.revert:
+                    end_signal.global_state._annotations += copy(
```
norhh: something like `global_state.add_annotations()` would be good. WDYT?

palkeo: Yes, I improved it.
However, are you sure that's the right way to proceed ? I see that the StateAnnotation class also has a persist_to_world_state property (I tried setting it to true for the MutationAnnotation, it didn't help). Should a similar property be created for propagating to the parent maybe ? It won't cause bugs if I propagate everything ?

norhh: `Should a similar property be created for propagating to the parent maybe` Yeah we can, but it would be good to create a separate annotation for it which will be propogated.

palkeo: Not sure what you mean. Having a whitelist of annotations that are propagated? A second property (_propagating_annotations or something) that is propagated?
I'd like to know what you prefer so I can move on with the best option, or know if the current PR seems ok.

norhh: A whitelist might do. Or we can persist that annotation to world state and then it can be retrieved after the call to global state from world state.(I think retrieving wasn't implemented before).

palkeo: Ok actually I moved to use persist_to_world_state, I think it makes much more sense.

However, it means that reverted calls still keep the world_state annotation. But I feel like it's meant this way, otherwise why would you have world_state annotations ?

I tested again on #674, this PR will finally fix it for good.

Issue #1019, opened 2019-05-04
feat: support only detect paths initiating from a function


yxliang01:
## Description

As for as I know, currently, Mythril only supports specifying the contract of interest, but not function of interest. I think would be great if it supports specifying function of interest.

## Background

When Mythril detected a smart contract is vulnerable, we would want to modify the reportedly vulnerable statements. Thus, to verify we have successfully addressed the problem, we only need Mythril care about the paths starting from the modified functions, not every path in the smart contract. So, supporting such functionality would allow faster debugging time (arguable, Mythril can be very slow under some smart contracts, run > 20 mins).

Thanks

norhh:
i should still do some refactoring for this(and it's not in our current priorities), meanwhile you can try this branch https://github.com/ConsenSys/mythril-classic/pull/979 I think thats what you need right?

yxliang01:
Thanks! This looks to be exactly but a more generic solution to the feature I am talking about @norhh . Any estimate for this being stably merged?

norhh:
It might take quite a while, as it’s not in the v1 roadmap currently, I might probably be able to put a pr in a week or 2 but I am not sure when it might get merged

JoranHonig:
@yxliang01 While @norhh's approach allows you to force mythril towards some executions. It won't solve your problem entirely @yxliang01.

The changes in the code can change behaviour that relates to the persistent state, changed code can also be dependent on some persistent variables. 
Only analyzing a given sequence of functions will only give a limited guarantee that the vulneability is actually patched.

PR #1199, opened 2019-08-13
Add LASER plugin for whitelisting and blacklisting functions

b-mueller:
Here is a PR from the airport. It adds a plugin that allows whitelisting or blacklisting of functions. Also adds two new CLI arguments `--func-blacklist` and `--func-whilelist`. E.g.:

```
$myth a test.sol --func-blacklist "guess(uint8),refund(address)"
```

This addresses #1019 and is a prerequisite for adding better formal verification support.

**Known limitation:**

"fallback" is not affected by the lists because Mythril currently does not differentiate between the function prologue and the actual fallback function (i.e. if the user filters "fallback" then Mythril executes nothing at all).

### DIFF
--- a/mythril/interfaces/cli.py
+++ b/mythril/interfaces/cli.py
@@ -368,7 +368,20 @@ def create_analyzer_parser(analyzer_parser: ArgumentParser):
         action="store_true",
         help="turns off getting the data from onchain contracts (both loading storage and contract code)",
     )
-
+    options.add_argument(
+        "-w",
+        "--func-whitelist",
+        type=str,
+        help="Execute only functions with these signatures (comma-separated list)",
+        metavar="FUNCTION_SIGS",
+    )
+    options.add_argument(
+        "-n",
+        "--func-blacklist",
+        type=str,
+        help="Don't execute functions with these signatures (comma-separated list)",
+        metavar="FUNCTION_SIGS",
+    )
     options.add_argument(
         "--phrack", action="store_true", help="Phrack-style call graph"
     )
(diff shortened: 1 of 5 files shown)


mythril/interfaces/cli.py
```diff
@@ -556,6 +569,17 @@ def execute_command(
             print("Disassembly: \n" + disassembler.contracts[0].get_creation_easm())
 
     elif args.command in ANALYZE_LIST:
+
+        if args.func_whitelist:
+            _whitelist = args.func_whitelist.split(",")
```
norhh: There can be multiple args, so we can't split by comma. `batchTransfer(uint a, uint b)` function will be split

norhh:
I don't think this can work.
As code will be like 
```
func1_space
func2_space
...
```
if we skip state at one function(say func2) then we will skip all the below functions. I think we should constrain calldata for this.

gkevinyen5418:
Constraining the calldata should work. I am working on a feature similar to this recently. Basically, I let user specify a list of functions of interest, then only verifies transaction sequences which calls one of them at least once.

I also generate callgraph from solidity AST, so that transaction sequences indirectly calling them are also verified.

This should be useful for users changing parts of their contracts and do not want to reverify the contract entirely.

b-mueller:
Yeah you're right, constraining the calldata is a better way to do this.

@gkevinyen5418 would you like to contribute this feature?

gkevinyen5418:
I would love to.

However, the feature I proposed requires source-code-level callgraphs (To resolve dependencies among functions).
I tried to use [surya](https://github.com/ConsenSys/surya) for this purpose, but it is unable to handle overloading and inheritance correctly.

As an alternative, I wrote my own AST visitor to do so. I will try to create a pull request to [solidity](https://github.com/ethereum/solidity) recently and see what happens.

PR #1200, opened 2019-08-14
Change actor addresses in line with MythX test case conventions

b-mueller:
Changes "actor" addresses to the MythX standard format.


b-mueller:
For some reason this is failing passing tests.. not sure why.

JoranHonig:
Maybe the additional actor is making some queries more difficult, and as a result some behaviours aren't covered.

b-mueller:
OK, I'll re-run with 3 actors then. I only added the 4th one because 4 is a nice round number.

PR #1202, opened 2019-08-20
Fixing constructor arguments

e-ngo:


mythril/laser/ethereum/instructions.py
```diff
@@ -1026,14 +1047,24 @@ def codecopy_(self, global_state: GlobalState) -> List[GlobalState]:
...
-            size=size,
-            op="CODECOPY",
-            global_state=global_state,
-        )
+
+        if (
+            isinstance(global_state.current_transaction, ContractCreationTransaction)
+            and code_offset >= len(global_state.environment.code.bytecode) // 2
+        ):
+            offset = code_offset - len(global_state.environment.code.bytecode) // 2
```
norhh: It would be good for the future if we document here why we are doing this.

norhh:
We should also somehow include the contract creation calldata trace in the output as we won't know the initial state of the contract in these cases.

PR #1205, opened 2019-08-28
Test issue constraints at the end of each transaction

### LINKED ISSUE #1183: Constraints are only checked at the vulnerable instruction
Any constraints added after a vulnerability is found will not be checked. Not only is concretion incorrect, but false positives are also reported.

The following reports a false positive:
```solidity
pragma solidity ^0.5.0;

contract PwnMe {
    constructor() public payable {

    }

    function die(uint256 x) public {
        require(x == 1337);
        require(false);
    }
}
```

This is true for every module that reports issues before a terminating instruction. Ideally we'd have some logic shared by all of them for this.

nbanmp:
Fixes #1183

Did not make changes to these modules:
 - `suicide.py` - Suicide occurs at the end of a transaction.
 - ~~`delegatecall.py` - This should report an issue the moment a `delegatecall` instruction is reached, as it can be a `suicide` equivalent.~~ Now updated to report an issue at the right time.
 - `dependence_on_predictable_vars.py` - Already had its own logic for reporting issues at the end of a tx. Can be updated.
 - `dos.py` - This doesn't reach the end of a tx by design.
 - `exceptions.py` - Like `suicide`, `assert` is always the final instruction in a tx.
 - ~~`external_calls.py` - This one should probably be changed. I might have just missed it. :upside_down_face:~~ Now updated.
 - `integer.py` - Has its own logic for reporting issues at the end of a tx. Can probably be updated.
 - `multiple_sends.py` - Has its own logic for reporting issues at the end of a tx. May be updateable.
 - `unchecked_retval.py` - Has its own logic for reporting issues at the end of a tx. Can be updated.


mythril/analysis/potential_issues.py
```diff
@@ -0,0 +1,108 @@
...
+def check_potential_issues(state: GlobalState) -> None:
+    """
+    Called at the end of a transaction, checks potential issues, and
+    adds valid issues to the detector.
+
+    :param state: The final global state of a transaction
+    :return:
+    """
+    annotation = get_potential_issues_annotation(state)
+    for potential_issue in annotation.potential_issues[:]:
```
norhh: do we need [:] in here?

mythril/analysis/modules/base.py
```diff
@@ -3,6 +3,8 @@
 
 import logging
 from typing import List, Set
+
+from mythril.analysis.potential_issues import PotentialIssue
```
norhh: we can remove this right?

norhh:
`delegatecall.py - This should report an issue the moment a delegatecall instruction is reached, as it can be a suicide equivalent. Now updated to report an issue at the right time.` Delegate call can hit revert and get reverted right?

mythril/analysis/potential_issues.py
```diff
@@ -0,0 +1,108 @@
...
+        try:
+            transaction_sequence = get_transaction_sequence(
+                state, state.mstate.constraints + potential_issue.constraints
+            )
+        except UnsatError:
+            continue
+
+        annotation.potential_issues.remove(potential_issue)
+        potential_issue.detector._cache.add(potential_issue.address)
+        potential_issue.detector._issues.append(
```
norhh: argh! now we should make these variables public.

nbanmp:
@norhh A delegatecall can call selfdestruct, meaning that the revert can't be reached.

PR #1218, opened 2019-09-25
Remove truffle support from mythril

norhh:
Finally removing truffle support from mythril


norhh:
Oh! yeah, forgot to `git add` that change

Issue #1217, opened 2019-09-25
Confused with solc version


MountainTrap:
**I want to use mythirl to test a given contract with any solidity version.**
I look at the doc and found that  --solv can change the specified solidity complier version，but then I find --solv is using py-solc which only can install following version to /root/.py-solc:
``` v0.4.1, v0.4.11, v0.4.12, v0.4.13, v0.4.14, v0.4.15, v0.4.16, v0.4.17, v0.4.18, v0.4.19, v0.4.2, v0.4.20, v0.4.21, v0.4.22, v0.4.23, v0.4.24, v0.4.25, v0.4.6, v0.4.7, v0.4.8, v0.4.9```
and the py-solc looks like not maintained.
So How do I use myth with specified solidity compiler，if I use myth directly
```
myth a call.sol
```
it will use the solc in system right?
So I need to check the contract version in the first line and change my system solc everytime(and I need to install all the version), is it the rightway or is there is a better practice. is there a easy way to chang my system solc(I found a package crytic/solc-select)?

I am very confused， thanks for your help!!

norhh:
Hi @MountainTrap , I didn't check solc-select, but one way would be to have the latest version of solc installed and for analysis just change the pragma by adding a ^(don't forget to revert it after the analysis is done if it wasn't intended), In nearly all cases the analysis results wouldn't change with that difference at the moment. This only works for versions >= 0.5.0 , in other cases you can just use --solc arg,
Another easier way is to try out mythx truffle plugin https://docs.mythx.io/en/latest/tools/truffle/ , it detects more vulnerabilities than mythril(it also uses mythril) and doesn't have this problem.

nbanmp:
This has been resolved, the `--solc` argument will now also work with compiler versions greater than 0.5.0.

PR #1219, opened 2019-09-26
Added support for solidity 0.5.x

mderka:
Mythril relies on `py-solc` to install versions specified by the `--solv` parameter. Unfortunately, `py-solc` does not seem maintained and cannot install solc with versions 0.5.x. Thus, Mythril currently supports only the system solc installation for 0.5.x. 

This PR adds `py-solc-x` dependency which is used for versions other than 0.4.x. For 0.4.x, handling by `py-solc` is preserved, because `py-solc-x` declares incompatibility of 0.4.x installations with OSX.

This should solve Mythril's dynamic support for 0.5.x that is reference e.g. in ethereum/py-solc#63, andindirectly also in #1217.

Tested with Docker as follows:

```
docker build .
docker run -v /tmp:/tmp 3ff13f1e1c8b analyze /tmp/contract.sol --solv 0.5.10 -o jsonv2
docker run -v /tmp:/tmp 3ff13f1e1c8b analyze /tmp/contract.sol --solv 0.4.24 -o jsonv2
```

### DIFF
--- a/tests/native_tests.sol
+++ b/tests/native_tests.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.5.0;
+pragma solidity 0.5.3;
 
 
 contract Caller {
--- a/tests/native_tests.sol
+++ b/tests/native_tests.sol
@@ -1,4 +1,4 @@
-pragma solidity 0.5.3;
+pragma solidity ^0.5.0;
 
 
 contract Caller {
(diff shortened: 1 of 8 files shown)


tests/native_tests.sol
```diff
@@ -1,4 +1,4 @@
-pragma solidity 0.5.0;
+pragma solidity 0.5.3;
```
norhh:
```suggestion
pragma solidity ^0.5.3;
```

norhh:
Hi @mderka thanks for the PR. Just a small change is required for the native_tests.sol, as using a ^ would be simpler there at the moment.

tests/native_tests.sol
mderka: Done

PR #1227, opened 2019-10-02
Added returning global state to SIGNEXTEND opcode.

e-ngo:
When arguments to SIGNEXTEND are symbolic, a type error is thrown but no global_state is returned.

### DIFF
--- a/mythril/laser/ethereum/instructions.py
+++ b/mythril/laser/ethereum/instructions.py
@@ -628,7 +628,9 @@ class Instruction:
         except TypeError:
             log.debug("Unsupported symbolic argument for SIGNEXTEND")
             mstate.stack.append(
-                global_state.new_bitvec("SIGNEXTEND({},{})".format(s0, hash(s1)), 256)
+                global_state.new_bitvec(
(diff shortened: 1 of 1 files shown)


mythril/laser/ethereum/instructions.py
```diff
@@ -619,23 +619,24 @@ def signextend_(self, global_state: GlobalState) -> List[GlobalState]:
...
-        s0, s1 = state.stack.pop(), state.stack.pop()
+        mstate = global_state.mstate
+        s0, s1 = mstate.stack.pop(), mstate.stack.pop()
 
         try:
             s0 = util.get_concrete_int(s0)
             s1 = util.get_concrete_int(s1)
         except TypeError:
-            return []
+            mstate.stack.append(s1)
```
norhh: Adding a TODO here about something more appropriate might be good.

norhh: And how about pushing a new variable named "SIGNEXTEND(hash(s0), hash(s1))"? This also has it's own problems.

PR #1230, opened 2019-10-03
Revert "Enable optimizer when compiling Soldity files"

norhh:
Reverts ConsenSys/mythril#1226 
@b-mueller do we want this?(problems with FPs according to aesir).


b-mueller:
I'd leave the optimiser on. It's a matter of taste :)

PR #1231, opened 2019-10-05
Added 'context' to loop annotations.

e-ngo:
Previously, function calls were falsely considered as loops. This PR fixes this issue by considering the 'context' of the call. If a jumpdest was reached by a different instruction, it would not add to its loop annotation.

### DIFF
--- a/mythril/laser/ethereum/state/machine_state.py
+++ b/mythril/laser/ethereum/state/machine_state.py
@@ -95,6 +95,7 @@ class MachineState:
         depth=0,
         max_gas_used=0,
         min_gas_used=0,
+        prev_pc=-1,
     ) -> None:
         """Constructor for machineState.
 
@@ -106,8 +107,9 @@ class MachineState:
         :param depth:
         :param max_gas_used:
         :param min_gas_used:
+        :param prev_pc:
         """
-        self.pc = pc
+        self._pc = pc
         self.stack = MachineStack(stack)
         self.memory = memory or Memory()
         self.gas_limit = gas_limit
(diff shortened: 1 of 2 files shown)


mythril/laser/ethereum/state/machine_state.py
```diff
@@ -224,6 +228,19 @@ def __str__(self):
...
+    @property
+    def pc(self) -> int:
+        """
+
+        :return:
+        """
+        return self._pc
+
+    @pc.setter
+    def pc(self, value):
```
norhh: cool

mythril/laser/ethereum/strategy/extensions/bounded_loops.py
```diff
@@ -64,22 +64,30 @@ def get_strategic_global_state(self) -> GlobalState:
...
-            address = state.get_current_instruction()["address"]
+            cur_instr = state.get_current_instruction()
 
-            if address in annotation._reached_count:
-                annotation._reached_count[address] += 1
+            if cur_instr["opcode"].upper() != "JUMPDEST":
+                return state
+
+            # create unique instruction identifier
+            key = "{}{}{}".format(
```
JoranHonig:
```suggestion
            key = "{};{};{}".format(
```
Address and prev_pc are both integers, I believe that without delimiters there is a slight chanche for collision

norhh: Oh! Yeah :rofl:

e-ngo: Nice, thanks for the suggestion! Changed!

mythril/laser/ethereum/state/machine_state.py
e-ngo: Thanks

PR #1237, opened 2019-10-11
Fix propagation of JumpdestCountAnnotation

nbanmp:


mythril/laser/ethereum/strategy/extensions/bounded_loops.py
```diff
@@ -74,7 +71,7 @@ def get_strategic_global_state(self) -> GlobalState:
                 cur_instr["opcode"], cur_instr["address"], state.mstate.prev_pc
             )
 
-            if key in annotation._reached_count:
+            if key in annotation._reached_count.keys():
```
norhh: we don't need this change right?
```suggestion
            if key in annotation._reached_count:
```

PR #1239, opened 2019-10-14
Display instruction profiler results on ctrl+c or error

nbanmp:


mythril/laser/ethereum/svm.py
```diff
@@ -72,7 +71,7 @@ def __init__(
         :param strategy: Execution search strategy
         :param transaction_count: The amount of transactions to execute
         :param requires_statespace: Variable indicating whether the statespace should be recorded
-        :param enable_iprof: Variable indicating whether instruction profiling should be turned on
+        :param iprof: Instruction Profiler indicating whether instruction profiling should be turned on
```
norhh: We can remove the rest of the comment part right? As this isn't just used for indicating.

PR #1240, opened 2019-10-15
Add tests for extcodecopy

norhh:


tests/instructions/extcodecopy_test.py
```diff
@@ -0,0 +1,86 @@
...
+    (MessageCallTransaction(world_state=WorldState(), gas_limit=8000000), None)
+)
+ans_state.mstate.memory.extend(32)
+ans_state.mstate.memory[2] = 95
+
+
+@patch(
+    "mythril.laser.ethereum.instructions.Instruction._code_copy_helper",
+    return_value=[ans_state],
+)
```
nbanmp: Is there a reason we have to mock `_code_copy_helper` here?

Can we test that the code returned by `accounts_exist_or_load` is written to memory at the expected memory offset after `_code_copy_helper` is called?

We might also not even need to mock `accounts_exist_or_load` if we do something like:
```
new_account = new_world_state.create_account(balance=10, address=10)
new_account.code = Disassembly("61626364")
```

nbanmp: I'm not sure there's a downside to testing it again, and I think there is potential downside to the current testing.

If there was a bug in extcodecopy that caused the input to `_code_copy_helper` to be invalid, that bug would not be caught by the test case. In addition, testing should usually be as implementation-agnostic as is reasonable. If we end up modifying `extcodecopy` to use logic that does not involve using _code_copy_helper in the same way, this test case would break and need to be rewritten.

PR #1247, opened 2019-10-17
Add statespace tests

norhh:


tests/graph_test.py
```diff
@@ -1,45 +1,25 @@
...
-            )
+        analyzer.graph_html(transaction_count=1)
 
-            html = analyzer.graph_html(transaction_count=1)
-            output_current.write_text(html)
-
-            lines_expected = re.findall(
-                r"'label': '.*'", str(output_current.read_text())
-            )
-            lines_found = re.findall(r"'label': '.*'", str(output_current.read_text()))
```
norhh: I always wondered how these tests work without replacing the files, apparently it's comparing [] and [] as there is no label in most of the testable circumstances.

PR #1248, opened 2019-10-17
Add statespace tests

norhh:
I always wondered why i didn't need to update the graph tests after each change.Graph tests were testing for exceptions and they check `assert [] == []` at the end, will remove graph tests files after this PR.


tests/graph_test.py
```diff
@@ -1,45 +1,22 @@
...
-class GraphTest(BaseTestCase):
-    def test_generate_graph(self):
-        for input_file in TESTDATA_INPUTS.iterdir():
-            output_expected = TESTDATA_OUTPUTS_EXPECTED / (
-                input_file.name + ".graph.html"
-            )
-            output_current = TESTDATA_OUTPUTS_CURRENT / (
-                input_file.name + ".graph.html"
-            )
+def test_generate_graph():
```
nbanmp: Since you removed checking `input_file.graph.html`, do you want to you also remove the files themselves?

norhh: Yeah, I planned on removing the directory after this get's merged. Or I can remove it now

JoranHonig:
With this pr there are no assertions anymore right?
So the only thing that would be captured is an exceptional halt for the generation of the graphs?

norhh:
` So the only thing that would be captured is an exceptional halt for the generation of the graphs?` Yes.

JoranHonig:
@norhh okay, do you think we should maybe add a disclaimer along the lines of "This test only checks whether dumping is successfull, not whether the dumped statespace makes sense"

PR #1249, opened 2019-10-18
Refactor code by removing usused and repeated code

norhh:
Also balance_() opcode uses wrong address.

### DIFF
--- a/mythril/laser/ethereum/instructions.py
+++ b/mythril/laser/ethereum/instructions.py
@@ -901,9 +901,7 @@ class Instruction:
         state = global_state.mstate
         address = state.stack.pop()
 
-        balance = global_state.world_state.balances[
-            global_state.environment.active_account.address
-        ]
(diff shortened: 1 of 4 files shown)


mythril/laser/ethereum/instructions.py
```diff
@@ -901,9 +901,7 @@ def balance_(self, global_state: GlobalState) -> List[GlobalState]:
         state = global_state.mstate
         address = state.stack.pop()
 
-        balance = global_state.world_state.balances[
-            global_state.environment.active_account.address
-        ]
+        balance = global_state.world_state.balances[address]
```
nbanmp: Bugs like this seem like something that we could easily add tests for, and probably should have had tests for in our opcode tests. :upside_down_face:

norhh: Yeah

PR #1252, opened 2019-10-21
Add cli options to specify attacker and creator addresses

### LINKED ISSUE #798: Allow setting msg.sender in command line
## Description

Some contracts may have authorization such that only certain addresses are allowed to interact with the contract.  It would be cool if there was a CLI argument that set the value of msg.sender in Mythril Classic so that mythril could run analysis from the perspective of that user.

nbanmp:
I've seen a couple requests for this, so here it is.

resolves #798


mythril/laser/ethereum/transaction/symbolic.py
```diff
@@ -17,14 +18,33 @@
 
 log = logging.getLogger(__name__)
 
-CREATOR_ADDRESS = 0xAFFEAFFEAFFEAFFEAFFEAFFEAFFEAFFEAFFEAFFE
-ATTACKER_ADDRESS = 0xDEADBEEFDEADBEEFDEADBEEFDEADBEEFDEADBEEF
+ACTOR_ADDRESSES = {
```
norhh: How about using a dataclass over a Dict(), It might take some effort now, but our future would be simple due to autocomplete.

nbanmp: The reason I initially went with a dict is so that we could be flexible with adding other addresses, if for example we wanted to add a `SOMEGUY2`. I can use a class with a property for the `creator` and `attacker` address and have that look it up in its dict though.

norhh: Yeah, this should be cool.

ytrezq:
Does this fetch storage data as needed when used on mainnet? Because many times, the constraint like the one in #798 are specified not in code but through storage at contract initialization.

norhh:
@ytrezq it doesn’t, but I think you can query the storage and use that as an arg for cli.

ytrezq:
@norhh of course, I was meaning when analysing.

Issue #1258, opened 2019-10-26
Please support ꜱᴡᴄ 127


ytrezq:
## Description :

Detects the possibility to ᴊᴜᴍᴘ at arbitrary address either because of low level functions used in source code or for well written and audited solidity code but with compiler bugs like `OptimizerStateKnowledgeNotResetForJumpdest` or `OptimizerClearStateOnCodePathJoin` and of course, not simply if they are in the code bug which input sequence to trigger them.

## Note :

A better thing would be to leverage those issues in order to trigger more serious one like being able to withdraw ether or fire a token `Transfer` event without background checks.

## Tests ;

For compiler bugs I couldn’t find the issue number in the GitHub project nor the commit which fixed them so I couldn’t get enough details for a test case.

b-mueller:
@ytrezq #1268 introduces this feature, would you like to test it?

ytrezq:
@b-mueller if it doesn’t detect almost arbitrary jumps from the call graph, then no I can’t. What I’m requesting doesn’t only require adding a module. It also requires adding the capability to Laser accordingly (otherwise it should detects only bugs from solidity code and not low level compiler bugs).

norhh:
`otherwise it should detects only bugs from solidity code and not low level compiler bugs` Hey, mythril is a bytecode analyzer, it doesn't know the difference between solidity source code and bytecode, so there is no need to change anything in laser. Also #1269 is the arbitrary jump feature.

ytrezq:
@norhh yes, but based on a call graph. If the call graph say that a jump is static whereas it could be dynamic then nothing will happen.

norhh:
What I am checking is that if the jump destination is a symbol rather than a concrete value then a multilocation jump is possible.

ytrezq:
We don’t have a minimal example of `OptimizerClearStateOnCodePathJoin` and `OptimizerStateKnowledgeNotResetForJumpdest`. My test case are taking days.

Any ideas of their bug number in https://github.com/ethereum/solidity/issues?

ytrezq:
Ok, here’s the test cases : https://github.com/ethereum/solidity/issues/692 and https://github.com/ethereum/solidity/issues/1416 (the latter being harder to trigger). But I don’t have my personal machines right now (nor in the future) so I can’t compile and run Mythril on those examples.

PR #1268, opened 2019-11-04
Add arbitrary storage write module

norhh:


b-mueller:
Can we add SWC-127 detection to the same module (#1258)?

norhh:
We would need to hook just the JUMP for that module right? In that case it might make more sense to add it to a different module right?

b-mueller:
Your call! If it doesn't fit well with this module we can create a separate one.

norhh:
How about this https://github.com/ConsenSys/mythril/pull/1269

mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,83 @@
...
+            write_slot == symbol_factory.BitVecVal(324345425435, 256)
+        ]
+
+        potential_issue = PotentialIssue(
+            contract=state.environment.active_account.contract_name,
+            function_name=state.environment.active_function_name,
+            address=state.get_current_instruction()["address"],
+            swc_id=WRITE_TO_ARBITRARY_STORAGE,
+            title="Write to an arbitrary storage slot",
+            severity="Medium",
```
nbanmp: Is medium the correct severity here? I feel like this issue is actually critical - It will allow for essentially changing the entire contract's state.

mythril/analysis/modules/arbitrary_write.py
```diff
@@ -0,0 +1,83 @@
...
+            bytecode=state.environment.code.bytecode,
+            description_head="Any storage slot can be written by the caller.",
+            description_tail="The caller can write a value into an arbitrary storage slot."
+            + " This can lead to unintended consequences.",
```
nbanmp: Perhaps we can say something like "An attacker can modify any value in storage" instead of the generic "unintended consequences" line?

ytrezq:
Ok, here’s the test cases : https://github.com/ethereum/solidity/issues/692 (the former being harder to trigger but should be possible with https://gist.github.com/chriseth/3514a465ab4b7d8168f53d786bd0df7c https://github.com/ethereum/solidity/pull/762/commits/ac0d1388008ed616ab7b35a5b3a27e7400decd46) and https://github.com/ethereum/solidity/issues/1416. But I don’t have my personal machines right now (nor in the future) so I can’t compile and run Mythril on those examples. @JoranHonig : I have huge doubts the current method at least works with the first case.

Those bugs are so rare and hard to trigger we might better to have them in the tests cases, isn’t it ?

mythril/analysis/modules/arbitrary_write.py
ytrezq: @nbanmp : it depends : there’s writing to an arbitrary storage slot with almost any values and being able to write only a single restricted value to such arbitrary storage slot.

mythril/analysis/modules/arbitrary_write.py
norhh: `An attacker can modify any value in storage` is same as the first part of the description right?

nbanmp: Yeah, it pretty much means the same thing, it just feels more direct to me at least. Up to you.

mythril/analysis/modules/arbitrary_write.py
nbanmp: Hmm. I would say in that case, it could still be a high severity issue, as it can likely result in a DoS of the contract.

ytrezq: @nbanmp : except if owner functions are only for emergency actions such as stopping transfers on the token.

PR #1269, opened 2019-11-05
Add arbitrary jump module

norhh:
This is much faster than giving out two solver queries out of which one won't use cache.


JoranHonig:
@norhh how does this module perform on: https://swcregistry.io/docs/SWC-127#functiontypessol ?

norhh:
The module looks pathetic but still detects that test case

ytrezq:
Ok, here’s the test cases : https://github.com/ethereum/solidity/issues/692 and https://github.com/ethereum/solidity/issues/1416 (the latter being harder to trigger but should be possible with https://gist.github.com/chriseth/3514a465ab4b7d8168f53d786bd0df7c https://github.com/ethereum/solidity/pull/762/commits/ac0d1388008ed616ab7b35a5b3a27e7400decd46). But I don’t have my personal machines right now (nor in the future) so I can’t compile and run Mythril on those examples.

Those bugs are so rare and hard to trigger we might better to have them in the tests cases, isn’t it ?

PR #1289, opened 2019-12-09
Fix Function signatures

norhh:


b-mueller:
What bug is this related to?

norhh:
If we are not storing the hash from solidity_file we endup returning weird collisions from 4bytes like `many_msg_babbage()` instead of `transfer`.

norhh:
`What bug is this related to?` If we run `solidity_examples/token.sol` It prints out `many_msg_babbage()` instead of `transfer()`

PR #1292, opened 2019-12-25
Support Istanbul fork

norhh:
- Implemented 2 opcodes
- Added a precompile
- some gas changes
- Removed python3.5 support as blake2 bindings doesn't support it.(Our own Implementation of blake2 hash would be tedious, slow and would take 200 - 300 lines)


mythril/laser/ethereum/state/environment.py
```diff
@@ -41,7 +41,10 @@ def __init__(
         self.active_function_name = ""
 
         self.address = active_account.address
+
+        # TODO: Block number should increase with transactions
         self.block_number = symbol_factory.BitVecSym("block_number", 256)
+        self.chainid = symbol_factory.BitVecSym("chain_id", 256)
```
JoranHonig: Maybe we should use a default concrete value here? (cli option maybe)
Different chain_id's could in theory have different opcode implementations right?

mythril/laser/ethereum/state/environment.py
```diff
@@ -41,7 +41,10 @@ def __init__(
         self.active_function_name = ""
 
         self.address = active_account.address
+
+        # TODO: Block number should increase with transactions
```
JoranHonig: This is not necessarily true. A series of transactions can occur within one block. 
Although there would be a constraint that (new blocknumber >= previous blocknumber)

requirements.txt
```diff
@@ -17,6 +17,7 @@ mock
 persistent>=4.2.0
 plyvel
 py-flags
+py-evm
```
JoranHonig: should we consider version locking?

setup.py
```diff
@@ -19,7 +19,7 @@
 URL = "https://github.com/ConsenSys/mythril"
 AUTHOR = "ConsenSys Dilligence"
 AUTHOR_MAIL = None
-REQUIRES_PYTHON = ">=3.5.0"
+REQUIRES_PYTHON = ">=3.6.0"
```
JoranHonig: Is this installed by default in Ubuntu now? / Do we have info on the environments of the people that use mythril?

mythril/laser/ethereum/state/environment.py
JoranHonig: (note to self: tox.ini is also changed) 😄

mythril/laser/ethereum/natives.py
```diff
@@ -192,6 +196,19 @@ def ec_pair(data: List[int]) -> List[int]:
...
 
+def blake2b_fcompress(data: List[int]) -> List[int]:
+    """
+    blake2b hashing
+    :param data:
+    :return:
+    """
+    try:
+        parameters = extract_blake2b_parameters(bytes(data))
+    except ValidationError:
```
JoranHonig: should we log a message here?

JoranHonig: maybe debug or something?

mythril/laser/ethereum/natives.py
```diff
@@ -192,6 +196,19 @@ def ec_pair(data: List[int]) -> List[int]:
     return [0] * 31 + [1 if result else 0]
 
 
+def blake2b_fcompress(data: List[int]) -> List[int]:
```
JoranHonig: are there ethereum consensus tests that we could add to check if the implementation is correct?

mythril/laser/ethereum/instruction_data.py
```diff
@@ -210,7 +212,9 @@ def calculate_native_gas(size: int, contract: str):
     elif contract == "identity":
         gas_value = opcodes.GIDENTITYBASE + word_num * opcodes.GIDENTITYWORD
     else:
-        raise ValueError("Unknown contract type {}".format(contract))
+        # TODO: Add gas for other precompiles, computation should be shifted to natives.py
+        #  as some data isn't available here
+        pass
```
JoranHonig: Now gas_value will contain a None value. Could this break callee's that expect an integer?

mythril/laser/ethereum/natives.py
norhh: There are state tests, but we need to do something with logs to add them. As they check against logs.

norhh: Also, I have added the fork tests in the tests. So the implementation is correct

mythril/laser/ethereum/instruction_data.py
norhh: Yep, But this was actually fixed in the develop, should merge with it.

setup.py
norhh: People who log issues mainly use docker. So it should be fine. The latest Ubuntu has 3.7.5, only the ancient ones like Ubuntu 16(which is dead) has python 3.5 installed by default.

mythril/laser/ethereum/state/environment.py
norhh: Yes, but we can leave it at the moment, as users won't get anything out of it now.

mythril/laser/ethereum/natives.py
JoranHonig: Right I was more wondering about the integration part of it all. To check if the delegation code passes arguments correctly to the function.
The tests that ckeck blake2b directly are nice though 👍

norhh: I think it tests the delegation trail

PR #1306, opened 2020-01-15
Fix create salt

### LINKED ISSUE #1303: TypeError: 'BitVec' object cannot be interpreted as an integer
**Error**
[This error](https://github.com/ConsenSys/mythril/issues/1243) still exists when I run on the address `0x5AbdfcF089573C4E588E7278277E104E45B0fDa9`

**Reproduce**
`docker run --rm mythril/myth -v4 analyze -1 -a 0x5AbdfcF089573C4E588E7278277E104E45B0fDa9 --execution-timeout 3600`

norhh:
fixes #1303 
A bunch of commits from different PR got merged into here


mythril/laser/ethereum/instructions.py
```diff
@@ -1674,20 +1674,37 @@ def _create_transaction_helper(
...
+                salt = hex(create2_salt)[2:]
+                salt = "0" * (64 - len(salt)) + salt
 
-            addr = hex(caller.value)[2:]
-            addr = "0" * (40 - len(addr)) + addr
+                addr = hex(caller.value)[2:]
+                addr = "0" * (40 - len(addr)) + addr
 
-            Instruction._sha3_gas_helper(global_state, len(code_str[2:]) // 2)
+                Instruction._sha3_gas_helper(global_state, len(code_str[2:]) // 2)
```
JoranHonig: Should this also be indented?

JoranHonig:
I'm just wondering if there might be an efficient way to test correct functioning of this code with our unit test suite?
To make sure no regressions occur in the future

norhh:
`efficient way to test correct functioning of this code with our unit test suite?` We can, but this error may not probably occur in the future. I was planning on adding tests related to the whole working of Create and Create2, in other words, some hardcore multi-contract test cases(I am still a bit skeptical of some cases with "create2 + contract creation" and it's a bit complicated to test it without using an example solidity contract).

PR #1308, opened 2020-01-15
Add create and create2 test

norhh:


tests/instructions/create2_test.py
```diff
@@ -0,0 +1,62 @@
...
+    )
+    value = 3
+    salt = 10
+    og_state.mstate.stack = [salt, 6, 0, value]
+    instruction = Instruction("create2", dynamic_loader=None)
+
+    # Act + Assert
+    try:
+        _ = instruction.evaluate(og_state)[0]
+        assert False, "Transaction Start Signal should occur"
```
JoranHonig: `whith pytest.raises(TransactionStartSignal) as t:`
This is the best practice way of doing this I think

tests/instructions/create_test.py
```diff
@@ -0,0 +1,49 @@
...
+    # Act + Assert
+    try:
+        _ = instruction.evaluate(og_state)[0]
+        assert False, "Transaction Start Signal should occur"
+    except TransactionStartSignal as t:
+        assert t.transaction.call_value == value
+        assert t.transaction.code.bytecode == code
+        assert (
+            t.transaction.callee_account.address
+            == world_state._generate_new_address(account.address.value)
```
JoranHonig: This will be a new random address right?

norhh: Currently no nounce is used, that's in the pipeline of stuff to be done.

PR #1314, opened 2020-01-29
Save code to world state

norhh:


mythril/laser/ethereum/state/world_state.py
```diff
@@ -73,26 +73,25 @@ def __copy__(self) -> "WorldState":
...
-                balance=0, address=addr_bitvec.value, dynamic_loader=dynamic_loader
-            )
-        if code is None:
-            code = ""
-        else:
-            code = code.bytecode
-        return code
+            return self.accounts[addr_bitvec.value]
+        if dynamic_loader is None:
+            raise ValueError("dynamic_loader is None")
```
JoranHonig: If the dynamic loader is none shouldn't the default behaviour be to create a new account (similar to  what would happen if the dynamic loader does not find code for the provided address)?

mythril/laser/ethereum/state/world_state.py
```diff
@@ -73,26 +73,25 @@ def __copy__(self) -> "WorldState":
...
+        return self.create_account(
+            balance=0,
```
JoranHonig: Should we not take the concrete balance from on-chain?

norhh: yes, Let's add it after this PR.

mythril/laser/ethereum/state/world_state.py
norhh: A bunch of  opcodes push a symbol to stack if this function raises a ValueError.

mythril/laser/ethereum/state/world_state.py
JoranHonig: 👍

mythril/laser/ethereum/state/world_state.py
JoranHonig: Right, you seem to be correct

PR #1315, opened 2020-01-30
Requirement violation

norhh:


mythril/analysis/analysis_module_helpers.py
```diff
@@ -0,0 +1,6 @@
+import traceback
+
+
+def is_prehook() -> bool:
```
norhh: This change isn't required but it makes sense to split it out.

PR #1329, opened 2020-02-12
Loop bound only JUMPIs

norhh:
Quick fix for VerifyEscrow


JoranHonig:
This is also fixed by #1330 right?

norhh:
yes, but do we need to bound JUMPs?

PR #1330, opened 2020-02-13
New algo for creating loop counter keys

b-mueller:
Replaces the loop counter key in the bounded loops module by a checksum over the last 32 pc addresses. This is to prevent unwanted pruning when an internal function is called repeatedly.

### DIFF
--- a/mythril/laser/ethereum/strategy/extensions/bounded_loops.py
+++ b/mythril/laser/ethereum/strategy/extensions/bounded_loops.py
@@ -14,7 +14,7 @@ class JumpdestCountAnnotation(StateAnnotation):
     """State annotation that counts the number of jumps per destination."""
 
     def __init__(self) -> None:
-        self._reached_count = {}  # type: Dict[str, int]
+        self._reached_count = {}  # type: Dict[int, int]
         self.trace = []  # type: List[int]
 
     def __copy__(self):
--- a/mythril/laser/ethereum/strategy/extensions/bounded_loops.py
+++ b/mythril/laser/ethereum/strategy/extensions/bounded_loops.py
@@ -94,7 +94,7 @@ class BoundedLoopsStrategy(BasicSearchStrategy):
                 return state
 
             elif annotation._reached_count[key] > self.bound:
-                log.info("Loop bound reached, skipping state")
+                log.debug("Loop bound reached, skipping state")
                 continue
 
             return state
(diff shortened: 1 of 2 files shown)


wuestholz:
@b-mueller Not sure, but maybe you want to consider fewer than the last 32 locations if those contain the current location. More concretely, assume your locations are [L0, ..., LK, ..., L31] and LK is the last location equal to L31. Then you might want to only compute the checksum based on [LK+1, ..., L31]. What do you think?

b-mueller:
@wuestholz sounds right but I just tried it and somehow incorrect branches get pruned 😅

wuestholz:
Hm, weird... Maybe that's related: it seems like your trace only contains the addresses. Shouldn't it contain PCs or the tuple of address and PC?

norhh:
`Shouldn't it contain PCs or the tuples of address and PC?` is it necessary? Like there is a one to one mapping between addresses and PCs right?

b-mueller:
>  Like there is a one to one mapping between addresses and PCs right? So wouldn't that info be redundant?

That's what I'm thinking. Checksum over the addresses should be sufficient to uniquely identify the path taken?

Loops of length n < 32 should be found after 32/n iterations. I suspect that using [LK+1, ..., L31] somehow creates the original problem where the same jump in an internal function is taken to be a loop. Not 100% sure though...

wuestholz:
Ah, maybe I misunderstood the term "address". I assumed that it refers to the address of the contract. :)

b-mueller:
Classic Mythril terminology!

norhh:
`Loops of length n < 32 should be found after 32/n iterations. I suspect that using [LK+1, ..., L31] somehow creates the original problem where the same jump in an internal function is taken to be a loop.` Won't we have the same problem if loop size >32? , Loop size can easily cross 32 if there are a few array manipulations in the loop.

b-mueller:
I think the crucial value is, if we're looking actual loops vs jumps inside a repeatedly called internal function, the distance from the start of the function to the JUMPDEST, because the path from which the function was reached should be different. Except if the function itself is being called in a loop, in that case it would be the distance to the start of the outer loop.

With a higher _n_  the risk of mistaking a JUMP inside a repeatedly called internal function for a loop decreases. However, it will take longer to identify short loops (since they have to unroll for _n/length_ iterations to generate the same key) so Mythril will overrun the user-defined bound.

Not sure what's a reasonable value here, how does 128 sound? 🙃

PR #1331, opened 2020-02-13
Base implementation for mythril plugins

JoranHonig:
I've implemented the base logic and interface for mythril plugins.  
The first type of plugin that I'd like to support is Detection Modules, to do this neatly I'll refactor the detection module logic a little bit(in a separate module).

## Changes
I've added a plugin module to the mythril namespace. In this module we'll add the required logic for mythril plugin interfaces, and mythril plugin loading. This eventually allows the separation of functionalities in mythril to plugins, and additionally allows the support of external plugins by researchers and developers.

### DIFF
--- a/mythril/plugin/interface.py
+++ b/mythril/plugin/interface.py
@@ -21,7 +21,8 @@ class MythrilPlugin:
         pass
 
     def __repr__(self):
-        pass
+        plugin_name = type(self).__name__
+        return f"{plugin_name} - {self.plugin_version} - {self.author}"
 
 
 class MythrilCLIPlugin(MythrilPlugin):
--- a/mythril/plugin/interface.py
+++ b/mythril/plugin/interface.py
@@ -21,7 +21,7 @@ class MythrilPlugin:
         pass
 
     def __repr__(self):
-        return ""
+        pass
 
 
 class MythrilCLIPlugin(MythrilPlugin):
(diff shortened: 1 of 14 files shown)


tests/plugin/interface_test.py
```diff
@@ -0,0 +1,9 @@
+from mythril.plugin.interface import MythrilCLIPlugin, MythrilPlugin
+
+
+def test_construct_cli_plugin():
+    _ = MythrilCLIPlugin()
+
+
+def test_construct_mythril_plugin():
+    _ = MythrilPlugin
```
norhh: shouldn't an instance be created here? In this case no difference anyway, but if the class is extended in the future it might not get tested.

mythril/plugin/interface.py
```diff
@@ -0,0 +1,34 @@
+from abc import ABC, abstractmethod
```
norhh: is ABC used?

tests/plugin/interface_test.py
JoranHonig: I'm going to make this class abstract, so it won't be constructable anyway. This test is not  very useful 🙃

mythril/plugin/interface.py
JoranHonig: Not yet, that's coming in the next pr

PR #1335, opened 2020-02-17
Clean up detection modules

JoranHonig:
## Description
In this pr I did some cleanup of the detection module interfaces & architecture.

- Move static variables to class scope vs constructor
- Use Enum for entry point over a string
- refactor security.py and symbolic.py with respect to detection module loading and instrumentation

### Refactoring security.py
The logic in this module could be separated in roughly three categories:
- module loading
- module utility methods
- analysis orchestration (retreiving results, and whatnot)

**Module Loading** this logic has now been moved to mythril.analysis.modules.loader, where a singleton instance manages the loaded modules.
Note that I removed the use of custom module directories, this will be replaced soon with the ability to install and enable detection modules through MythrilPlugins.

**Module Utility methods** this logic got moved into mythrill.analysis.modules.util

**Analysis Orchestration** this code got to stay (for now), but looks much cleaner now

### Benefits
The mayor benefit of this archtecture is that we can now easily "register" new detection modules. A mythril plugin will only have to execute `ModuleLoader().register_module(MyCoolModule())` to add a module that mythril will analyze.

This can also be automated even further, we'll see how that plays out

### To Do
- [x] document all the things
- [x] fix all the tests


mythril/analysis/module/base.py
```diff
@@ -0,0 +1,94 @@
...
+    - entry_point: Mythril can run callback style detection modules, or modules that search the statespace.
+                [IMPORTANT] POST entry points severely slow down the analysis, try to always use callback style modules
+    - pre_hooks: A list of instructions to hook the laser vm for (pre execution of the instruction)
+    - post_hooks: A list of instructions to hook the laser vm for (post execution of the instruction)
+    """
+
+    name = "Detection Module Name"
+    swc_id = "SWC-000"
+    description = "Detection module description"
+    entry_point = EntryPoint.POST  # type: EntryPoint
```
norhh: it makes sense if it's defaulted to CALLBACK right? as all the modules now are callback.

mythril/analysis/module/modules/user_assertions.py
```diff
@@ -28,15 +28,11 @@
...
-    def __init__(self):
-        """"""
-        super().__init__(
-            name="External calls",
-            swc_id=ASSERT_VIOLATION,
-            description=DESCRIPTION,
-            entrypoint="callback",
-            pre_hooks=["LOG1"],
-        )
+    name = "External calls"
```
norhh: This is a late find, this isn't `External calls` right? :laughing:

PR #1340, opened 2020-02-24
Modules selection usability overhaul

b-mueller:
I refactored the class names and "name" fields of each module for consistency and added the following:

1. Users can now select modules by classname:

```
$ myth analyze test.sol -mIntegerArithmetics,Exceptions
```

2. Adding a non-existent module throws an error:

```
$ myth analyze test.sol -mIntegerArithmetics,Exceptions,Lol
mythril.interfaces.cli [ERROR]: Invalid detection module: Lol 
```

3. The new command "list-detectors" provides a list of the available modules:

```
$ myth list-detectors
ArbitraryJump: Jump to an arbitrary bytecode location
ArbitraryStorage: Write to an arbitrary storage location
ArbitraryDelegateCall: Delegatecall to a user-specified address
PredictableVariables: Control flow depends on a predictable environment variable
DeprecatedOperations: Usage of deprecated instructions
EtherThief: Attacker can profitably withdraw Ether from the contract account
Exceptions: Exception or assertion violation
ExternalCalls: External call to another contract
IntegerArithmetics: Integer overflow or underflow
MultipleSends: Multiple external calls in the same transaction
StateChangeAfterCall: State change after external call
AccidentallyKillable: Contract can be accidentally killed by anyone
UncheckedRetval: Return value of an external call is not checked
UserAssertions: A user-defined assertion has been triggered
```

### DIFF
--- a/mythril/analysis/module/modules/state_change_external_calls.py
+++ b/mythril/analysis/module/modules/state_change_external_calls.py
@@ -99,7 +99,7 @@ class StateChangeAfterCall(DetectionModule):
     """This module searches for state change after low level calls (e.g. call.value()) that
     forward gas to the callee."""
 
-    name = "State Change After External calls"
+    name = "State change after external call"
     swc_id = REENTRANCY
     description = DESCRIPTION
     entry_point = EntryPoint.CALLBACK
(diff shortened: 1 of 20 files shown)


mythril/analysis/module/modules/state_change_external_calls.py
```diff
@@ -95,11 +95,11 @@ def get_issue(
         )
 
 
-class StateChange(DetectionModule):
+class StateChangeAfterCall(DetectionModule):
     """This module searches for state change after low level calls (e.g. call.value()) that
     forward gas to the callee."""
 
-    name = "State Change After External calls"
+    name = "State change after external call"
```
norhh:
```suggestion
    name = "State change after an external call"
```

PR #1344, opened 2020-03-01
Load intial account balances from blockchain and refactor loader flags

b-mueller:
- Adds new function `read_balance(address` to dynloader
- Account constructor now reads intial account balance from blockchain if dynloader is available
- Minor: Don't pass dynloader to `CREATOR_ACCOUNT` and `ATTACKER_ACCOUNT` which are virtual
- Use a single boolean / flag to determine if data should be loaded instead of separating code/storage/balances:

```
  --no-onchain-data. Don't attempt to retrieve contract code, variables and balances from the blockchain
```

### DIFF
--- a/mythril/support/loader.py
+++ b/mythril/support/loader.py
@@ -49,9 +49,7 @@ class DynLoader:
         :return:
         """
         if not self.active:
-            raise ValueError(
-                "Cannot load from storage when the loader is disabled"
-            )
+            raise ValueError("Cannot load from storage when the loader is disabled")
         if not self.eth:
             raise ValueError("Cannot load from the chain when eth is None")
 
--- a/mythril/support/loader.py
+++ b/mythril/support/loader.py
@@ -48,7 +48,7 @@ class DynLoader:
         :param address:
         :return:
         """
-        if not self.storage_loading:
+        if not self.active:
             raise ValueError(
                 "Cannot load from the storage when the storage_loading flag is false"
             )
--- a/mythril/support/loader.py
+++ b/mythril/support/loader.py
@@ -53,7 +53,7 @@ class DynLoader:
                 "Cannot load from storage when the loader is disabled"
             )
         if not self.eth:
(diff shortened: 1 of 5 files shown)


mythril/support/loader.py
```diff
@@ -47,6 +47,22 @@ def read_storage(self, contract_address: str, index: int) -> str:
...
         )
 
+    @functools.lru_cache(LRU_CACHE_SIZE)
+    def read_balance(self, address: str) -> str:
+        """
+
+        :param address:
+        :return:
+        """
+        if not self.storage_loading:
```
norhh: we should change this variable to `onchain_loading` as this isn't storage right?

b-mueller: I simplified this to a single boolean (see updated PR)

mythril/support/loader.py
```diff
@@ -36,25 +32,39 @@ def read_storage(self, contract_address: str, index: int) -> str:
...
+    @functools.lru_cache(LRU_CACHE_SIZE)
+    def read_balance(self, address: str) -> str:
+        """
+
+        :param address:
+        :return:
+        """
+        if not self.active:
+            raise ValueError(
+                "Cannot load from the storage when the storage_loading flag is false"
```
norhh:
```suggestion
                "Cannot load from storage when the loader is disabled"
```

mythril/support/loader.py
```diff
@@ -36,25 +32,39 @@ def read_storage(self, contract_address: str, index: int) -> str:
...
+
+        :param address:
+        :return:
+        """
+        if not self.active:
+            raise ValueError(
+                "Cannot load from storage when the loader is disabled"
+            )
+        if not self.eth:
+            raise ValueError("Cannot load from the storage when eth is None")
```
norhh:
```suggestion
            raise ValueError("Cannot load from the chain when eth is None")
```

PR #1351, opened 2020-03-06
Simplify EtherThief and fix pruning behaviour

b-mueller:
- Revert to hooking `CALL` and `STATICCALL` in EtherThief module
- Prevent`MutationPruner` from pruning paths that where ETH could have been received via msg.value

**TESTS**

**No payable functions:**

```
pragma solidity ^0.5.0;

contract SimpleEtherDrain {

  function withdrawAllAnyone() public {
  }
}
```
```
$ $MYTH a SimpleEtherDrain.sol -mEtherThief
```

**Payable constructor:**

```
pragma solidity ^0.5.0;

contract SimpleEtherDrain {

  constructor() public payable {
  }

  function withdrawAllAnyone() public {
  }
}
```
```
$ $MYTH a SimpleEtherDrain.sol -mEtherThief -t1
==== Unprotected Ether Withdrawal ====
SWC ID: 105
Severity: High
Contract: SimpleEtherDrain
Function name: withdrawAllAnyone()
PC address: 102
Estimated Gas Usage: 924 - 35205
Anyone can withdraw ETH from the contract account.
Arbitrary senders other than the contract creator can withdraw ETH from the contract account. This is likely to be a vulnerability.
--------------------
In file: SimpleEtherDrain.sol:9

msg.sender.transfer(address(this).balance)

--------------------
Initial State:

Account: [CREATOR], balance: 0x8, nonce:0, storage:{}
Account: [ATTACKER], balance: 0x2000, nonce:0, storage:{}
Account: [SOMEGUY], balance: 0x0, nonce:0, storage:{}

Transaction Sequence:

Caller: [CREATOR], calldata: , value: 0x1
Caller: [ATTACKER], function: withdrawAllAnyone(), txdata: 0x6aba6fa1, value: 0x0
```

**No payable constructor + payable fallback function:**

```
pragma solidity ^0.5.0;

contract SimpleEtherDrain {

  constructor() public payable {
  }

  function withdrawAllAnyone() public {
  }
}
```
```
$ $MYTH a SimpleEtherDrain.sol -mEtherThief -t1
```
```
(mythril) Bernhards-MacBook-Pro:Desktop bernhardmueller$ $MYTH a SimpleEtherDrain.sol -mEtherThief -t2
==== Unprotected Ether Withdrawal ====
SWC ID: 105
Severity: High
Contract: SimpleEtherDrain
Function name: withdrawAllAnyone()
PC address: 99
Estimated Gas Usage: 924 - 35205
Anyone can withdraw ETH from the contract account.
Arbitrary senders other than the contract creator can withdraw ETH from the contract account. This is likely to be a vulnerability.
--------------------
In file: SimpleEtherDrain.sol:6

msg.sender.transfer(address(this).balance)

--------------------
Initial State:

Account: [CREATOR], balance: 0x0, nonce:0, storage:{}
Account: [ATTACKER], balance: 0x2150200048852400, nonce:0, storage:{}
Account: [SOMEGUY], balance: 0x0, nonce:0, storage:{}

Transaction Sequence:

Caller: [CREATOR], calldata: , value: 0x0
Caller: [SOMEGUY], function: unknown, txdata: 0x40000402, value: 0x1
Caller: [ATTACKER], function: withdrawAllAnyone(), txdata: 0x6aba6fa1, value: 0x0
```

### DIFF
--- a/mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
+++ b/mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
@@ -3,12 +3,14 @@ from mythril.laser.ethereum.plugins.plugin import LaserPlugin
 from mythril.laser.ethereum.plugins.implementations.plugin_annotations import (
     MutationAnnotation,
 )
+from mythril.laser.smt.bitvec_helper import UGT
 from mythril.laser.ethereum.state.global_state import GlobalState
 from mythril.laser.ethereum.svm import LaserEVM
 from mythril.laser.ethereum.transaction.transaction_models import (
     ContractCreationTransaction,
 )
-from mythril.laser.smt import And, symbol_factory
+from mythril.analysis import solver
+from mythril.exceptions import UnsatError, CriticalError
 
 
 class MutationPruner(LaserPlugin):
@@ -38,27 +40,23 @@ class MutationPruner(LaserPlugin):
         def sstore_mutator_hook(global_state: GlobalState):
             global_state.annotate(MutationAnnotation())
 
-        @symbolic_vm.pre_hook("CALL")
-        def call_mutator_hook(global_state: GlobalState):
-            global_state.annotate(MutationAnnotation())
-
-        @symbolic_vm.pre_hook("STATICCALL")
-        def staticcall_mutator_hook(global_state: GlobalState):
-            global_state.annotate(MutationAnnotation())
-
         @symbolic_vm.laser_hook("add_world_state")
         def world_state_filter_hook(global_state: GlobalState):
-            if And(
-                *global_state.world_state.constraints[:]
-                + [
-                    global_state.environment.callvalue
-                    > symbol_factory.BitVecVal(0, 256)
-                ]
-            ).is_false:
-                return
+
             if isinstance(
                 global_state.current_transaction, ContractCreationTransaction
             ):
                 return
+
+            try:
+                constraints = global_state.world_state.constraints + [
+                    global_state.environment.callvalue > 0
+                ]
+
+                solver.get_model(constraints)
+                return
+            except UnsatError:
+                pass
+
             if len(list(global_state.get_annotations(MutationAnnotation))) == 0:
                 raise PluginSkipWorldState
--- a/mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
+++ b/mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
@@ -67,6 +67,7 @@ class MutationPruner(LaserPlugin):
                 solver.get_model(constraints)
                 return
             except UnsatError:
+                # callvalue is constrained to 0, therefore there is no balance based world state mutation
                 pass
 
             if len(list(global_state.get_annotations(MutationAnnotation))) == 0:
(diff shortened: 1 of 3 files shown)


mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
```diff
@@ -38,6 +39,10 @@ def initialize(self, symbolic_vm: LaserEVM):
         def sstore_mutator_hook(global_state: GlobalState):
             global_state.annotate(MutationAnnotation())
 
+        """FIXME: Check for changes in world_state.balances instead of adding MutationAnnotation for all calls.
```
JoranHonig: I don't think that works. 
Each transaction will by default modify the balances array. Just for unpayable ones the constraint `value == 0` get's added. Which then means the balances won't change.

So I think we can remove the fixme :sweat_smile:

mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
```diff
@@ -48,17 +53,21 @@ def staticcall_mutator_hook(global_state: GlobalState):
...
-                return
+
             if isinstance(
                 global_state.current_transaction, ContractCreationTransaction
             ):
                 return
+
+            try:
+                constraints = global_state.world_state.constraints + [
+                    global_state.environment.callvalue > 0
```
JoranHonig: I think we should use UGE here, because callvalue is an unsigned value

mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
```diff
@@ -48,17 +53,21 @@ def staticcall_mutator_hook(global_state: GlobalState):
...
+                ]
+
+                solver.get_model(constraints)
+                return
+            except UnsatError:
+                pass
```
JoranHonig:
```suggestion
                # callvalue is constrained to 0, therefore there is no balance based world state mutation
                pass
```

JoranHonig: Maybe a comment like this is useful to indicate the edge case that is introduced here

b-mueller:
Part of this fix would be storing starting_balances *after* msg.value has
been added to the callee balance. Then we can ask the solver things like
"is it possible for balances to change in any way in the course of this
tx". If there's no value calls and the constraint [msg.value==0] on the end
state the answer would be "no" amirite?

mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
b-mueller: Part of this fix would be storing starting_balances *after* msg.value has
been added to the callee balance. Then we can ask the solver things like
"is it possible for balances to change in any way in the course of this
tx". If there's no value calls and the constraint [msg.value==0] on the end
state the answer would be "no" amirite?

mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
b-mueller: UGT right? Done but it required some ugly type normalisation.

mythril/laser/ethereum/plugins/implementations/mutation_pruner.py
JoranHonig: Ah, yeah misunderstanding on my part :sweat_smile:

PR #1353, opened 2020-03-11
Mythril Laser Plugins

JoranHonig:
# Description
You can now install and dynamically load new laser plugins.
There are three main interfaces:

MythrilLaserPlugin:
 - A class that implements this interface exposes a property called `builder` 

PluginBuilder (laser):
 - A plugin builder does what the name suggests. When required (on initialisation of a new symbolic vm). It's called to construct a fresh copy of a laser plugin.

LaserPlugin: 
 - Your trusty old laser plugin, which hasn't changed

We build it this way to allow for the construction of multiple instances of a laser vm object.

## Other changes

1. I moved the plugin module to `mythril.laser`, where it belongs
2. I temporarily removed loading of the coverage based search strategy. 
     This created an unwanted dependency from laser to the plugin's, the loading should be done in symbolic.py or some other area that orchestrates the execution (IMO).


mythril/analysis/symbolic.py
```diff
@@ -126,18 +122,17 @@ def __init__(
             requires_statespace=requires_statespace,
             iprof=iprof,
             enable_coverage_strategy=enable_coverage_strategy,
```
norhh: shouldn't this be removed

PR #1357, opened 2020-03-28
Refactor all detectors

b-mueller:
- Update all issue texts
- Adjust severity of some issues
- Add a module for SWC-115,  use of `tx.origin`for authorisation
- Remove the "deprecated ops" module (SWC-111) which checked for the presence of `callcode` and `origin` instructions. This module is made redundant by the SWC-115 module.
- Simplified the `dependence_on_predictable_vars` module to only check for a `PredictableValueAnnotation` on `JUMPI` condition and do a reachability check on the `JUMPI` instead of following a path to a `STOP` / `CALL` / `RETURN`


mythril/analysis/module/modules/arbitrary_jump.py
```diff
@@ -14,9 +14,9 @@
 
 
 class ArbitraryJump(DetectionModule):
-    """This module searches for JUMPs to an arbitrary instruction."""
+    """This module searches for JUMPs to an user-specified location."""
```
norhh:
```suggestion
    """This module searches for JUMPs to a user-specified location."""
```

mythril/analysis/module/modules/deprecated_ops.py
```diff
@@ -1,85 +0,0 @@
...
-
-
-class DeprecatedOperations(DetectionModule):
-    """This module checks for the usage of deprecated op codes."""
-
-    name = "Usage of deprecated instructions"
-    swc_id = DEPRECATED_FUNCTIONS_USAGE
-    description = DESCRIPTION
-    entry_point = EntryPoint.CALLBACK
-    pre_hooks = ["ORIGIN", "CALLCODE"]
```
norhh: We might miss the CALLCODE if this is removed, right?

norhh:
with this PR we will have a description_head mismatch with harvey. I would be nice if all the decription_heads in a separate file(surely laborious task to move them ), so people who are modifying other tools can simply look into a single file for all the description heads. WDYT?

mythril/analysis/module/modules/deprecated_ops.py
b-mueller: Yes, but the CALLCODE is detection is useless anyway (I haven't even seen this triggered once) and if we wanted to search for deprecated instructions/call types we'd have to include all of them. Using Mythril for this is overkill IMO, simple static Solidity code analysis can do this more efficiently.

PR #1358, opened 2020-03-28
Delay search for plugins

norhh:


mythril/plugin/discovery.py
```diff
@@ -12,21 +12,30 @@ class PluginDiscovery(object, metaclass=Singleton):
     """
 
     # Installed plugins structure. Retrieves all modules that have an entry point for mythril.plugins
-    _installed_plugins = {
-        entry_point.name: entry_point.load()
-        for entry_point in pkg_resources.iter_entry_points("mythril.plugins")
-    }
+    _installed_plugins = {}  # type: Dict[str, Any]
```
JoranHonig:
```suggestion
    _installed_plugins = None  # type: Optional[Dict[str, Any]]
```

mythril/plugin/discovery.py
```diff
@@ -12,21 +12,30 @@ class PluginDiscovery(object, metaclass=Singleton):
...
+
+    def init_installed_plugins(self):
+        self._installed_plugins = {
+            entry_point.name: entry_point.load()
+            for entry_point in pkg_resources.iter_entry_points("mythril.plugins")
+        }
+
+    @property
+    def installed_plugins(self):
+        if self._installed_plugins is {}:
```
JoranHonig:
```suggestion
        if self._installed_plugins is None:
```

mythril/plugin/discovery.py
```diff
@@ -2,7 +2,7 @@
 from mythril.support.support_utils import Singleton
 from mythril.plugin.interface import MythrilPlugin
 
-from typing import List
+from typing import List, Dict, Any
```
JoranHonig:
```suggestion
from typing import List, Dict, Any, Optional
```

mythril/plugin/discovery.py
JoranHonig: Without something like this we could be running the init_installed_plugins forever rite?

PR #1377, opened 2020-05-06
Fix a multicontract bug

norhh:
Fixes a case where 3rd nested call reverts due to incorrect calldata.

### DIFF
--- a/mythril/interfaces/cli.py
+++ b/mythril/interfaces/cli.py
@@ -563,7 +563,7 @@ def set_config(args: Namespace):
     ):
         config.set_api_from_config_path()
 
-    if args.__dict__.get("address", None):
+    if args.__dict__.get("rpc", None):
         # Establish RPC connection if necessary
         config.set_api_rpc(rpc=args.rpc, rpctls=args.rpctls)
     if args.command in ("hash-to-address", "leveldb-search"):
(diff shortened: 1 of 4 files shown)


mythril/interfaces/cli.py
```diff
@@ -563,7 +563,7 @@ def set_config(args: Namespace):
     ):
         config.set_api_from_config_path()
 
-    if args.__dict__.get("address", None):
+    if args.__dict__.get("rpc", None):
```
b-mueller, line 566: Why this change? We also need to establish an RPC connection to the default URL when there's no `--rpc` arg right?

mythril/laser/ethereum/call.py
```diff
@@ -178,14 +178,6 @@ def get_call_data(
             else memory_size
         ),
     )
-
```
b-mueller, line 181: I'll just trust you that this fixes everything :) Can't replicate it right now because I'd have to context switch my whole test environment

mythril/interfaces/cli.py
norhh: If rpc is not set then it takes the value from config at line 564. Without this we can't use testnets or ganache for queries when using solidity files.

mythril/laser/ethereum/call.py
norhh: I am not sure yet. I didn't run it with the contract on atlassian. This is a bugfix for a tailor made contract which is a bit similar to original contract. But it should atleast fix one problem. Will test it then

PR #1394, opened 2020-06-09
Handle symbolic balance arg

norhh:


ytrezq:
Just out of curiosity. What s the purpose of this commit?

norhh:
account_exists_or_load does not gracefully handle symbolic addresses, This PR handles when the address encountered is symbolic

ytrezq:
What is a symbolic address?

PR #1418, opened 2020-08-02
Support Ifs for arrays and have handy id function for Node

norhh:
- Support Ifs for arrays
- have a function which updates global_id for nodes.


mythril/laser/ethereum/cfg.py
```diff
@@ -55,11 +66,7 @@ def __init__(
         self.function_name = function_name
         self.flags = NodeFlags()
 
-        # Self-assign a unique ID
-        global gbl_next_uid
-
-        self.uid = gbl_next_uid
-        gbl_next_uid += 1
+        self.uid = new_node_uid()
```
JoranHonig, line 69: what is this used for? could we just use `hash(self)`?

mythril/laser/smt/array.py
```diff
@@ -26,9 +24,6 @@ def __getitem__(self, item: BitVec) -> BitVec:
 
     def __setitem__(self, key: BitVec, value: BitVec) -> None:
         """Sets an item in the array, key can be symbolic."""
-        if isinstance(value, Bool):
```
JoranHonig, line 29: If you move this logic to the storage object, then it should also be duplicated in memory right?

What is the benefit to moving it up to the datamodels?

mythril/laser/smt/array_helper.py
```diff
@@ -0,0 +1,12 @@
+from typing import Union
+import z3
+
+from mythril.laser.smt.array import Array
+
+
+def z3_array_converter(array: Union[z3.Array, z3.K]):
```
JoranHonig, line 7: missing return value

tests/laser/test_transaction.py
```diff
@@ -8,7 +8,7 @@
 
 def test_intercontract_call():
     # Arrange
-    cfg.gbl_next_uid = 0
+    cfg.new_node_uid.gbl_next_uid = -1
```
JoranHonig, line 11: Why make this -1, when the class initializes it to 0?

mythril/laser/smt/bitvec_helper.py
```diff
@@ -32,6 +48,11 @@ def If(a: Union[Bool, bool], b: Union[BitVec, int], c: Union[BitVec, int]) -> Bi
     """
     if not isinstance(a, Bool):
         a = Bool(z3.BoolVal(a))
+
+    if isinstance(b, BaseArray) and isinstance(c, BaseArray):
+        array = z3.If(a.raw, b.raw, c.raw)
+        return z3_array_converter(array)
```
JoranHonig, line 54: Is z3_array_converter going to be used anywhere else? 

Otherwise it might be better to not separate this in a function / make z3_array_converter a private function in `bitvec_helper.py` WDYT?

mythril/laser/smt/array.py
norhh: There was a problem with circular imports. So I had to move it. Memory currently uses Dict so there won’t be a problem with duplication for now

tests/laser/test_transaction.py
norhh: The new uid function call increments and returns it, so it had to be different

mythril/laser/smt/bitvec_helper.py
norhh: The file name was bitvec_helper.py so I didn’t want to place it here as the function was array specific. But we can make it a private function in this file.

mythril/laser/ethereum/cfg.py
norhh: Sequential node id was useful to read statespace in ancient times(now statespace is unreadable because storage and symbolic balances), So we can replace it with hash.

mythril/laser/smt/bitvec_helper.py
```diff
@@ -3,10 +3,19 @@
 
 from mythril.laser.smt.bool import Bool, Or
 from mythril.laser.smt.bitvec import BitVec
+from mythril.laser.smt.array import BaseArray, Array
 
 Annotations = Set[Any]
 
 
+def _z3_array_converter(array: Union[z3.Array, z3.K]) -> Array:
```
JoranHonig, line 11: This is a helper for arrays, should it be in a file other than `bitvec_helper.py`?

JoranHonig: I should have looked at the comments I left before. This is  a duplicate

JoranHonig, line 11:
```suggestion
def _z3_array_converter(array: Union[z3.Array, z3.K]) -> BaseArray:
```

JoranHonig (approved):

PR #1419, opened 2020-08-02
Support merge annotations

norhh:


mythril/laser/plugin/plugins/instruction_profiler.py
```diff
@@ -32,7 +32,7 @@
 
 
 class InstructionProfilerBuilder(PluginBuilder):
-    plugin_name = "dependency-pruner"
+    plugin_name = "instruction-profiler"
```
JoranHonig, line 35: good catch !

mythril/laser/ethereum/state/annotation.py
```diff
@@ -34,6 +36,23 @@ def persist_over_calls(self) -> bool:
         """
         return False
 
+    def check_merge_annotation(self, annotation) -> bool:
+        """Used by State merging plugin.
```
JoranHonig, line 40:
```suggestion
        """Checks whether this annotation can be merged with annotation
```

mythril/laser/ethereum/state/annotation.py
```diff
@@ -34,6 +36,23 @@ def persist_over_calls(self) -> bool:
...
+
+        The default is set to false. Using this function
```
JoranHonig, line 42:
```suggestion
        The default is set to false.
```

mythril/laser/ethereum/state/annotation.py
```diff
@@ -34,6 +36,23 @@ def persist_over_calls(self) -> bool:
         """
         return False
 
+    def check_merge_annotation(self, annotation) -> bool:
```
JoranHonig, line 39: Shouldn't this be part of the MergeableStateAnnotation class?

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -42,18 +48,31 @@ def __copy__(self):
...
 
     def extend_storage_write_cache(self, iteration: int, value: object):
         if iteration not in self.storage_written:
-            self.storage_written[iteration] = [value]
+            self.storage_written[iteration] = {value}
         elif value not in self.storage_written[iteration]:
-            self.storage_written[iteration].append(value)
+            self.storage_written[iteration].add(value)
+
+    def check_merge_annotation(self, other):
```
JoranHonig, line 61: This should include a check to assert that other is a DependencyAnnotation

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -42,18 +48,31 @@ def __copy__(self):
 
     def get_storage_write_cache(self, iteration: int):
```
JoranHonig, line 49:
```suggestion
    def get_storage_write_cache(self, iteration: int):
        return self.storage_written.get(iteration, set())
```
We can clean this function up by using something like this

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -42,18 +48,31 @@ def __copy__(self):
...
         if iteration not in self.storage_written:
-            self.storage_written[iteration] = []
+            self.storage_written[iteration] = set()
 
         return self.storage_written[iteration]
 
     def extend_storage_write_cache(self, iteration: int, value: object):
         if iteration not in self.storage_written:
```
JoranHonig, line 56: The same would probably work here

```suggestion
        self.storage_written.get(iteration, set()).add(value)
```

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -18,15 +24,15 @@ def persist_over_calls(self) -> bool:
...
 
-class DependencyAnnotation(StateAnnotation):
+class DependencyAnnotation(MergeableStateAnnoation):
     """Dependency Annotation
 
     This annotation tracks read and write access to the state during each transaction.
     """
 
     def __init__(self):
         self.storage_loaded = []  # type: List
```
JoranHonig, line 34: Could this be a set as well?

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -42,18 +48,31 @@ def __copy__(self):
...
-            self.storage_written[iteration] = [value]
+            self.storage_written[iteration] = {value}
         elif value not in self.storage_written[iteration]:
-            self.storage_written[iteration].append(value)
+            self.storage_written[iteration].add(value)
+
+    def check_merge_annotation(self, other):
+        return self.has_call == other.has_call and self.path == other.path
+
+    def merge_annotation(self, other: "DependencyAnnotation"):
```
JoranHonig, line 64: What do you think about having merge_annotation create a new annotation? Vs mutating the current one?

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -42,18 +48,31 @@ def __copy__(self):
...
+        self.blocks_seen = self.blocks_seen.union(other.blocks_seen)
+        for v in other.storage_loaded:
+            if v not in self.storage_loaded:
+                self.storage_loaded.append(v)
+        for key, val in other.storage_written.items():
+            if key not in self.storage_written:
+                self.storage_written[key] = val
+            self.storage_written[key] = self.storage_written[key].union(val)
```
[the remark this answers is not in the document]
JoranHonig: same here

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -67,3 +86,30 @@ def __copy__(self):
...
+                    "check_merge_annotation() method doesn't exist "
+                    "for the annotation {}. Aborting merge for the state".format(
+                        type(a1)
+                    )
+                )
+                return False
+        return True
+
+    def merge_annotation(self, annotation: "WSDependencyAnnotation"):
+        for a1, a2 in zip(self.annotations_stack, annotation.annotations_stack):
```
JoranHonig, line 112: What happens if you merge annotations for two states which have seen different amounts of transactions?

I think in this case, their respective annotation stacks will look like:
[a, b]
[c, d, e]

I believe using zip, would ignore `e`

JoranHonig: nvm, this is captured by the check_merge_annotation 👍

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -67,3 +86,30 @@ def __copy__(self):
         result = WSDependencyAnnotation()
         result.annotations_stack = copy(self.annotations_stack)
         return result
+
+    def check_merge_annotation(self, annotation: "WSDependencyAnnotation") -> bool:
+        if len(self.annotations_stack) != len(annotation.annotations_stack):
```
JoranHonig, line 91:
```suggestion
        if len(self.annotations_stack) != len(annotation.annotations_stack):
            # We can only merge worldstate annotations that have seen an equal amount of transactions
            # since the beginning of symbolic execution
```

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -67,3 +86,30 @@ def __copy__(self):
...
+            return False
+        for a1, a2 in zip(self.annotations_stack, annotation.annotations_stack):
+            if a1 == a2:
+                continue
+            if type(a1) != type(a2):
```
JoranHonig, line 96: Shouldn't all the items in annotations_stack be DependencyAnnotations?

If so, we should add it to the typedef in `__init__` & then we don't need this clause

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -67,3 +86,30 @@ def __copy__(self):
...
+                return False
+            try:
```
JoranHonig, line 98:
```suggestion
            if isinstance(a1, MergeableAnnotation) and isinstance(a2, MergeableAnnotation)
```
It is probably cleaner to perform the check like this vs waiting on an exception 😅

mythril/laser/plugin/plugins/plugin_annotations.py
norhh: Yes, we can change it in a different PR, as I should also make changes in the dependency pruning file for this.

mythril/laser/ethereum/state/annotation.py
norhh: I added it here because StateAnnotation cannot be merged.

mythril/laser/plugin/plugins/plugin_annotations.py
norhh: Nevermind, making it a set seems easier when merging two annotations into a new variable.

mythril/laser/ethereum/state/annotation.py
JoranHonig: This adds an implicit dependency from `StateAnnotation` to `MergeableStateAnnotation`. 

It might be cleaner if callers of `check_merge_annotation()` would do a check like:`annotation instanceof MergeableStateAnnotation and annotation.check_merge_annotation(other)`

JoranHonig (changes requested):
one small suggestion left in annotation.py

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -41,29 +47,79 @@ def __copy__(self):
...
+            # We can only merge worldstate annotations that have seen an equal amount of transactions
+            # since the beginning of symbolic execution
+            return False
+        for a1, a2 in zip(self.annotations_stack, annotation.annotations_stack):
+            if a1 == a2:
+                continue
+            if isinstance(a1, MergeableStateAnnotation) and isinstance(
+                a2, MergeableStateAnnotation
+            ):
+                if a1.check_merge_annotation(a2) is False:
```
JoranHonig, line 104: Theis nested ifstatement can be merged because of equivalent else branche

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -41,29 +47,79 @@ def __copy__(self):
...
 
     def extend_storage_write_cache(self, iteration: int, value: object):
-        if iteration not in self.storage_written:
-            self.storage_written[iteration] = [value]
-        elif value not in self.storage_written[iteration]:
-            self.storage_written[iteration].append(value)
-
-
-class WSDependencyAnnotation(StateAnnotation):
+        self.storage_written.get(iteration, set()).add(value)
```
JoranHonig, line 53: This doesn't work when the default value is passed
image

mythril/laser/plugin/plugins/plugin_annotations.py
```diff
@@ -41,29 +47,79 @@ def __copy__(self):
...
+
+    def check_merge_annotation(self, other: "DependencyAnnotation"):
+        assert isinstance(other, DependencyAnnotation)
```
JoranHonig, line 56:
```suggestion
        if not isinstance(other, DependencyAnnotation):
            raise TypeError("Expected instance of DependencyAnnotation")
```
I think this should raise a type error https://docs.python.org/3/library/exceptions.html#TypeError

mythril/laser/ethereum/state/annotation.py
```diff
@@ -35,6 +37,16 @@ def persist_over_calls(self) -> bool:
         return False
 
 
+class MergeableStateAnnotation(StateAnnotation):
```
JoranHonig, line 40: Could you add documentation to this class?

JoranHonig (changes requested):

mythril/laser/plugin/plugins/plugin_annotations.py
norhh: 😱

JoranHonig (changes requested):

JoranHonig (approved):

Issue #1424, opened 2020-09-09
--solv throws various errors when not using the docker image, with fix


jimtje:
## Description

Hey guys. I ran into some issues while using the --solv switch while running the analysis module. Good news is that I have a fix ready for a pull request whenever you guys see fit.

### Errors

The first is that when using --solv, py-solc-x then proceeds to compile, and on MacOS Catalina, if brew runs into any sort of error, it would throw exceptions that doesn't get picked up correctly through the shell scripts and the whole thing simply throws a generic error.

```
mythril.interfaces.cli [ERROR]: Traceback (most recent call last):
  File "/Users/jim/opt/anaconda3/lib/python3.7/site-packages/mythril/interfaces/cli.py", line 842, in parse_args_and_execute
    enable_online_lookup=query_signature,
  File "/Users/jim/opt/anaconda3/lib/python3.7/site-packages/mythril/mythril/mythril_disassembler.py", line 38, in __init__
    self.solc_binary = self._init_solc_binary(solc_version)
  File "/Users/jim/opt/anaconda3/lib/python3.7/site-packages/mythril/mythril/mythril_disassembler.py", line 83, in _init_solc_binary
    solcx.install_solc("v" + version)
  File "/Users/jim/opt/anaconda3/lib/python3.7/site-packages/solcx/install.py", line 204, in install_solc
    _install_solc_osx(version, allow_osx)
  File "/Users/jim/opt/anaconda3/lib/python3.7/site-packages/solcx/install.py", line 312, in _install_solc_osx
    message="Running dependency installation script `install_deps.sh` @ {}".format(temp_path)
  File "/Users/jim/opt/anaconda3/lib/python3.7/site-packages/solcx/install.py", line 232, in _check_subprocess_call
    **proc_kwargs
  File "/Users/jim/opt/anaconda3/lib/python3.7/subprocess.py", line 347, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['sh', '/var/folders/db/pxs8v9056zs9sdcgs18wc9m80000gn/T/py-solc-x-tmp/solidity_0.6.10/scripts/install_deps.sh']' returned non-zero exit status 1.
```
[...]

PR #1425, opened 2020-09-09
Fixing py-solc-x usage to automatically download or compile correct solc binary and use it

jimtje:
#1424


mythril/ethereum/util.py
```diff
@@ -138,7 +139,20 @@ def solc_exists(version):
     elif sys.version_info[1] >= 6:
         # we are using solc-x for the the 0.5 and higher
         try:
-            return solcx.install.get_executable(version)
+            if platform.system() == "Darwin":
```
norhh, line 142: This part can be made a function which can be reused in the mythril_disassembler.py right?

jimtje: Yeah I wasn't quite sure if that was used anywhere else, but yes effectively it's the same code twice. The code in mythril_disassembler.py throws exception if no version is found either on the system or for download

norhh:
Looks good, but It would be good but many people specify versions as `--solc v0.5.0` , for that something like 
```
if version[0] != "v":
    version = "v"+version 
``` 
being at the beginning of the function would be good. This will also reduce the number of `"v"+ version` in the code.

jimtje:
> Looks good, but It would be good but many people specify versions as `--solc v0.5.0` , for that something like
> 
> ```
> if version[0] != "v":
>     version = "v"+version 
> ```
> 
> being at the beginning of the function would be good. This will also reduce the number of `"v"+ version` in the code.

I think the problem with that is because I'm doing version checks with semver for compatibility, it pretty much either needs a "v" + in front of half of the times I reference version, or I have to do version[1:] for the version checking after I add the "v" to the version anyway. Both are used 3-4 times or so I think. I think that in mythril_disassembler.py since version is only compared to the version number extracted by regex, it makes sense slice the input to cut out the leading "v" if it's there. But I'm not sure it's necessarily better to use Version(version[1:]) instead of solcx.install("v" + version) in utils.py, would have to fiddle with the two requiring different forms of input in either case. 🤷‍♂️

norhh (approved):
Great work, thanks for the PR 🚀

PR #1433, opened 2020-10-07
use nano seconds

norhh:


mythril/analysis/report.py
```diff
@@ -287,11 +287,12 @@ def as_swc_standard_format(self):
         meta_data.update(self._get_exception_data())
 
         # Add execution info to meta
-        meta_data["mythril_execution_info"] = {}
+        meta_data["mythril_execution_info"] = {
+            "analysis_duration": (time() - StartTime().global_start_time) * (10 ** 9)
```
JoranHonig, line 291:
```suggestion
            "analysis_duration": int(round((time() - StartTime().global_start_time) * (10 ** 9)))
```
Add rounding and cast because the return value should be an integer

JoranHonig, line 291:
```suggestion
            "analysis_duration": int(round((time() - StartTime().global_start_time) * (10 ** 9)))
```
Add rounding and cast because the return value should be an integer

JoranHonig (changes requested):

JoranHonig (approved):

PR #1454, opened 2021-01-02
s/tostring/tobytes/

snf:
tostring() has been deprecated since 3.2 and removed in 3.9

ref: https://docs.python.org/3/library/array.html#array.array.tobytes


AndreMiras:
Thanks for the PR.
Note the `tostring()` method was eventually removed in Python 3.9 hence the script would crash without this PR on 3.9

norhh:
Thanks for the PR

PR #1469, opened 2021-03-29
Fix edge case where sign_bit_set is symbolic

JoranHonig:
This pullrequest adds more accurate handling of the signextend instruction modelling.

In particular, it accepts situations where the sign bit is symbolic.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -4,13 +4,12 @@ coverage
 py_ecc==1.6.0
 eth_abi==1.3.0
 eth-account>=0.1.0a2,<=0.3.0
-ethereum>=2.3.2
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -19,7 +18,6 @@ numpy==1.19.0
 persistent>=4.2.0
 plyvel
 py-flags
-py-evm==0.3.0a13
```
norhh, line 22: native contracts depend on py-evm right?

JoranHonig: oh oops, didn't mean to include these changes

norhh (approved):
👍

PR #1490, opened 2021-07-03
Bugfix for lost leading zeros in address, refactor

bogatyy:
Noticed that when my contract had a "0x0ABCD..." dependency, Mythril sent the eth_getBalance request to my RPC with address "0xABCD.." instead.
Naturally, my RPC returned an error. Ultimately, there is an error somewhere between converting integers to strings, that loses leading zeros on addresses.

This patch fixes this bug, and further simplifies the two cases of isinstance(addr, int) and isinstance(addr, str) by merging them. Less duplicated code, less bugs.

### DIFF
--- a/mythril/laser/ethereum/state/world_state.py
+++ b/mythril/laser/ethereum/state/world_state.py
@@ -82,8 +82,9 @@ class WorldState:
         """
 
         if isinstance(addr, str):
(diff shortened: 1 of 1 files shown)


norhh (approved):
Cool! Thanks for the PR

PR #1491, opened 2021-07-05
Changes most fixed dependencies

gnarvaja:
The previous requirements.txt had many fixed dependencies and I think
most of them don't need to be fixed.

Those fixed dependencies generate conflicts when mixing mythril with
other python/ethereum packages, notably brownie-eth.


norhh (approved):
Cool, Thanks for the pr

PR #1492, opened 2021-07-06
Change in setup.py requirements

gnarvaja:
Previous setup.py had a duplicated list of dependencies and wasn't using
requirements.txt file. That's why previous commit didn't had effect.

I removed the duplication and now setup.py reads dependencies from
requirements.txt file.

Tested from another environment with brownie-eth installed using `pip install -e git+https://github.com/gnarvaja/mythril.git#egg=mythril` and now works as expected, doesn't produce any conflicts.


norhh:
@gnarvaja reformat the code using black(https://pypi.org/project/black/), this will fix the issue with the CI.

gnarvaja:
The tests are still failing. I can't see the test logs without CircleCI access... 

Tell me if there is anything else I can do.

norhh:
Unfortunately a dependency is dependent on `requests==2.25.0 `

gnarvaja:
> Unfortunately a dependency is dependent on `requests==2.25.0 `

I fixed requests to 2.25.0 but is still failing. I found that other packages of the _Myth family_ have fixed dependencies that I am almost sure don't need to be fixed:

- https://github.com/dmuhs/pythx/blob/master/requirements.txt
- https://github.com/dmuhs/mythx-models/blob/master/requirements.txt

Perhaps you can try talking with the maintainer of those packages if it's part of ConsenSys.

norhh:
@dmuhs should be able to answer it

norhh:
Hi @gnarvaja , I have removed the dependencies to pythx and it's related features since there are better tools to do the same job, now I think it's possible for you to make changes.

gnarvaja:
I don't know why CicleCI tests are failing. But anyway after the change you did in requirements, this PR is only a cosmetic change, to avoid duplication of dependencies.

norhh:
It's an issue with py-evm, previously we used fixed versions to avoid version conflicts whenever py-evm is upgraded. One way is to find some py-evm version which doesn't lead to this error.
Can you also try logging into circleci via github, I think it should open for you.

norhh:
You should be able to reproduce this issue locally by trying out `python3 setup.py install ` in a new python environment. Now there seems to be an issue with py-ecc

gnarvaja:
Finally checks are OK! 

The change avoids the duplication between `setup.py` dependencies and `requirements.txt`. 

Also removes some dependencies that are not used directly in mythril (will be installed by required packages if those packages keep those dependencies).

And avoids freezing most of the dependencies to a specific version.

requirements.txt
```diff
@@ -1,36 +1,24 @@
...
-pytest_mock
-requests
-rlp>=1.0.1,<2.0.0
-semantic_version
-transaction>=2.2.1
-z3-solver>=4.8.8.0
-pysha3
+ethereum-input-decoder>=0.2.2
 matplotlib
-pre-commit
```
norhh, line 35: Can you add pre-commit, since devs can use it to run black

norhh:
I have removed all the old py-ethereum dependencies and cleaned up, Now it shouldn't cause any issue. If the issue persists feel free to open up an issue.

PR #1507, opened 2021-08-29
Fix bug during checking potential issues

tczpl:
In the original code, Mythril removes the potential_issue in the loop.
However, in Python, during a loop of a list, removing an item will make things go wrong.
Talk is cheap, lets try the following code:

```python
myList = [1,2,3,4,5,6]
for i in myList:
    print(i)
    myList.remove(i)
```

It only prints (1,3,5), and (2,4,6) remains.

The same things go wrong in Mythril, as it just skip some potential issues.
For example, if the len(potential_issues) is 10, it might just check 6 SMT of them.

In Mythril's funciton, after checking a potential_issue, I think there is no need to remove it from the list.
Hence, deleting the ```remove``` code should work for this bug.

### DIFF
--- a/mythril/analysis/potential_issues.py
+++ b/mythril/analysis/potential_issues.py
@@ -81,12 +81,14 @@ def check_potential_issues(state: GlobalState) -> None:
     :return:
     """
     annotation = get_potential_issues_annotation(state)
+    unsat_potential_issues = []
     for potential_issue in annotation.potential_issues:
         try:
             transaction_sequence = get_transaction_sequence(
                 state, state.world_state.constraints + potential_issue.constraints
             )
         except UnsatError:
+            unsat_potential_issues.append(potential_issue)
             continue
 
         potential_issue.detector.cache.add(potential_issue.address)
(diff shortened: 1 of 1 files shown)


mythril/analysis/potential_issues.py
```diff
@@ -89,7 +89,6 @@ def check_potential_issues(state: GlobalState) -> None:
         except UnsatError:
             continue
 
-        annotation.potential_issues.remove(potential_issue)
```
norhh, line 92: Thanks for the catch! But these potential issues are going to persist in the next transactions, and this will cause a slow down to due increased solving time, so can you deep copy those potential_issues in the for loop?

norhh: NVM I think just `copy` should do

tczpl: Thank you. I add a new commit, as it uses an ```unsat```  list to record them and finally assign it to the annotation.
Thank your quick response and help very much!

norhh (approved):
Great!   thanks for the contribution

PR #1622, opened 2022-04-28
Fixed broken links in Analysis Modules of documentation

zobront:
The links to the modules on the [Analysis Modules page](https://mythril-classic.readthedocs.io/en/master/analysis-modules.html) of the documentation used an old file structure with 'analysis/modules/..." instead of "analysis/module/modules/..."


norhh (approved):
Thanks for the contribution 🚀

PR #1632, opened 2022-05-22
Correct comment

72521:


mythril/laser/smt/bitvec_helper.py
```diff
@@ -103,7 +103,7 @@ def ULT(a: BitVec, b: BitVec) -> Bool:
 
 
 def ULE(a: BitVec, b: BitVec) -> Bool:
-    """Create an unsigned less than expression.
+    """Create an unsigned less or equals than expression.
```
norhh, line 106: Nice catch! can you switch it to `less than or equal to expression`

72521: Okay

norhh (approved):
Great! Thanks for the contribution.

PR #1670, opened 2022-08-23
last_jump has type `int` but is used as type `None`.

luca-digrazia:
"filename": "mythril/analysis/module/modules/exceptions.py"
"warning_type": "Incompatible variable type [9]"
"warning_message": " last_jump is declared to have type `int` but is used as type `None`."
"warning_line": 28
"fix": None to 0

### DIFF
--- a/mythril/analysis/module/modules/exceptions.py
+++ b/mythril/analysis/module/modules/exceptions.py
@@ -25,7 +25,7 @@ PANIC_SIGNATURE = [78, 72, 123, 113]
 class LastJumpAnnotation(StateAnnotation):
     """State Annotation used if an overflow is both possible and used in the annotated path"""
 
-    def __init__(self, last_jump: Optiona[int] = None) -> None:
+    def __init__(self, last_jump: Optional[int] = None) -> None:
         self.last_jump: int = last_jump
 
     def __copy__(self):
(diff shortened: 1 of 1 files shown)


mythril/analysis/module/modules/exceptions.py
```diff
@@ -25,7 +25,7 @@
 class LastJumpAnnotation(StateAnnotation):
     """State Annotation used if an overflow is both possible and used in the annotated path"""
 
-    def __init__(self, last_jump: int = None) -> None:
+    def __init__(self, last_jump: int = 0) -> None:
```
norhh, line 28: Changing the type-hint from `int` to `Optional[int]` is more precise than changing the default value.

luca-digrazia: True, `Optional[int]` is a good patch.

mythril/analysis/module/modules/exceptions.py
```diff
@@ -25,7 +25,7 @@
 class LastJumpAnnotation(StateAnnotation):
     """State Annotation used if an overflow is both possible and used in the annotated path"""
 
-    def __init__(self, last_jump: int = None) -> None:
+    def __init__(self, last_jump: Optiona[int] = None) -> None:
```
norhh, line 28:
```suggestion
    def __init__(self, last_jump: Optional[int] = None) -> None:
```

norhh (approved):
👍

Issue #1702, opened 2022-10-27
Crash without internet


gsalzer:
## Description

Mythril throws an exception when started without internet connection, apparently because it tries to load the list of installable solc compilers proactively, even when no compiler download will be needed.

## How to Reproduce

Turn off the internet connection (like the Wifi) and start Mythril. 

```console
$ myth analyze solidity_examples/etherstore.sol
Traceback (most recent call last):
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connection.py", line 174, in _new_conn
    conn = connection.create_connection(
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/util/connection.py", line 72, in create_connection
    for res in socket.getaddrinfo(host, port, family, socket.SOCK_STREAM):
  File "/usr/lib/python3.9/socket.py", line 954, in getaddrinfo
    for res in _socket.getaddrinfo(host, port, family, type, proto, flags):
socket.gaierror: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 703, in urlopen
    httplib_response = self._make_request(
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 386, in _make_request
    self._validate_conn(conn)
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 1042, in _validate_conn
    conn.connect()
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connection.py", line 358, in connect
    self.sock = conn = self._new_conn()
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connection.py", line 186, in _new_conn
    raise NewConnectionError(
urllib3.exceptions.NewConnectionError: <urllib3.connection.HTTPSConnection object at 0x7fbfba1de490>: Failed to establish a new connection: [Errno -2] Name or service not known

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/mythril/venv/lib/python3.9/site-packages/requests/adapters.py", line 489, in send
    resp = conn.urlopen(
  File "/home/mythril/venv/lib/python3.9/site-packages/urllib3/connectionpool.py", line 787, in urlopen
    retries = retries.increment(
[...]

PR #1703, opened 2022-10-27
Fix issue #1702 (crash without internet)

gsalzer:


mythril/ethereum/util.py
```diff
@@ -139,7 +139,11 @@ def solc_exists(version):
     return solc_binary
 
 
-all_versions = solcx.get_installable_solc_versions()
+try:
+    all_versions = solcx.get_installable_solc_versions()
+except:
```
norhh: Adding Something like
```suggestion
except ConnectionError:
```
might be better.
PS: Don't forget to import it from `requests`

PR #1739, opened 2023-02-08
fix returndata copy index

Notation:
wrong index in returndata copy, this may cause index out of bounds while copying


norhh (approved):
Thanks for the fix!

PR #1741, opened 2023-02-18
fix: pad function hashes to always be 8 characters

ToonVanHove:
## Why it's needed
Currenty, when specifying a transaction sequence containing a function hash with leading zeros, eg: `--transaction-sequence '[[0x0651cb35]]'` these leading zeros are dropped because of conversion to and from strings etc.. This causes the call to bytes.fromhex to fail and causes Mythril to crash.

## What I did
Add leading zeros to function hashes so that a function hash is always 8 characters. This allows function hashes with leading zeros to be used.


norhh (approved):
Thanks for the fix!

PR #1770, opened 2023-05-10
Arm64 Docker support

### LINKED ISSUE #1517: Add support for arm64 images
## Description

It would be super cool if docker images are built supporting also arm64 architecture

## Background

I have a raspberry pi and I would love to run docker images on my rasp, however I get the error that Docker Hub does not have an arm64 image, instead only x64 is supported.

## Suggestion

It would be possible to perform multi build using [docker buildx](https://docs.docker.com/buildx/working-with-buildx/)

h4l:
This resolves #1517.

I was trying to use mythril in a vscode devcontainer environment on an ARM Mac. It turned out to be quite a challenge to install mythril because many of your dependencies don't provide arm64 binaries, and some don't publish recent source releases to build from (blake2b).

I also found that your Docker image is only built for amd64, so I tried to build it on arm64 locally, and found it couldn't build for arm64. So set about adjusting the image so that I could use it. Now that I've got a working arm64 image, I figured I should offer the changes back. Hopefully it might make life easier for other people on ARM systems!

I've configured the build with a Docker Buildx bake file, so you can build the mythril/myth image for amd64 and arm64 simultaneously like this:

```console
$ docker buildx bake
```
When you're not doing a CI build, it probably makes more sense to just build the image for your platform, which you can do like this:

```console
$ docker buildx bake --set='*.platform=linux/arm64'
# or
$ docker buildx bake --set='*.platform=linux/amd64'
```
 
 (Also, running with `--print` is useful to understand what bake is going to do, it'll print out the configuration that would be built without actually building.)

The image should be functionally the same as the previous version, but there are a few changes.

- The base image is now python:3.10-slim rather than ubuntu:focal.
- It's smaller now, ~400M arm64 / 500M amd64 vs 1.3G before. This is partly because the base image is a lot smaller, and partly because it's now a multi-stage build that builds wheels in separate stages and copies them into the final stage, so there aren't any dev dependencies/files in the image now.
- It uses svm to download solc versions, both at image build time and in the image entrypoint. This is because the solcx Python package doesn't provide arm64 solc binaries. svm does, and it's also used by foundry.

  This doesn't seem ideal, but hopefully it's a reasonable starting point.
- The image runs as the mythril user rather than root by default. It can still be run as root using the `--user` docker option. This has the potential surface file permission problems for users bind-mounting directories into a myth container that aren't readable by uid 1000.

I also updated `docker_build_and_deploy.sh` to build & publish using `buildx bake`, as shown above. It's possible that your circleci environment will need a tweak. I don't use circleci myself, but they seem to support Docker Buildx: https://support.circleci.com/hc/en-us/articles/360058095471-How-To-Use-Docker-Buildx-in-Remote-Docker-

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,4 @@
+/.*
+/build
+/docker-bake.hcl
+/Dockerfile
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,51 +1,140 @@
-FROM ubuntu:focal
+# syntax=docker/dockerfile:1
+ARG PYTHON_VERSION=3.10
+ARG INSTALLED_SOLC_VERSIONS
 
-ARG DEBIAN_FRONTEND=noninteractive
 
-# Space-separated version string without leading 'v' (e.g. "0.4.21 0.4.22") 
-ARG SOLC
+FROM python:${PYTHON_VERSION:?} AS python-wheel
+WORKDIR /wheels
 
-RUN apt-get update \
-  && apt-get install -y \
-     libsqlite3-0 \
-     libsqlite3-dev \
-  && apt-get install -y \
-     apt-utils \
-     build-essential \
-     locales \
-     python-pip-whl \
-     python3-pip \
-     python3-setuptools \
-     software-properties-common \
-  && add-apt-repository -y ppa:ethereum/ethereum \
-  && apt-get update \
-  && apt-get install -y \
-     solc \
-     libssl-dev \
-     python3-dev \
-     pandoc \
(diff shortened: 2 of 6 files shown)


norhh (approved):
Thank you for the PR

h4l:
No problem, thanks for merging! If you have trouble with building the arm64 version, let me know, as I was experimenting with cross-compiling to avoid needing to emulate arm64 from amd64. I've got a version that can build on GitHub Actions which only has amd64 runners.

PR #1845, opened 2024-03-12
chore: remove repetitive words

suchsoon:


norhh (approved):
Thanks for the contribution! :smile_cat:

PR #1846, opened 2024-03-16
fix typos

xiaoxianBoy:
1. Fix `solidity_examples/rubixi.sol` 
2. Fix `tests/testdata/input_contracts/rubixi.sol`


norhh (approved):
Thank you for the fix :smiley_cat:

PR #1855, opened 2024-04-24
chore: fix some typos in comments

alwayshang:


norhh (approved):
Nice catch! Thank you for the fix!

PR #1864, opened 2024-08-02
Fully switch from mock to unittest.mock

dbast:
mock became part of the Python stdlib in Python 3.3, see
* https://docs.python.org/3/library/unittest.mock.html
* https://mock.readthedocs.io/en/latest/

Some tests already use unittest.mock, this PR now changes all mock imports to unittest.mock and removes mock from the requirements.txt file.


norhh (approved):
Hi @dbast, thank you for the contribution! :+1:

PR #1868, opened 2024-08-08
Update module-list.rst

shikahJS:
Mythril Doc has been updated, adding the three missing detectors.


norhh (approved):
Thank you for the contribution!

PR #1869, opened 2024-08-08
Drop un-used dependencies from requirements.txt

dbast:
Reduces the requirements.txt to only contain runtime dependencies = removes all dependencies, that are either not directly imported or are not a transitive dependency (dependency of a dependency). Test dependencies are anyways defined in [setup.py](https://github.com/Consensys/mythril/blob/develop/setup.py#L56) and [tox.ini](https://github.com/Consensys/mythril/blob/develop/tox.ini#L5).

This got possible via PR https://github.com/Consensys/mythril/pull/1865 where all the un-used imports got identified and removed via pre-commit + ruff.


Dockerfile
```diff
@@ -120,6 +120,12 @@ print("\n".join(str(v) for v in solcx.get_installed_solc_versions()))
   exit 1
 }
 
+# Show installed packages and versions
+pip list
+
+# Show outdated packages
+pip list -o
```
norhh: do we need this part?

dbast: not that important, removed in https://github.com/Consensys/mythril/pull/1869/commits/771171e5a29db828c60f274e5df50098c7d402cd 

keeping the plain `pip list` call as that helps to see what is actually installed in the container during build time.

norhh (approved):
Thanks for the contribution!

PR #1871, opened 2024-08-17
Enable+apply isort via ruff/pre-commit

dbast:
This enables [isort](https://pycqa.github.io/isort/) via ruff/pre-commit and applies it fully automatically (no manual changes), which results in sorting of the imports (no additions or removals).

While this looks like a cosmetic thing, the intention is different:
* Sorting in to sections (stdlib, 3rd party libs, own code..): Ensures that any 3rd party / own code does not overwrite imports from stdlib and the sections also help to see what comes from where.
* Alphabetical ordering within sections: helps to find things within sections, when reading the imports.

This now also runs the black hook after ruff, as reordering imports can result in some work for black.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -10,15 +10,15 @@ repos:
     hooks:
     -   id: check-toml
     -   id: check-yaml
--   repo: https://github.com/psf/black
-    rev: 24.8.0
-    hooks:
-    -   id: black
 -   repo: https://github.com/astral-sh/ruff-pre-commit
     rev: v0.5.6
     hooks:
     -   id: ruff
         args: [--fix, --show-fixes]
+-   repo: https://github.com/psf/black
+    rev: 24.8.0
+    hooks:
+    -   id: black
 -   repo: https://github.com/scop/pre-commit-shfmt
     rev: v3.8.0-1
     hooks:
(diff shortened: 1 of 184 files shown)


.pre-commit-config.yaml
```diff
@@ -10,15 +10,15 @@ repos:
     hooks:
     -   id: check-toml
     -   id: check-yaml
--   repo: https://github.com/psf/black
-    rev: 24.8.0
-    hooks:
-    -   id: black
 -   repo: https://github.com/astral-sh/ruff-pre-commit
```
norhh, line 13: wasn't ruff enabled in the previous commit? Do you know why it did not it apply isort previously?

dbast: yes, but the previous PR only enabled the `F` ruleset, see https://github.com/Consensys/mythril/blob/develop/pyproject.toml#L7 ... to find and remove unused code/imports.

ruff has many more rules that can be activated, see https://docs.astral.sh/ruff/rules/ 

This PR enables isort by adding the `I` to the activated rules, see https://github.com/dbast/mythril/blob/isort/pyproject.toml#L7

norhh: I mean, I don't see that in this PR. Is the `I` in a different PR?

norhh: nvm, found it,
Github compressed some files for me

norhh (approved):
Thanks for the contribution!

PR #1877, opened 2024-08-20
Add pre-commit hook definition + test

dbast:
This adds a `.pre-commit-hooks.yaml` (+ doc + test), which allows other projects to run Mythril via pre-commit.

[pre-commit](https://pre-commit.com/) became pretty popular as framework to combine linters / code-formatters via [hooks](https://pre-commit.com/hooks.html) across different languages. It can be enabled as actual git hook or just used by running `pre-commit run -a`. pre-commit does its own dependency management (+cashing) and installs each hook into its own isolated folder/env.

pre-commit installs a hook by cloning the `repo`, checking out the specified git `rev` (=tag, branch, commit hash), parsing the .pre-commit-hooks.yaml and installing the hook (install method depending on the hook type).

Tested also via my Mythril fork, see https://github.com/dbast/sol-press/pull/17

Requires a new tagged release after the PR merge to be able to reference a tag that contains the hook definition.

### DIFF
--- a/.pre-commit-hooks.yaml
+++ b/.pre-commit-hooks.yaml
@@ -1,0 +1,8 @@
+- id: mythril
+  name: Mythril
+  description: Analyze EVM bytecode with Mythril
+  entry: myth
+  args:
+    - analyze
+  language: python
+  types: ["solidity"]
--- a/README.md
+++ b/README.md
@@ -31,6 +31,15 @@ Install from Pypi (Python 3.7-3.10):
 $ pip3 install mythril
 ```
 
+Use it via pre-commit hook (replace `$GIT_TAG` with real tag):
+
+```YAML
+- repo: https://github.com/Consensys/mythril
+  rev: $GIT_TAG
+  hooks:
+    - id: mythril
+```
+
 See the [docs](https://mythril-classic.readthedocs.io/en/master/installation.html) for more detailed instructions. 
 
 ## Usage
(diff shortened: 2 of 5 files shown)


.pre-commit-hooks.yaml
```diff
@@ -0,0 +1,8 @@
+- id: mythril
+  name: Mythril
+  description: Analyze EVM bytecode with Mythril
+  entry: myth
+  args:
+    - analyze
```
norhh, line 6: isn't this fairly restrictive to just one argument? Mythril has multiple commands

norhh: Adding `disassemble` and `read-storage` might be worth it.

dbast: yeah... didn't want to add a hook for each possible command ... all subcommands can be selected via setting (=overwriting) `args` on user side... maybe worth to add to the README

dbast: added to the README

README.md
```diff
@@ -31,6 +31,17 @@ Install from Pypi (Python 3.7-3.10):
...
+Use it via pre-commit hook (replace `$GIT_TAG` with real tag):
+
+```YAML
+- repo: https://github.com/Consensys/mythril
+  rev: $GIT_TAG
+  hooks:
+    - id: mythril
+```
+
+Additionally set `args: [disassemble]` or  and `args: [read-storage]` to use a different command than `analyze`.
```
norhh:
```suggestion
Additionally, set `args: [disassemble]` or `args: [read-storage]` to use a different command than `analyze`.
```

norhh (approved):
Thanks for the contribution!

PR #1879, opened 2024-08-21
Fix container build: build svm-rs in upstream rust layer

dbast:
A new svm-rs version was released yesterday, which breaks the container build see:
* https://github.com/alloy-rs/svm-rs/tags 
* https://app.circleci.com/pipelines/github/Consensys/mythril/1128/workflows/6ea1da95-5ed5-4154-ab56-04b4916baa6f/jobs/7061

This PR fixes the svm-rs build by building in upstream rust container, which has a more complete rust tooling setup and does not run into the ```error: linking with `cc` failed``` issue.

Also removes some build layers not needed anymore.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,11 +3,9 @@ ARG PYTHON_VERSION=3.10
 ARG INSTALLED_SOLC_VERSIONS
 
 
-FROM python:${PYTHON_VERSION} AS python-wheel
+FROM python:${PYTHON_VERSION} AS mythril-wheels
 WORKDIR /wheels
 
-
-FROM python-wheel AS python-wheel-with-cargo
 # Enable cargo sparse-registry to prevent it using large amounts of memory in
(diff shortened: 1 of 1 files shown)


dbast:
The container build is green again after the recent PR merges. Maybe this PR is not required.

norhh:
This happens due to issues with emulating some arm builds on x86 or emulating some x86 builds on arm (It is a bit tedious to perform multiplatform builds on multiple platforms and combine them). So, it crashes once in a while. Will work when restarted.

dbast:
There was definitely something else as I could reproduce the issue multiple times locally even when building without multiarch emulation... maybe this resolved via some changes in the rustup installer... anyways, it resolved thus closing this PR

norhh:
Isn't it still useful even without the issue?

norhh (approved):

PR #1887, opened 2024-10-30
Fix 'pragma solidity' parsing

gsalzer:
remove comments and strings before searching for a line containing 'pragma solidity'

### DIFF
--- a/mythril/ethereum/util.py
+++ b/mythril/ethereum/util.py
@@ -178,7 +178,7 @@ QUOTE_END = re.compile("(?<!\\\\)'")
 DQUOTE_END = re.compile('(?<!\\\\)"')
 
 def remove_comments_strings(prg):
-    todo = "\n".join(prg) # normalize line ends
+    todo = "\n".join(prg)  # normalize line ends
     done = ""
     while True:
         m = VOID_START.search(todo)
@@ -186,23 +186,23 @@ def remove_comments_strings(prg):
             done += todo
             break
         else:
-            done += todo[:m.start()]
+            done += todo[: m.start()]
             if m[0] == "//":
-                end = todo.find('\n', m.end())
+                end = todo.find("\n", m.end())
                 todo = "" if end == -1 else todo[end:]
             elif m[0] == "/*":
                 end = todo.find("*/", m.end())
                 done += " "
-                todo = "" if end == -1 else todo[end+2:]
+                todo = "" if end == -1 else todo[end + 2 :]
             else:
                 if m[0] == "'":
-                    m2 = QUOTE_END.search(todo[m.end():])
+                    m2 = QUOTE_END.search(todo[m.end() :])
                 else:
-                    m2 = DQUOTE_END.search(todo[m.end():])
+                    m2 = DQUOTE_END.search(todo[m.end() :])
                 if not m2:
                     # unclosed string
                     break
-                todo = todo[m.end()+m2.end():]
+                todo = todo[m.end() + m2.end() :]
     return done
 
 
--- a/mythril/ethereum/util.py
+++ b/mythril/ethereum/util.py
@@ -6,8 +6,8 @@ import json
 import logging
 import os
 import platform
-import typing
 import re
+import typing
 from json.decoder import JSONDecodeError
 from subprocess import PIPE, Popen
 from typing import Tuple
@@ -181,7 +181,6 @@ def remove_comments_strings(program):
     :return: program with strings emptied and comments removed
     :rtype: str
     """
-    
     result = ""
     while True:
         match_start_of_void = VOID_START.search(program)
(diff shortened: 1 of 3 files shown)


gsalzer:
The patch tries to be minimally invasive; you may want to integrate it better. The comment citing the source can be removed (no need to give credits), it is just to inspire trust (?) that the code works since we used it for processing source codes in one of our studies.

norhh:
For some reason, I’ve lost access to the CircleCI pipeline results over the past month. @psantos-consensys, could you look into the cause of CircleCI’s failure? Additionally, could you change the permissions to view Mythril’s CircleCI results? Since Mythril is open-source, these results should ideally be public as they were previously.

gsalzer:
@norhh @psantos-consensys Any chance that something is going to happen in the near future? I'm fine with the pull request being rejected but this CircleCI pipeline issue occurring with every pull request now seems to be unhealthy.

norhh:
I don't know when it will be fixed/visible. I've asked someone to look into it.

psantos-consensys:
@norhh can you please double check if it is working as expected now?

norhh:
@gsalzer, the test failures seem to be due to the test case related to these files: [version_2.sol](https://github.com/Consensys/mythril/blob/develop/tests/testdata/input_contracts/version_2.sol
), [version_3.sol](https://github.com/Consensys/mythril/blob/develop/tests/testdata/input_contracts/version_3.sol).
The comments should be removed from these files.

gsalzer:
> @gsalzer, the test failures seem to be due to the test case related to these files: [version_2.sol](https://github.com/Consensys/mythril/blob/develop/tests/testdata/input_contracts/version_2.sol), [version_3.sol](https://github.com/Consensys/mythril/blob/develop/tests/testdata/input_contracts/version_3.sol). The comments should be removed from these files.

@norhh Please check the updated code. Now, it first searches for a regular pragma that does not occur in a comment or string, and if this fails, falls back to how Mythril searches the pragma currently (i.e., searching within comments as well).

gsalzer:
What is the proper way to execute the tests in the `tests` folder?

mythril/ethereum/util.py
```diff
@@ -168,18 +168,72 @@ def parse_pragma(solidity_code):
     all_versions = solcx.get_installed_solc_versions()
 
 
+# remove_comments_strings(str):
+# Return str without Solidity comments and strings
+import re
```
norhh: Can you move the import to the top of the file?

mythril/ethereum/util.py
```diff
@@ -168,18 +168,72 @@ def parse_pragma(solidity_code):
...
     if file is None:
         return None
     version_line = None
-    for line in file.split("\n"):
-        if "pragma solidity" not in line:
-            continue
-        version_line = line.rstrip()
-        break
+
+    # normalize line endings
```
norhh, line 220: This part of extracting the version_line can now be moved to separate function. It'll be easier to read and understand.

mythril/ethereum/util.py
```diff
@@ -168,18 +168,72 @@ def parse_pragma(solidity_code):
...
+# remove_comments_strings(str):
+# Return str without Solidity comments and strings
+import re
+
+VOID_START = re.compile("//|/\\*|\"|'")
+QUOTE_END = re.compile("(?<!\\\\)'")
+DQUOTE_END = re.compile('(?<!\\\\)"')
+
+
+def remove_comments_strings(todo):
```
norhh: Can you improve the names for this function? (namely: todo, done, m, m2)

mythril/ethereum/util.py
```diff
@@ -168,18 +168,72 @@ def parse_pragma(solidity_code):
     all_versions = solcx.get_installed_solc_versions()
 
 
+# remove_comments_strings(str):
+# Return str without Solidity comments and strings
```
norhh: This can be moved to the function def.

norhh (changes requested):
Thanks a lot for the PR. Can you make these quick refactoring changes?

gsalzer:
@norhh Refactoring done.

norhh (approved):
Thanks for the contribution!

mythril/ethereum/util.py
```diff
@@ -168,16 +169,77 @@ def parse_pragma(solidity_code):
...
 
 
-def extract_version(file: typing.Optional[str]):
-    if file is None:
+VOID_START = re.compile("//|/\\*|\"|'")
+QUOTE_END = re.compile("(?<!\\\\)'")
+DQUOTE_END = re.compile('(?<!\\\\)"')
+
+
+def remove_comments_strings(program):
```
psantos-consensys: thanks all, just a minor thing before merging and I will leave it to @norhh to decide, should we add annotations to the new methods definitions (remove_comments_strings, extract_version_line) for the argument and return types similar to:
`def extract_binary(file: str) -> Tuple[str, str]:`
`def extract_version(program: typing.Optional[str]):`

gsalzer: I have added type annotations to the functions that I have touched.

gsalzer: Doing it is probably faster than discussing it ...

psantos-consensys (approved):

PR #1892, opened 2025-01-13
Update pre-commit hooks, replace black with black compatible ruff-format

dbast:
The [Ruff Formatter](https://docs.astral.sh/ruff/formatter/) is an extremely fast Python code formatter designed as a drop-in replacement for Black.

This leads to some adjustments as ruff already contains 2025 black style updates, see https://github.com/astral-sh/ruff/issues/13371


dbast:
@norhh Ready for review / merge. Thanks

norhh (approved):
Thanks for the contribution!

dbast:
Green again :)

